PR #6, opened 2021-05-18
Generic Directory Storage

bbengfort:
This PR paves the way for other embedded directory storage such as sqlite3. The `store.Open` method now takes generic DSNs in the form of `protocol:///relative/path/to/db` or `protocol:////absolute/path/to/db`, e.g. `leveldb:///fixtures/db` or `sqlite3:///fixtures/directory.db`. The `ldbStore` is moved to its own module and a new `sqlite` module with the `store.Store` interface is created. The sqlite store is not implemented and returns an error when it is opened.


rebeccabilbro (approved):
Thank you @bbengfort for this PR that opens the GDS up to relational storage options!

PR #8, opened 2021-05-21
WIP: GDS UI Local Storage and Load/Download

bbengfort:
Enhances the GDS UI registration form with local storage so that the
form persists between browser tabs and when the browser is reloaded,
making the long form easier to fill out. It also adds a reset button the
summary page to clear the form out and start over.

Additionally adds a download button to download the JSON data for the
form and a load button to load downloaded data back into the form. This
makes the form usable for editing registration forms before the database
is live.

TODO:

- [x] Reset doesn't work if you reload without changing tabs?
- [x] No form validation on load is occurring, so it's pretty dangerous.

### DIFF
--- a/web/src/App.js
+++ b/web/src/App.js
@@ -23,8 +23,14 @@ class App extends React.Component {
         results: {},
         alerts: [...prevState.alerts, {variant: 'danger', message: err.message}]
       }));
-      console.log(err);
+      console.warn(err);
     }
+  }
+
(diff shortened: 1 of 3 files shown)


bbengfort:
Right now form validation is just a check if there is a version: "v1beta1" key which isn't robust but will prevent 75% of accidents. If a bad json is loaded right now the app will just crash - which is probably the best thing to happen.

PR #9, opened 2021-05-24
Maintenance mode

bbengfort:
Adds a configuration and interceptor to respond with unavailable if the server is in maintenance mode. Currently using this to deploy the backups, ensuring data won't get corrupted since no RPCs will respond. The purpose of this PR is to determine if the mode/config is generally useful and should be merged into the main codebase. 

If it is generally useful, another discussion point is if we should use middleware, or if we should deploy a maintenance server that correctly responds to status messages and does a better job of guaranteeing that no data corruption will happen accidentally.


cmd/gds/main.go
```diff
@@ -490,8 +492,17 @@ func initClient(c *cli.Context) (err error) {
...
+	opts := protojson.MarshalOptions{
+		Multiline:       true,
+		Indent:          "  ",
+		AllowPartial:    true,
+		UseProtoNames:   true,
+		UseEnumNumbers:  false,
+		EmitUnpopulated: true,
+	}
+
+	data, err := opts.Marshal(m)
```
bbengfort, line 505: @rebeccabilbro made a minor change here to use `protojson` to serialize the responses rather than the `json` library - that way the enums are printed correctly. `protojson` handles data much better, so I'll probably switch to it in all of the random `printJSON` functions we have in the library!

pkg/gds/interceptor.go
```diff
@@ -0,0 +1,28 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func (s *Server) serverInterceptor(ctx context.Context, in interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (out interface{}, err error) {
+	// Track how long the method takes to execute.
+	start := time.Now()
+
+	// Check if we're in maintenance mode - status method should still return a full response.
+	if s.conf.Maintenance && info.FullMethod != "/trisa.gds.api.v1beta1.TRISADirectory/Status" {
```
bbengfort, line 18: @rebeccabilbro made a minor change here so that the `Status` RPC would not return `Unavailable`. Instead, the `Status` RPC will report maintenance mode in a successful response.

pkg/gds/server.go
```diff
@@ -522,12 +524,17 @@ func (s *Server) Status(ctx context.Context, in *api.HealthCheck) (out *api.Serv
...
 	out = &api.ServiceState{
 		Status:    api.ServiceState_HEALTHY,
 		NotBefore: now.Add(30 * time.Minute).Format(time.RFC3339),
 		NotAfter:  now.Add(60 * time.Minute).Format(time.RFC3339),
 	}
 
+	// If we're in maintenance mode, update the service state.
+	if s.conf.Maintenance {
+		out.Status = api.ServiceState_MAINTENANCE
+	}
```
bbengfort, line 537: @rebeccabilbro last minor change since your review - the `status` rpc can report maintenance mode without an error.

PR #12, opened 2021-05-26
Updates GDS with v1beta1 protocol buffer changes.

bbengfort:
Adds protocol buffers that are directory specific, both moved from trisa protos and created in response to v1beta1 changes. Implements changes to Service to accommodate the v1beta1 changes.

There are a lot of changes that allow the code to compile, but the server has not been checked for correctness. The tests pass, but there are limited tests in this package. This code cannot be deployed until the database is migrated. We should either continue working off this branch or merge a broken version, tagging it as such.

See: https://github.com/trisacrypto/trisa/pull/49


rebeccabilbro (approved):
Looks good @bbengfort; we're very close to a complete release! I'm going to approve and merge to main, and then tag main as `v1beta1-pre-release` (hopefully that's what you had in mind also).

For the record, here is an overview of the changes made in this PR:
- updating `pb` package to `models`
- change old `Status` to `Verification`, and add new `Status` (aka health check)
- updating `VASPCategories` from enum to []string
- extract admin token and contact verification from 'extra' from the protobuf
- add `CertificateRequest` in `models.proto`
- moving pw creation and return from `VerifyEmail` to `Register`
- JS changes related to updated enums in TRISA (which fixed a bug and made the default/0 value 'Unknown')

PR #14, opened 2021-05-27
Secrets manager for passwords

### LINKED ISSUE #10: Implement Google Secrets for password storage
This task involves refactoring the current password encryption and storage strategy to adopt Google Secrets for consistency across projects. Subtasks include:

- Setup
    - [x] create new project in Google Secrets
    - [x] enable API access
    - [x] get service creds
    - [x] modify `config.go` to know about service creds path
- Password Storage 
    - [x] in `server.go` ~make `Register` return pw instead of `VerifyEmail`.~ Deprecate encryption methodology in favor of storing pw in Google secrets.
    - [x] in `secret.go` in `CreateToken` add special chars
    - [x] in `secret.go` check that `Encrypt` and `Decrypt` aren't used anywhere else and if so, flag for deprecation in favor of Google secrets.
    - [x] in `secret.go` add `StorePassword` and `RetrievePassword` using the `DirectoryID` and `commonName` in the store path in Google Secrets
- Password Retrieval
    - [x] in `cert.go` on startup check that `CertMan` can connect to Google Secrets
    - [x] in `cert.go` in `submitCertificateRequest` look up pw in Google Secrets
    - [x] create `secret_test.go` create a test that first retrieves an environment variable that determines whether or not the test should run, and if it does run, tests password retrieval (note this will be the only way to test this without going through the certificate process)
- Protocol Buffers
    - [x] remove `pkcs12password` and `pkcs12signature` from `models.proto` (inside `proto/models/v1` ~which will be introduced in a forthcoming PR from @bbengfort~)

rebeccabilbro:
This PR contains the start of the implementation of Google Secret Manager for password storage and retrieval. Closes #10 

## Here is what I have done:
- created new project in Google Secrets
- enabled API access
- set up `testnet` service creds (stored locally in fixtures for now)
- modified `config.go` to know about service creds
- deprecated encryption methodology in favor of storing pw in Google secrets in `server.go`
- added special chars to `CreateToken` in `secret.go`
- removed `Encrypt`, `Decrypt`, and related helper methods from `secret.go`
- looked for other places where `Encrypt` and `Decrypt` were used , and removed (some marked as todos)
- added new helper methods for interacting with secret manager (`NewSecretManager`, `CreateSecret`, `AddSecretVersion`, `GetLatestVersion`, etc) to `secret.go` 
- in `cert.go` on startup check that `CertMan` can connect to Google Secrets and updated `submitCertificateRequest` to lookup pw from Google Secrets
- removed `pkcs12password` and `pkcs12signature` from `models.proto`

## TODO
- regenerate protocol buffers (two fields removed in this PR)

## Questions:
- unsure if other decryption code needs to be removed


pkg/gds/secret.go
```diff
@@ -28,108 +27,180 @@ func CreateToken(length int) string {
...
+		// If the API call is malformed, it will hang until the internal context times out
+		if errors.Is(err, context.DeadlineExceeded) {
+			return err
+		}
+
+		// If the secret does not exist (e.g. has been deleted or hasn't been created yet)
+		// we'll get a Not Found error
+		serr, ok := status.FromError(err)
+		if ok && serr.Code() == codes.NotFound {
+			return err
```
bbengfort, line 139: Should we create our own NotFoundSecret? E.g. at the top of the file we could add:

```
var (
    ErrSecretNotFound = errors.New("could not add secret version - not found")
)
```

Here we would `return ErrSecretNotfound` Then in downstream code we could use `errors.Is(err, ErrSecretNotFound)` to do testing similar to the `context.DeadlineExceeded`.

pkg/gds/secret.go
```diff
@@ -28,108 +27,180 @@ func CreateToken(length int) string {
...
+		}
+
+		// If the error is something else, something went wrong.
+		return errors.New("unknown error: unable to create secret version")
```
bbengfort, line 143: It would be preferable to pass the error up to the CertManager for logging, e.g.:

```
return fmt.Errorf("unable to create secret version: %s", err)
```

pkg/gds/secret.go
```diff
@@ -28,108 +27,180 @@ func CreateToken(length int) string {
...
-	ciphertext = aesgcm.Seal(nil, nonce, []byte(plaintext), nil)
-	if len(ciphertext) == 0 {
-		return nil, nil, errors.New("could not encrypt secret with aes gcm")
+// AddSecretVersion adds a new secret version to the given secret and the
+// provided payload. Returns an error if one occurs.
+// Note: to add a secret version, the secret must first be created using CreateSecret.
+func (sm *SecretManager) AddSecretVersion(ctx context.Context, secret string, payload []byte) error {
+
+	secretPath := fmt.Sprintf("%s/secrets/%s-%s", sm.parent, sm.requestId, secret)
+	// Create an internal context, since a failed API call will result in infinite hang
```
bbengfort, line 114: Good comments!

pkg/gds/secret.go
```diff
@@ -28,108 +27,180 @@ func CreateToken(length int) string {
...
+		// If the API call is malformed, it will hang until the internal context times out
+		if errors.Is(err, context.DeadlineExceeded) {
+			return nil, err
+		}
 
-	aesgcm, err := cipher.NewGCM(block)
-	if err != nil {
-		return "", err
+		// If the error is something else, something went wrong.
+		return nil, errors.New("unknown error: unable to access latest secret version")
```
bbengfort, line 173: Could we also make this `return nil, fmt.Errorf("unable to access latest secret version: %s", err)` - thank you!

pkg/gds/secret.go
```diff
@@ -28,108 +27,180 @@ func CreateToken(length int) string {
...
-		return errors.New("HMAC mismatch")
+	// Call the API.
+	err := sm.client.DeleteSecret(sctx, req)
+	if err != nil {
+		// If the API call is malformed, it will hang until the internal context times out
+		if errors.Is(err, context.DeadlineExceeded) {
+			return err
+		}
+		// If the error is something else, something went wrong.
+		return errors.New("unknown error: unable to delete secret")
```
bbengfort, line 203: Similar to above: `return fmt.Errorf("unable to delete secret; %s", err)`

pkg/gds/server.go
```diff
@@ -214,16 +217,85 @@ func (s *Server) Register(ctx context.Context, in *api.RegisterRequest) (out *ap
...
+		}
+		// If it's a more serious problem, that likely signals a permissions error
+		// and merits a return.
+		log.Error().Err(err).Msg("secret manager error; is GOOGLE_PROJECT_NAME set?")
+		out.Error = &api.Error{
+			Code:    403,
+			Message: err.Error(),
+		}
+		return out, nil
+	}
```
bbengfort, line 263: Because this checking already happens in `CreateSecret`, I don't think it's necessary to do it again - we want to push as much responsibility to the SecretManager as possible. The only question is about what error to return for `AlreadyExists` and there are two options for how to handle that:

Option 1 - `CreateSecret` returns nil on AlreadyExists; then the check here is simply:

```go
if err = sm.CreateSecret(ctx, secret); err != nil {
    log.Error().Err(err).Msg("could not create new secret on registration")
    out.Error = &api.Error{Code: 500, Message: "internal error with registration, please contact admins")
    return out, nil
}
```

Note that the `log.Error().Err(err)` will be giving us details about the error from those `fmt.Errorf` comments I made before. We're also hiding the error from the user - we just want them to know something really bad happened and they should email us. We don't want them to know it was google secrets. Probably we should make this kind of change everywhere in the code, returning the `Message: err.Error()` is my own 🎃   code and I can create a technical debt issue for dealing with it. 

Option 2- If we want to return an `AlreadyExists` error - then we can convert the error into our own custom error like the `ErrSecretNotFound` suggestion I made earlier. The handler would then be:

```go
if err = sm.CreateSecret(ctx, secret); err != nil && !errors.Is(err, ErrSecretAlreadyExists) {
    log.Error().Err(err).Msg("could not create new secret on registration")
    out.Error = &api.Error{Code: 500, Message: "internal error with registration, please contact admins")
    return out, nil
}
```

pkg/gds/server.go
```diff
@@ -214,16 +217,85 @@ func (s *Server) Register(ctx context.Context, in *api.RegisterRequest) (out *ap
...
+			}
+			return out, nil
+		}
+
+		log.Error().Err(err).Msg("unable to write new secret version to secret manager")
+		out.Error = &api.Error{
+			Code:    403,
+			Message: err.Error(),
+		}
+		return out, nil
```
bbengfort, line 294: As above, my review comments have moved a lot of responsibility for error handling to the client. I propose this should be:

```go
if err = sm.AddSecretVersion(ctx, secret, []byte(password)); err != nil {
    log.Error().Err(err).Msg("unable to add secret version on registration")
    out.Error = &api.Error{Code: 500, Message: "internal error during registration, please contact admins"}
    return out, nil
}
```

pkg/gds/server.go
```diff
@@ -214,16 +217,85 @@ func (s *Server) Register(ctx context.Context, in *api.RegisterRequest) (out *ap
...
 	}
 
 	if err = s.db.SaveCertRequest(certRequest); err != nil {
-		log.Error().Err(err).Msg("could save certificate request for VASP")
+		log.Error().Err(err).Msg("could not save certificate request for VASP")
```
bbengfort, line 298: Good catch!

pkg/gds/server.go
```diff
@@ -406,8 +478,7 @@ func (s *Server) Verification(ctx context.Context, in *api.VerificationRequest)
 
 // VerifyEmail checks the contact tokens for the specified VASP and registers the
 // contact email verification. If successful, this method then sends the verification
-// request to the TRISA Admins for review and generates a PKCS12 password in the RPC
-// response to decrypt the certificate private keys when they're emailed.
+// request to the TRISA Admins for review.
```
bbengfort, line 481: Thank you for being so thorough with the comments and making sure the documentation is up to date! Great catch!

proto/gds/models/v1/models.proto
```diff
@@ -28,23 +28,9 @@ message CertificateRequest {
...
-    // stored in the certificate request which is maintained by independent replicas.
-    // TODO: Rebecca to remove for Google Secrets
-    bytes pkcs12password = 13;
-    bytes pkcs12signature = 14;
-
     // Logging information timestamps
-    string created = 15;
-    string modified = 16;
+    string created = 13;
+    string modified = 14;
```
bbengfort, line 33: Hmm … so this will cause some issues with the migration. We have two options:

1. Delay the "actual" migration until this code is merged in (preferred)
2. Leave these as fields 15/16 and note in a comment that we have empty fields for future versions

I'm happy to go with Option 1, but we'd need to get this merged in tomorrow - any thoughts?

bbengfort (commented):
Wow, this was a massive effort and a major contribution to the TRISA Global Directory Service! Thank you so much for your help getting the Google Secrets feature added from researching the API to an implementation that fits right in with the codebase. It is excellently documented, well thought out, and it really takes a load off my shoulders knowing that we're relying on Google secure storage for these certificate passwords, adding an extra layer of security to the directory service process. 

This was a big PR and there are a lot of comments. I hope they're mostly understandable and none of them are too frustrating to deal with. 

I am excited - though - because this PR gives me the opportunity to suggest a code pattern that I really like, but that I rarely have the opportunity to use! Your implementation of the `SecretManager` fits neatly into it! 

At the moment, the code relies on calling `NewSecretManager` in `Register()`, `CertManager()`, `submitCertificateRequest()`, and `downloadCertificateRequest()` - this is totally understandable, particularly because the `NewSecretManager` takes a `certRequest` argument. Unfortunately, this causes `secretmanager.NewClient` to be called multiple times, reconfiguring from the environment each time. 

In a long-running service like GDS, we only want the configuration to happen once at the beginning to avoid problems with the environment changing during runtime. Plus there are performance issues - connecting like that each time slows things down; a single gRPC client (which secretManager.Client is) is thread-safe for multiple go routines and is optimized for throughput. 

So how do we get both a call to `NewClient` once plus the opportunity to use the `certRequest` as internal state? We use a chaining method trick that creates a child with the temporary state for the request, then goes away when the request is over. E.g. the majority of the code will look something like:

```
sm, err := NewSecretManager(config)

req := sm.With(certRequest, "password")
err = req.CreateSecret(ctx)
err = req.AddSecretVersion(ctx, payload) 

… 

password, err := sm.With(certRequest, "password").GetLatestVersion()
```

Because of the `With` function, let's tag this as the "with context pattern". 

So how is this implemented? Broadly:

In [`type Sever struct{}`](https://github.com/trisacrypto/directory/blob/main/pkg/gds/server.go#L70) add a property `secret *SecretManager`. 

In [`func New()`](https://github.com/trisacrypto/directory/blob/main/pkg/gds/server.go#L61) create the new secret manager, `if s.secret, err = NewSecretManager(conf.Secret); err != nil {}`. This will be the only place where the client is created. 

Everywhere where `NewSecretManager` is called now, replace with `s.secret.With(certRequest, secretType)`. 

Then `SecretManager` has to be updated with a `With` method that returns a contextual temporary struct:

```go
type SecretManager struct {
    parent string
    client *secretmanager.Client
}

type secretManagerContext struct {
    manager *SecretManager
    requested string
    secretType string 
}

func (sm *SecretManager) With(certRequest, secretType string) *secretManagerContext {
    return &secretManagerContext{
        manager: sm, 
        requestID: certRequest,
        secretType: secretType,
    }
}
```

The `CreateSecret`, `AddSecretVersion`, `GetLatestVersion`, and `DeleteSecret` methods will need a bit of tweaking - either they can be converted to be methods of `secretManagerContext` then all they will need to do is call `smc.manager.client` and tweak the arguments to use the internal state. Or the methods can be copied and the context methods simply call the parent methods using the internal state. 

I hope this makes sense - I think it is a pretty simple change that will have a big impact, and use a cool pattern also! I recognize that this is a lot of text and that with all of the other PR comments it might be hard to digest; so I'm happy to handle all of the PR changes in 🍐 .

cmd/debug/main.go
```diff
@@ -133,20 +133,6 @@ func main() {
...
-			Category:  "cipher",
-			Action:    cipherDecrypt,
-			Flags: []cli.Flag{
-				cli.StringFlag{
-					Name:   "k, key",
-					Usage:  "secret key to decrypt the cipher text",
-					EnvVar: "GDS_SECRET_KEY",
-				},
-			},
-		},
```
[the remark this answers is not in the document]
rebeccabilbro: Ok, I've added it back.

proto/gds/models/v1/models.proto
rebeccabilbro: Ok, I changed them back to 15/16

rebeccabilbro:
Thank you for the feedback @bbengfort. I think I've made all the changes you requested, but let me know if there's anything I missed!

bbengfort (approved):
Thank you so much for working through all those comments. I really like the `With()` design for this, it makes the code look very clean, safer, and more performant. I hope it wasn't too much trouble to refactor it. There is one minor item remaining, a missing `return nil, err` -- but besides this the code is ready to be merged! Excellent work!

pkg/gds/server.go
```diff
@@ -217,80 +220,16 @@ func (s *Server) Register(ctx context.Context, in *api.RegisterRequest) (out *ap
...
-		// If it's a more serious problem, that likely signals a permissions error
-		// and merits a return.
-		log.Error().Err(err).Msg("secret manager error; is GOOGLE_PROJECT_NAME set?")
-		out.Error = &api.Error{
-			Code:    403,
-			Message: err.Error(),
-		}
+	secretType := "password"
+	if err = s.secret.With(certRequest.Id, secretType).CreateSecret(ctx, secretType); err != nil {
+		log.Error().Err(err).Msg("could not create new secret on registration")
```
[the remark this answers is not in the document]
rebeccabilbro: I changed CreateSecret to return nil on AlreadyExists

PR #15, opened 2021-05-28
Fix pb marshalling for the extra attribute

rebeccabilbro:
This PR fixing the marshalling of the extra attribute using `anypb`.


bbengfort (approved):
Thank you so much for fixing this bug - we definitely want `extra` fields to be `nil` by default, so we can't rely on them being values or pointers that we can use the `MarshalFrom` method. Using `anypb.New` was the right way to do this. Nice catch!

PR #19, opened 2021-05-31
Write certificates to Secret Manager

### LINKED ISSUE #11: Implement Google Secrets for Certificate Storage
Related to #10, blocked by #10

This task involves updating the certificate storage mechanism to adopt Google Secrets storage for consistency across projects. Subtasks include:

- [x] ~in `cert.go` in [`downloadCertificateRequest`](https://github.com/trisacrypto/directory/blob/main/pkg/gds/certs.go#L253-L256) create a temp dir and store locally*. Temp dir should be deleted at the end of the function.~
- [x] in `cert.go` in [`downloadCertificateRequest` before the certificate request is marked as downloaded](https://github.com/trisacrypto/directory/blob/main/pkg/gds/certs.go#L270), store zipped cert in Google Secrets using `certRequestId`
- [x] After successfully storing in Google Secrets and marking CertificateRequest as Downloaded, call `defer os.Remove(path)` on the downloaded [`path`](https://github.com/trisacrypto/directory/blob/main/pkg/gds/certs.go#L266) 
- [x] Change [log message](https://github.com/trisacrypto/directory/blob/main/pkg/gds/certs.go#L277) to reflect storage in Google Secrets
- [x] Change the `context.Background` to be a shared context inside `downloadCertificateRequest` to be used for all calls to secret manager inside that function
- [x] Add test to `secret_test` that tests that you can store a cert to secret manager, and then access it, unzip it and decrypt it using the password; the test cert and test pw should be environment variables
- [x] ~in `cert.go` in [`downloadCertificateRequest`](https://github.com/trisacrypto/directory/blob/main/pkg/gds/certs.go#L280-L284) retrieve pw from google secrets & use to call `extractCertificate`~(done in #14)

*Temporary directories on pods use `tempfs`, which is an in-memory store that can [cause the pod to run out of memory](https://stackoverflow.com/questions/42963337/kubernetes-pod-running-out-of-memory-when-writing-to-a-volume-mount). We have three options:

1. Mount a volume as we're currently doing with [`getCertStorage()`](https://github.com/trisacrypto/directory/blob/main/pkg/gds/certs.go#L385) and simply treat it as a temporary dir and manage deleting things ourselves (easiest solution). However, this option will not automatically clean up if something goes wrong. 
2. Use `tempfs` and increase the memory of the pod (this will use standard golang `ioutil.TempDir`) 
3. Use `emptyDir` in the pod - this will only affect the manifests (in code, still use golang `ioutil.TempDir`)

rebeccabilbro:
In #14 we introduced the new secret package, which allowed us to refactor the password storage and retrieval process to Google Secret Manager for consistency across projects. 

In this PR, I have similarly updated the certificate storage to store zipped certificates directly to Secret Manager as bytes. The only catch is that as currently implemented, the TRISA trust package requires that zipped certificates be [read in as files from disk](https://github.com/trisacrypto/trisa/issues/51); so the tests added to validate that cert storage is working as expected leverage an intermediate temporary file write/delete.  

In the process of working on this PR, I also experimented more with the Google Secret Manager API to see what other kinds of errors can arise; the main one that I think we want to capture is the file size limit error (`rpc error: code = InvalidArgument desc = Secret payload data length must be at most [65536] bytes.`), now that we'll be storing the zipped certificates. The limit is 65KiB (or 64KiB, [depending on where you look](https://cloud.google.com/secret-manager/quotas)), so we probably won't hit it, but I've added a new error code to `secret.go` to try to catch this just in case.

Closes #11

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -64,6 +64,7 @@ github.com/envoyproxy/go-control-plane v0.9.7/go.mod h1:cwu0lG7PUMfa9snN8LXBig5y
 github.com/envoyproxy/go-control-plane v0.9.9-0.20201210154907-fd9021fe5dad/go.mod h1:cXg6YxExXjJnVBQHBLXeUAgxn2UodCpnH306RInaBQk=
 github.com/envoyproxy/go-control-plane v0.9.9-0.20210217033140-668b12f5399d/go.mod h1:cXg6YxExXjJnVBQHBLXeUAgxn2UodCpnH306RInaBQk=
 github.com/envoyproxy/protoc-gen-validate v0.1.0/go.mod h1:iSmxcyjqTsJpI2R4NaDN7+kN2VEUnK/pcBlmesArF7c=
+github.com/fsnotify/fsnotify v1.4.7 h1:IXs+QLmnXW2CcXuY+8Mzv/fWEsPGWxqefPtCP5CnV9I=
 github.com/fsnotify/fsnotify v1.4.7/go.mod h1:jwhsz4b93w/PPRr/qN1Yymfu8t87LnFCMoQvtojpjFo=
 github.com/go-gl/glfw v0.0.0-20190409004039-e6da0acd62b1/go.mod h1:vR7hzQXu2zJy9AVAgeJqvqgH9Q5CA+iKCZ2gyEVpxRU=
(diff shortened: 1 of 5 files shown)


bbengfort (approved):
:shipit:  based on 🍐  review

PR #28, opened 2021-06-01
Remove bang from token special characters

rebeccabilbro:
Removes the bang character from the token generation function for creating pkcs12 passwords, which was causing an issue on the command line since bang wasn't getting escaped.


bbengfort (approved):
Thanks! Glad to have my bash terminal copy and paste working again!

PR #29, opened 2021-06-01
Front-End QA Items

bbengfort:
Fixes the following front end bugs caught in QA:

- On registration validation error, submit button stays disabled
- Typos in TRIXO form and confirmation modal
- Uncaught exception when no identity ceritificate is present

### DIFF
--- a/web/src/components/Registration.js
+++ b/web/src/components/Registration.js
@@ -122,6 +122,12 @@ class Registration extends React.Component {
           });
         })
         .catch(err => {
+          this.setState({
+            submitting: false,
+            validated: false,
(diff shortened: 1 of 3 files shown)


web/src/components/Registration.js
```diff
@@ -122,6 +122,12 @@ class Registration extends React.Component {
           });
         })
         .catch(err => {
+          this.setState({
+            submitting: false,
+            validated: false,
```
rebeccabilbro, line 127: maybe should be `validated: true` ?

rebeccabilbro (approved):

PR #35, opened 2021-06-05
Implements JSON logging with GCP severity

### LINKED ISSUE #33: Logging for Google Console
Change logging to JSON format in production (e.g. add a configuration that allows human-readable vs json) so logs are easier to parse in the Google console. Add GCP severity levels. Note that these items are implemented in Traveler for zerolog, which we can use as a template for enhanced logging in GDS. 

Configure GCP to capture logs from GDS pods (if there is a unified logging mechanism like AWS).

bbengfort:
By default GDS now emits JSON formatted logs for parsing in a cloud
console. This PR also adds GCP severity levels for GCP-specific parsing
of logs. The logger is added to its own module so other cloud-specific
handlers can be added as needed.

Fixes #33

### DIFF
--- a/pkg/gds/server.go
+++ b/pkg/gds/server.go
@@ -33,7 +33,7 @@ func init() {
 	// Initialize zerolog with GCP logging requirements
 	zerolog.TimeFieldFormat = time.RFC3339
 	zerolog.TimestampFieldName = logger.GCPFieldKeyTime
-	zerolog.MessageFieldName = logger.GCPFIeldKeyMsg
+	zerolog.MessageFieldName = logger.GCPFieldKeyMsg
 
 	// Add the severity hook for GCP logging
 	var gcpHook logger.SeverityHook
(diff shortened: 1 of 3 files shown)


pkg/gds/server.go
```diff
@@ -28,6 +29,15 @@ import (
 func init() {
 	zerolog.TimeFieldFormat = zerolog.TimeFormatUnix
 	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
+
+	// Initialize zerolog with GCP logging requirements
+	zerolog.TimeFieldFormat = time.RFC3339
+	zerolog.TimestampFieldName = logger.GCPFieldKeyTime
+	zerolog.MessageFieldName = logger.GCPFIeldKeyMsg
```
rebeccabilbro, line 36: Should be `logger.GCPFieldKeyMsg`?

rebeccabilbro (approved):
Looks good @bbengfort; I noticed what I believe might be a small typo &mdash; could you double check before merging?

pkg/gds/server.go
bbengfort: Good catch!

PR #52, opened 2021-06-07
Reindex and gzip index sync without DB in maintenance mode

### LINKED ISSUE #30: Reindexing LevelDB and index sync
Create a configuration `GDS_REINDEX_ON_BOOT` so that on `Serve`, if the store is an [Indexer](https://github.com/trisacrypto/directory/blob/main/pkg/gds/store/store.go#L69) then reindex the database:

```
func (s *Server) Serve() (err error) {
...

// Reindex the database if configured
if s.conf.ReindexOnBoot {
    if indexer, ok := s.store.(store.Indexer); ok {
        indexer.Reindex()
        log.Info().Msg("reindex complete")
    }
}

...
}
```

We should also periodically sync the leveldb indices back to disk (`s.sync()`). This is currently only being called on `ldbStore.Close()` - but we should have a goroutine that periodically performs the sync according to some interval.

### LINKED ISSUE #47: LevelDB Index enhancements
1. More compact LevelDB indices. Right now indices are just saved as raw JSON (_note: both [here](https://github.com/trisacrypto/directory/blob/main/pkg/gds/store/leveldb/index.go#L170) and [here](https://github.com/trisacrypto/directory/blob/main/pkg/gds/store/leveldb/index.go#L290)_). We can either save them as a protocol buffer, or gzip them. This should be an optional configuration. This is only a problem if the indices get really big!

2. In `Update` [check errors from index removal in case the index must be reset](https://github.com/trisacrypto/directory/blob/main/pkg/gds/store/leveldb/leveldb.go#L210)

bbengfort:
This PR more intelligently handles leveldb indices and reindexing by allowing a configuration that forces reindex on boot and creates a maintenance mode that allows database interactions from the command line. 

Updates the database configuration to accept a bool for reindexing on
boot and the store reindexes if necessary on open. Adds a call to sync
to on Backup so that the indices are resync'd routinely.

Fixes #30

This commit enables more compact LevelDB indices by compressing them
before writing them to disk. However, this requires deleting old indices
and regenerating them. To facilitate this, during maintenance mode, the
database no longer opens a connection to the database or runs any
management processes. This will allow us to ssh into the server and
access the database as needed to perform any changes.

Fixes #47

**Warning: Before Deployment a Maintenance Mode is Required to Delete Old Indices!!**

### DIFF
--- a/cmd/gds/main.go
+++ b/cmd/gds/main.go
@@ -265,7 +265,7 @@ func load(c *cli.Context) (err error) {
 	}
 
 	var db store.Store
-	if db, err = store.Open(dsn); err != nil {
+	if db, err = store.Open(config.DatabaseConfig{URL: dsn}); err != nil {
 		return cli.NewExitError(err, 1)
 	}
 	defer db.Close()
--- a/cmd/gdsutil/main.go
+++ b/cmd/gdsutil/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"compress/gzip"
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/hmac"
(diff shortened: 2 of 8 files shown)


rebeccabilbro (approved):
Looks great @bbengfort; I noted a place where we might want to capture the Reindex error, since it looks like there are a few different things that could happen there. This also makes me wonder if it would be a good idea to start documenting maintenance mode and admin instructions somewhere, so that we have a handy checklist admin-level fixes in the future (such as a deliberate reindex).

PR #53, opened 2021-06-08
Anti-Entropy Protocol Buffers

bbengfort:
Creates the global Replication service and related object versioning and anti-entropy messages along with a Peers message definition and management service. Modifies other models with anti-entropy related requirements.

TODO:

- [x] Discuss peer management service (e.g. is it necessary, should we simplify the number of messages instead of req/reply)
- [x] Discuss CertificateRequest changes and required migration
- [x] Finalize protocol buffers for first version of anti-entropy (e.g. should we use any or oneof?)
- [x] Create the generate.go files in `pkg`


proto/gds/global/v1/global.proto
```diff
@@ -0,0 +1,94 @@
...
+    // Version vectors of objects without data.
+    repeated Object objects = 1;
+
+    // A request to consider only the objects specified in the version vector and not
+    // the entire namespace (e.g. ignore objects that are not specified).
+    bool partial = 2;
+
+    // Limit the comparison only to the specified namespace(s). If not set or empty,
+    // then all namespaces are considered.
+    repeated string namespaces = 3;
```
bbengfort, line 50: Added this field which allows for partial replication, but may also simplify things; e.g. if we just want to deal with Peers.

proto/gds/global/v1/global.proto
```diff
@@ -0,0 +1,94 @@
...
+// recommended that an objects table be kept for fast lookups of object versions.
+// Additionally, it is recommended that a history table be maintained locally so that
+// Object versions can be rolled back to previous states where necessary.
+message Object {
+    // The object metadata that must be populated on both VersionVectors and Updates
+    string key = 1;                  // A unique key/id that represents the object across the namespace of the object type
+    string namespace = 2;            // The namespace of the object, used to disambiguate keys or different object types
+    Version version = 3;             // A version vector representing this objects current or latest version
+    string region = 4;               // The region code where the data originated
+    string owner = 5;                // The replica that created the object
```
bbengfort, line 78: How should we identify this back to the Peer?

proto/gds/global/v1/global.proto
```diff
@@ -0,0 +1,94 @@
...
+
+    // The object data that is only populated on Updates.
+    oneof data {
```
bbengfort, line 81: Right now using `oneof`, which specifies the types that can be replicated and requires no additional parsing (plus is a bit more compact) - however, if we add data types that can be replicated, then we will have to change the protocol buffers. 

Alternative: use `Any` - which will allow us to put any object here and is more generic, but would contain extra information (the type URL) and would require parsing on the other end. Maybe not so bad though given that we have the namespace identifier? 

So what do we prefer, flexibility or type safety?

proto/gds/models/v1/models.proto
```diff
@@ -8,29 +8,36 @@ option go_package = "github.com/trisacrypto/directory/pkg/gds/models/v1;models";
...
 message CertificateRequest {
-    // ID only needs to be unique with the requesting directory service.
+    // A unique identifier generated by the directory service, should be a globally
+    // unique identifier generated by the replica specified in requesting_replica.
     string id = 1;
 
+    // The replica that generated the certificate request should be the only replica
+    // that handles the certificate request. The request is replicated for information
+    // and backup purposes only.
+    string requesting_replica = 2;
```
bbengfort, line 18: Adding this here will require us to migrate the old CertificateRequest versions, but I think that's important so that only the requesting replica handles the certificate request.

proto/gds/models/v1/models.proto
```diff
@@ -8,29 +8,36 @@ option go_package = "github.com/trisacrypto/directory/pkg/gds/models/v1;models";
 // issuance and only public keys and certificate metadata should be exchanged between
 // directories.
 message CertificateRequest {
-    // ID only needs to be unique with the requesting directory service.
+    // A unique identifier generated by the directory service, should be a globally
+    // unique identifier generated by the replica specified in requesting_replica.
```
bbengfort, line 12: We should replicate certificate requests even though only one replica will handle the request. That way if the owner of the certificate request dies, it will be "automatically" backed up via anti-entropy and can then submit the certificate request. This means we need to track the owner of the certificate request, so I've added the "requesting replica" field below.

proto/gds/peers/v1/peers.proto
```diff
@@ -0,0 +1,34 @@
...
+    // specific clouds or data that's been parsed (optional).
+    map<string, string> extra = 15;
+}
+
+// PeerManagement provides a simple interface for administrators to debug the
+// anti-entropy network and to manage services. Note that administrators can also use
+// Gossip to perform the same requests, but this service simplifies things.
+service PeerManagement {
+    rpc GetPeers(GetPeersRequest) returns (GetPeersReply) {};
+    rpc AddPeers(AddPeersRequest) returns (AddPeersReply) {};
```
bbengfort, line 25: This could just be `rpc AddPeers(Peer) returns (Status)` - which might simplify things.

proto/gds/peers/v1/peers.proto
```diff
@@ -0,0 +1,34 @@
...
+message Peer {
+    uint64 id = 1;                   // the process id of the peer must be unique in the network; used for distributed versions
+    string addr = 2;                 // the network address to connect to the peer on (don't forget the port!)
+    string name = 3;                 // optional - a unique, human readable name for the peer
+    string region = 4;               // optional - the region the peer is running in
+
+    // extra information that might be relevant to process-specific functions; e.g. for
+    // specific clouds or data that's been parsed (optional).
+    map<string, string> extra = 15;
+}
```
bbengfort, line 18: Pretty basic Peer definition. I've added an extra field for storing arbitrary information, e.g. maybe cloud-specific or k8s information. I thought about storing anti-entropy information here as well; but I think all of that would be tracked in memory. Thoughts?

proto/gds/global/v1/global.proto
rebeccabilbro: Makes sense to me &mdash; probably a lot of partial updates will be at the level of granularity of namespaces, so I agree that this should simplify things

proto/gds/models/v1/models.proto
rebeccabilbro: Sounds good; glad we decided to name the requesting replica `owner` since that will make it align nicely with the `owner` field on `Object`

proto/gds/peers/v1/peers.proto
rebeccabilbro: Perhaps we should make a backlog issue to look into what we can get from kubernetes so we have a better notion of what might populate this map; I know we had a task like that in the backlog for another project but didn't get around to it yet,

rebeccabilbro (approved):
Thanks so much for fleshing out the RPC after our initial conversation @bbengfort! I think this is shaping up really nicely and I like the optimizations we've made (e.g. human-readable names as well as pids to differentiate pods, and the `google.protobuf.Any` protobufs on `Object` to support more flexible CD).

proto/gds/peers/v1/peers.proto
bbengfort: Issue made,  #55

PR #54, opened 2021-06-10
Implements Resend Admin RPC

bbengfort:
There have been a couple of issues with email deliveries over the first few registrations, requiring that emails be resent. This RPC allows an admin to trigger a resend of different email types, based on the condition of the VASP record. This commit also tracks the number of emails that are sent for logging purposes.

This PR also changes the deliver certificates handling to just send one email with the attachment, attempting until success the technical, administrative, legal, and billing contacts (in that order). 

The handling of the resend request depends on the resend type:

- _verify contact_: sends emails only to contacts who have not yet been verified, but (currently) doesn't change the verification token
- _review_: resends the review email to the TRISA admins (doesn't change admin token)
- _deliver certificates_: returns not implemented right now because we would have to retrieve the cert from Google secrets and there is a lot of work to be done to achieve this
- _rejection_: resends the rejection notice iff the verification state is rejected; the client must specify the reason because it is not stored in the VASP. 

I'm hesitant to put this up before authentication is added to the server, but we could do a quick and dirty mTLS authentication if we were worried about it, or enable the feature to resend emails, then disable it when we were done.


pkg/gds/emails/client_test.go
```diff
@@ -56,8 +56,19 @@ func TestClientSendEmails(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, 1, sent)
+
+	sent, err = email.SendRejectRegistration(vasp, "this is a test rejection from the test runner")
+	require.NoError(t, err)
+	require.Equal(t, 1, sent)
+
+	sent, err = email.SendDeliverCertificates(vasp, "testdata/foo.zip")
+	require.NoError(t, err)
+	require.Equal(t, 1, sent)
```
rebeccabilbro, line 73: I can't quite remember what's in our test example, but we might want to test to make sure that if there is a technical and say billing contact, that only the technical one gets it. And maybe we should double check that sent is not ever greater than 1 (but it shouldn't be).

pkg/gds/server.go
```diff
@@ -256,13 +256,14 @@ func (s *Server) Register(ctx context.Context, in *api.RegisterRequest) (out *ap
 	}
 
 	// Send contacts with updated tokens
-	if err = s.email.SendVerifyContacts(vasp); err != nil {
-		log.Error().Err(err).Str("vasp", vasp.Id).Msg("could not sennd verify contacts emails")
+	var sent int
+	if sent, err = s.email.SendVerifyContacts(vasp); err != nil {
+		log.Error().Err(err).Str("vasp", vasp.Id).Int("sent", sent).Msg("could not sennd verify contacts emails")
```
rebeccabilbro, line 261: small typo ("sennd" => "send")

rebeccabilbro (approved):
Looks good @bbengfort &mdash; and I like the newly added logic of limiting how many people on the contact list should get the most secure information like certificates (this is probably something we'd also want to do when and if we implement resending the passwords?). Just made a few notes about tiny things I noticed.

cmd/gds/main.go
```diff
@@ -476,6 +509,41 @@ func status(c *cli.Context) (err error) {
...
+	req := &admin.ResendRequest{
+		Id:     c.String("id"),
+		Reason: c.String("reason"),
+	}
+
+	if req.Id == "" {
+		return cli.NewExitError("missing VASP record ID, specify with --id", 1)
+	}
+
+	// NOTE: if multiple type flags are specified, only one will be used
```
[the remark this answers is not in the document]
bbengfort: Good question, I checked to make sure! Switches automatically break in Go: https://play.golang.org/p/jq2GCE-YoqG

pkg/gds/emails/client_test.go
bbengfort: I've added a test to make sure that only one contact gets the email; but it's tougher to test who got sent the email since the function doesn't return that email. We'd have to mock the email backend (which might be a good idea anyway) to determine the destination. Hopefully, the test I just pushed will hold us over for now.

PR #60, opened 2021-06-30
Setting the stage for anti-entropy replication

bbengfort:
This PR does a lot to set the foundation for anti-entropy including:

- Implements the protocol buffers for peers and global replication
- Updates the configuration for replication
- Decouples the GDS, Admin, and Replica services to run on different ports (for different levels of protection)
- Adds global version management to the Store 
- Refactors and unifies object store interface
- Adds tests, including a leveldb test suite


rebeccabilbro (approved):
Looks great @bbengfort! These were a lot of twisty, tangly updates that give us a clear path to do the replication implementation (and also gave us some ideas about future optimizations around data storage and management layer a la Honu)! 

In working on these updates, just wanted to note that we've decided to update the TRISA protocol buffers to remove the `version` field from the VASP definition, which wasn't being used there, and which will prevent name collision/confusion here in the GDS.

PR #62, opened 2021-06-30
Gossip Service

bbengfort:
Implements the Gossip Service for replicas - e.g. the RPC endpoint that responds to Gossip requests from remote replicas (e.g. half of the anti-entropy method) - as well as a gdsutil command to initiate and test gossip for debugging purposes. This PR implements:

- Version comparisons using Lamport scalars
- gdsutil to load objects from namespace and send to remote replica
- a Replica endpoint to handle gossip requests
- update version to make it easier to identify v1.1-alpha 

Gossip occurs in three phases:

1. Review incoming objects and determine if local is later (send back to repair remote), if remote is later (request in p hase three) or if the versions ar the same (do nothing)
2. If not partial, go through all local objects not in the incoming request and send back to remote
3. Not implemented: request all versions that are later on remote. 

Note: we have to be able to identify the remote peer, in order to make a request back. This will have to be implemented later. 

Next steps:

- Implement store object handling interface (don't directly use ldb)
- Better object parsing and handling (occurs in several places) 
- Identify remote peer for fetch objects  
- Implement timer to trigger anti-entropy from replica 
- Secure anti-entropy


cmd/gdsutil/main.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/syndtr/goleveldb/leveldb"
```
bbengfort, line 24: @rebeccabilbro one of the reasons there are so many changes in this file is because I reordered the peer management and gossip functions to keep everything together and moved the `decrypt` function to the bottom. I did this primarily to resolve the merge conflict when I merged v1.1 into this branch. However, this will make it tricky for you to work on gdsutil without this branch merged in. I'll wrap up the gossip implementation so we can merge it to unblock you ASAP.

pkg/gds/replica.go
```diff
@@ -112,46 +118,209 @@ func (r *Replica) Shutdown() error {
...
+			}
+
+			// Certreq specific method to retrieve metadata
+			localObj = certreq.Metadata
+
+			// Ensure certreq is stored on the object in case it is sent back to remote
+			if localObj.Data, err = anypb.New(certreq); err != nil {
+				log.Error().Err(err).Msg("could not marshal CertificateRequest Any back to remote replica")
+				return nil, status.Error(codes.FailedPrecondition, "replica requires repair")
+			}
```
rebeccabilbro, line 200: #TODO add case for "are you a Peer"?

pkg/gds/replica.go
```diff
@@ -112,46 +118,209 @@ func (r *Replica) Shutdown() error {
...
+		default:
+			// Log error but continue processing, foreign namespace will be ignored
```
rebeccabilbro, line 202: #TODO will we eventually want to figure out how to replicate new namespaces that the local doesn't know about? Maybe leave it to the honu upgrade?

pkg/gds/replica.go
```diff
@@ -112,46 +118,209 @@ func (r *Replica) Shutdown() error {
...
-	// // Step
+				// Ensure certreq is stored on the object in case it is sent back to remote
+				if localObj.Data, err = anypb.New(certreq); err != nil {
+					log.Error().Err(err).Msg("could not marshal CertificateRequest Any back to remote replica")
+					return nil, status.Error(codes.FailedPrecondition, "replica requires repair")
+				}
+			case store.NamespaceIndices:
+				// Ignore indices without warning and don't count as part of local objects
+				nLocalObjs--
+				continue outgoingLoop
```
rebeccabilbro, line 287: #TODO is it a Peer?

pkg/version.go
```diff
@@ -8,10 +8,10 @@ import "fmt"
 // Version component constants for the current build.
 const (
 	VersionMajor         = 1
-	VersionMinor         = 0
+	VersionMinor         = 1
 	VersionPatch         = 0
-	VersionReleaseLevel  = "rc"
-	VersionReleaseNumber = 5
+	VersionReleaseLevel  = "alpha"
+	VersionReleaseNumber = 1
```
rebeccabilbro, line 14: Thanks for remembering to do this!

rebeccabilbro (approved):
Looks great @bbengfort! Thank you for walking me through these updates - fun that we have a good example of lamport scalars to show to our reading group; great timing 😉 

Thanks also for thinking ahead to the gdsutil for migrating the current db over to the new versioned one; that should make it a lot faster for us to deploy when we're ready!

Added a few notes and todos for future tasks, just so we don't forget.

PR #64, opened 2021-07-06
Peer management

rebeccabilbro:
This PR adds preliminary functionality to facilitate management of a network of replicas, including adding, removing, and getting status reports. 

I've implemented the rpc methods such that they work with the current leveldb database, and peers can be managed via functions defined in `replica.go` rather than interacting directly with the database. In this implementation, peers are not really deleted, just tombstoned. This required updating the protocol buffer definitions of Peers to include deletion date and other metadata fields, which is why the pb files have been impacted in this commit. 

Marking as WIP since the code still requires tests to make sure everything works. I also haven't tested out the gdsutil commands yet, since I was hoping we could clarify some of the requirements together before I finalized anything.


pkg/gds/store/leveldb/leveldb.go
```diff
@@ -572,6 +575,166 @@ func (s *Store) DeleteCertReq(id string) (err error) {
...
+			return nil, ErrEntityNotFound
+		}
+		return nil, err
+	}
+
+	p = new(peers.Peer)
+	if err = proto.Unmarshal(val, p); err != nil {
+		return nil, err
+	}
+
```
bbengfort, line 693: We should check for Tombstones here and return `ErrEntityNotFound` if it has been deleted. 

```go
if p.Deleted != "" {
	return nil, ErrEntityNotFound
}
```

pkg/gds/store/leveldb/leveldb.go
```diff
@@ -572,6 +575,166 @@ func (s *Store) DeleteCertReq(id string) (err error) {
...
+// RetrievePeer returns a peer by it's stringified PID.
+func (s *Store) RetrievePeer(id string) (p *peers.Peer, err error) {
+	if id == "" {
+		return nil, ErrEntityNotFound
+	}
+
+	var val []byte
+	if val, err = s.db.Get(s.peerKey(id), nil); err != nil {
+		if err == leveldb.ErrNotFound {
+			return nil, ErrEntityNotFound
```
bbengfort, line 684: Just a note, that the caller can disambiguate not found vs db errors as:

```go
if peer, err = s.RetrievePeer("42"); err != nil {
    if errors.Is(err, ldb.ErrEntityNotFound) {
        return nil, status.Error(codes.NotFound, "peer not found")
    } else {
        return nil, status.Error(codes.FailedPrecondition, "could not fetch peer from db")
    }
}
```

pkg/gds/store/store.go
```diff
@@ -122,6 +124,14 @@ type CertificateStore interface {
 	DeleteCertReq(id string) error
 }
 
+// ReplicaStore describes how the service interacts with and manages Peers.
+type ReplicaStore interface {
```
bbengfort, line 128: In the next PR, we need to add `ReplicaStore` to the `Store` interface on line `106` (before `VersionManager`).

proto/gds/peers/v1/peers.proto
```diff
@@ -32,7 +37,7 @@ service PeerManagement {
 // Used to filter the peers that are returned. If no filters are specified then all
 // known peers on the remote replica are returned.
 message PeersFilter {
-    repeated string region = 1;      // Specify the region(s) to return the peers for.
+    repeated string region = 1;      // Specify the region(s) to return the peers for. Only effects PeersList not PeersStatus
```
bbengfort, line 40: Thank you so much for adding this comment, it's necessary to clarify what the filter does!

proto/gds/peers/v1/peers.proto
```diff
@@ -14,6 +14,11 @@ message Peer {
     string name = 3;                 // optional - a unique, human readable name for the peer
     string region = 4;               // optional? - the region the peer is running in
 
+    // Logging information timestamps
+    string created = 9;
+    string modified = 10;
+    string deleted = 11;
```
bbengfort, line 20: Tombstones and resurrecting peers is a very cool idea!

pkg/gds/replica.go
```diff
@@ -149,3 +153,89 @@ func (r *Replica) Gossip(ctx context.Context, in *global.VersionVectors) (out *g
...
+	if pl, err := r.peerStatus(ctx, ftr); err != nil {
+		return nil, err
+	} else {
+		out = pl.Status
+	}
+	return out, nil
+}
+
+// Helper to get the peer network status
+func (r *Replica) peerStatus(ctx context.Context, in *peers.PeersFilter) (out *peers.PeersList, err error) {
```
bbengfort, line 207: Excellently designed helper function for all three RPC methods!

go.mod
```diff
@@ -7,7 +7,6 @@ require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/ghodss/yaml v1.0.0
-	github.com/golang/protobuf v1.5.2
```
bbengfort, line 10: Nice to remove this older protobuf dependency!

pkg/gds/peers/v1/peers.go
```diff
@@ -0,0 +1,7 @@
+package peers
+
+import "fmt"
+
+func (p *Peer) Key() string {
```
bbengfort, line 5: Thank you so much for adding the `Key` method; this absolutely helps keep the store interface clean while simplifying how we construct a key in the future.

bbengfort (approved):
I really enjoyed going over the Peer management interface; I think we have a neat solution to gossiping Peers across the network, simplifying the devops and sysadmin tasks by a lot. I like that we can easily interact with Peers at runtime rather than dealing with configuration. Excellent work creating yet another store interface and keeping things consistent! Looking forward to merging this in now; we'll let client development and integration testing be the next steps for the next PR.

pkg/gds/store/leveldb/leveldb.go
bbengfort: No action is required; this note was just part of our discussion.

PR #65, opened 2021-07-07
Better object management

bbengfort:
This PR creates a helper pkg, wire for marshaling and unmarshaling
object data from disk and memory - standardizing how we handle objects
for replication. It also adds NamespaceReplicas to ensure that Peers are
also replicated. This PR should delete and simplify a lot of code.

### DIFF
--- a/cmd/gdsutil/main.go
+++ b/cmd/gdsutil/main.go
@@ -2,14 +2,12 @@ package main
 
 import (
 	"bytes"
-	"compress/gzip"
 	"context"
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/hmac"
 	"crypto/sha256"
 	"encoding/base64"
-	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -29,6 +27,7 @@ import (
 	"github.com/trisacrypto/directory/pkg/gds/models/v1"
 	"github.com/trisacrypto/directory/pkg/gds/peers/v1"
 	"github.com/trisacrypto/directory/pkg/gds/store"
+	"github.com/trisacrypto/directory/pkg/gds/store/wire"
 	pb "github.com/trisacrypto/trisa/pkg/trisa/gds/models/v1beta1"
 	"github.com/urfave/cli"
 	"google.golang.org/grpc"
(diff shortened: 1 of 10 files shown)


cmd/gdsutil/main.go
```diff
@@ -58,8 +57,8 @@ func main() {
 					EnvVar: "GDS_DATABASE_URL",
 				},
 				cli.BoolFlag{
-					Name:  "s, stringify",
-					Usage: "stringify keys otherwise they are base64 encoded",
+					Name:  "b, b64encode",
+					Usage: "base64 encode keys (otherwise they will be utf-8 decoded)",
```
bbengfort, line 61: Figured the common case was stringify, so mind as well make that the default!

pkg/gds/store/store.go
```diff
@@ -52,6 +52,7 @@ const (
 	NamespaceCertReqs = "certreqs"
 	NamespaceReplicas = "peers"
 	NamespaceIndices  = "index"
+	NamespaceSequence = "sequence"
```
bbengfort, line 55: This is an artifact of older code, but leaving it in if we want to store ints later.

pkg/gds/store/wire/wire.go
```diff
@@ -0,0 +1,204 @@
...
+		if err = proto.Unmarshal(data, certreq); err != nil {
+			return nil, fmt.Errorf("could not unmarshal %s to %T: %s", namespace, certreq, err)
+		}
+
+		// Fetch the object metadata and add the CertificateRequest data back onto the Any field
+		obj = certreq.Metadata
+		if obj.Data, err = anypb.New(certreq); err != nil {
+			return nil, err
+		}
+	case store.NamespaceReplicas:
```
bbengfort, line 89: Because this handles NamespaceReplicas, I believe that gdsutil will replicate peers as a result.

cmd/gdsutil/main.go
rebeccabilbro: Nice!

cmd/gdsutil/main.go
```diff
@@ -1036,6 +923,10 @@ certreqLoop:
 	iter.Release()
 
 	fmt.Printf("migrated %d CertificateRequest objects\n", migrated)
+
+	// NOTE: currently required to migrate peers since they are already versioned
```
rebeccabilbro, line 927: currently not required?

rebeccabilbro (approved):
Lots of code removal! I really like how we've moved all the different namespace handling stuff to a single file in `wire`; was a little worried that we'd lose track of the different places they were mentioned and now it's all nicely sorted! Thank you @bbengfort !

PR #66, opened 2021-07-08
WIP

frankienicoletti:
Still needs some clean up and to be tested more fully but it runs!


bbengfort:
@frankienicoletti let us know when you'd like us to review; just trying to respect the WIP!

pdeziel (commented):
I realize that this PR is still tagged as a WIP draft but I wanted to give my two cents and figure out what the next steps are. First of all, thank you for putting in the effort of prototyping out the health check service! Although a few things have changed in the main branch since this was created, it's nice to have a good idea of what all the moving parts are. From my perspective, here are the tasks that we need to complete (not necessarily all in the same PR).

1. Update VASP model and DB storage code to work with the main branch. There have been some recent changes which set the pattern for interacting with VASP objects and the underlying DB. I've tried to reflect some of that in the comments. It's also possible that we could use another database for this and not tie it so directly to the VASP data store.
2. Ideally we would want to wrap the `main.go` code into a GDS routine that runs continually as a background service.
3. Create an endpoint which can be requested to retrieve health check data from the data store, formatting it so that it can be displayed in some useful way on the front end.

bbengfort:
@frankienicoletti thank you so much for stubbing this out for us! We'll certainly use this as a reference moving forward but unfortunately we've gotten too far passed where the codebase was in July to easily merge this, so we're going to close the PR. Thanks again!

PR #67, opened 2021-07-15
Add mock db

frankienicoletti:
Usage: 

Anywhere that you need to test but don't want to actually call a database you can use the mock struct instead of Level. When you write the test you create the `On` version of the function and that's what will run during the test. You can also check if the `Invoked` bool is true to ensure it ran (vs not throwing an error because it didn't run). Will be used in #66


bbengfort (approved):
This makes a lot of sense! Thanks for including the interface type check on line 9!

PR #79, opened 2021-08-18
Display Country Name next to Flag Emoji

bbengfort:
Per the request of JJ.


web/src/lib/country.js
```diff
@@ -52,4 +52,265 @@ export function emojiCountryCode(flag) {
   return String.fromCodePoint(...codePoints);
 }
 
+const isoCountries = {
```
bbengfort, line 55: Not exactly the most brilliant solution to this, but it gets the job done.

PR #81, opened 2021-08-31
setup admin ui layout

elysee15:
Capture d’écran 2021-08-31 à 14 34 16


bbengfort (approved):
@elysee15 looks good - this is just a copy of the theme files over to the repository right? I saw a bunch of stuff that we should clean up when we're done with version 1 (e.g. assets and styles that will probably remain unused) and we'll need to manage the metadata of the site -- but those can be different tasks for later. I'll go ahead and merge this in now.

bbengfort:
@elysee15 Thank you for copying the footer from the public UI over!

PR #82, opened 2021-08-31
Logging Severity Audit

bbengfort:
Updates logging severities to correct levels for monitoring and alerts.

Co-authored with @rebeccabilbro


rebeccabilbro (approved):
Thanks @bbengfort ! Marking as approved since we went through this together this morning; looking forward to tracking these new log messages and seeing them come through with our new incident monitoring system on Slack!

PR #83, opened 2021-09-01
feat: add dashboard page

elysee15:
Capture d’écran 2021-08-31 à 23 59 48


bbengfort (approved):
@elysee15 excellent progress! Since this is just the mockup stage and since we don't have a live API yet, I'm not reviewing components in detail, just how the external mockup is looking. Things look very good so far!

PR #84, opened 2021-09-01
feat: add vasps list page

elysee15:
Capture d’écran 2021-09-01 à 09 54 36


bbengfort (approved):
Excellent - is this using data tables to make search and filtering easier? 

We'll need to update the name field, but we can handle that later on when we get feedback from Cynthia.

PR #89, opened 2021-09-06
fix: vasps details

elysee15:


bbengfort:
@elysee15 would you mind rebasing from main?

bbengfort (approved):

PR #92, opened 2021-09-14
feat: Sign-In with Google Button

elysee15:
Capture d’écran 2021-09-14 à 15 24 19


web/gds-admin-ui/src/pages/account/Login.js
```diff
@@ -1,69 +1,38 @@
...
-        history.push("/dashboard")
-    };
-
-    const schemaResolver = yupResolver(
-        yup.object().shape({
-            username: yup.string().required('Please enter Username'),
-            password: yup.string().required('Please enter Password')
-        })
-    );
+    const clientId = process.env.REACT_APP_GOOGLE_CLIENT_ID
```
bbengfort, line 12: Perhaps it would be better to store this in config.js?

web/gds-admin-ui/src/components/Auth/SignWithGoogle.js
```diff
@@ -0,0 +1,32 @@
+
+
+import React from 'react'
+
+function SignWithGoogle({ loginResponse, text, clientId }) {
+
+    React.useEffect(() => {
+        window.onload = function () {
+            if (window.google) {
+                window.google.accounts.id.initialize({
```
bbengfort, line 10: Note that when I was attempting to use these functions even with the correct Client ID - I was getting 403 errors; so we may need to look into that when we get there.

web/gds-admin-ui/src/redux/auth/actions.js
```diff
@@ -0,0 +1,19 @@
+import { AuthActionTypes } from './constants';
+
+export const loginUser = (data) => ({
+    type: AuthActionTypes.LOGIN_USER,
+    payload: data
+});
```
bbengfort, line 6: Is the plan to have this action eventually call the API server with the credentials and then dispatch, or is the dispatch going to happen from the Login component?

bbengfort (commented):
Looks like a good start to the Sign in with Google Button! I like the direct use of the client-library; I used HTML data properties when I prototyped this. I had a few questions, then we can move forward with this PR.

web/gds-admin-ui/public/index.html
```diff
@@ -34,6 +34,8 @@
       The build step will place the bundled scripts into the <body> tag.
 
       To begin the development, run `npm start` or `yarn start`.
-      To create a production bundle, use `npm run build` or `yarn build`.
-    --></body>
+      To create a production bundle, use `npm run build` or `yarn build`. 
+    -->
+        <script src="https://accounts.google.com/gsi/client" async defer></script>
```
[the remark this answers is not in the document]
elysee15: Could you explain me ?

web/gds-admin-ui/src/components/Auth/SignWithGoogle.js
```diff
@@ -0,0 +1,32 @@
+
+
+import React from 'react'
+
+function SignWithGoogle({ loginResponse, text, clientId }) {
+
+    React.useEffect(() => {
+        window.onload = function () {
```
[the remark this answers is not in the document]
elysee15: Yes

web/gds-admin-ui/src/pages/account/Login.js
elysee15: You’re right

web/gds-admin-ui/src/components/Auth/SignWithGoogle.js
elysee15: Okay

web/gds-admin-ui/src/redux/auth/actions.js
elysee15: The dispatch happen from the Login Component and I decode the passed credentials in the saga file

elysee15: To thus dispatch the user info in the Store

web/gds-admin-ui/src/components/Auth/SignWithGoogle.js
elysee15: It's related to ClientId configuration 
`m=credential_page_library:41 [GSI_LOGGER]: The given origin is not allowed for the given client ID.`

web/gds-admin-ui/env.example
```diff
@@ -1 +1,2 @@
-REACT_APP_API_URL=
\ No newline at end of file
+REACT_APP_API_URL=
+REACT_APP_GOOGLE_CLIENT_ID
```
bbengfort, line 2: 👍🏽

web/gds-admin-ui/public/index.html
```diff
@@ -34,6 +35,6 @@
       The build step will place the bundled scripts into the <body> tag.
```
bbengfort, line 35: Note this comment - the bundled scripts will be placed into the body tag - by putting the Google scripts in the head, it will be requested before the React application.

web/gds-admin-ui/src/components/Auth/SignWithGoogle.js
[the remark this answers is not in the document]
bbengfort: I don't want to deal with this now, we have more important things to do; but should we create a story to handle this case, e.g. set a time out and check again in 1500ms?

web/gds-admin-ui/src/components/Auth/SignWithGoogle.js
bbengfort: I have; I'll keep playing with it, but let's not let us block us from moving forward for now since the server doesn't have authentication added yet.

web/gds-admin-ui/src/redux/auth/actions.js
bbengfort: Right, but eventually you'll have to make a request to the server to get the GDS Admin API token; I'm still working on that, but the Google Credential isn't the end, you'll be holding a GDS access and refresh token as the final say for if the login was successful. 

There is a way to configure Google to make the POST request if you'd like, but I don't think that makes sense for a React app.

bbengfort (approved):
There is still more work to do on logging in; but this is a good start. Let's merge this in so you can start working on other PRs.

PR #93, opened 2021-09-15
Shared Secret Configuration for Admin JWT Tokens

bbengfort:
Implements rotating RSA secret keys for use in signing JWT tokens for authentication with the admin API. 

After some research, this PR standardizes on using RSA asymmetric signatures for JWT tokens to allow different services to verify the token using public keys that can be shared. It is my intent that in the future the Admin API will host a non-authenticated endpoint that returns the public keys, and these keys can be specified in the jku field. For now, however, I recommend we simply add a kid field to the header to identify the keys used to generate the tokens.

Speaking of keys - it is important that RSA keys are regularly rotated. To that end, the configuration expects multiple token keys as a map environment variable; e.g. `keyid1:path/to/key1.pem,keyid2:path/tokey2.pem` (thanks [envconfig](https://github.com/kelseyhightower/envconfig)!) Generally speaking each host will have two keys, the current key and the previously rotated key. The host will only generate keys with the new key but can still verify tokens from the previously rotated key if other nodes are behind in configuration. 

Because we need to identify the "latest" key, the keys need to be ordered. Therefore instead of using UUIDs as key identifiers, I've selected [ksuid](https://pkg.go.dev/github.com/segmentio/ksuid@v1.0.4) as the random identifier, a random, universal, time-sortable key created by segmentio. 

To simplify key generation with ksuids, this PR also implements a `gdsutil admin:tokenkey` command to generate a new key for rotation. We should add key rotation every 3 months to our operations for GDS. 

Fixes SC-563


cmd/gdsutil/main.go
```diff
@@ -1110,6 +1132,41 @@ func findCertificateRequest(vaspID string) (cr *models.CertificateRequest, err e
...
+	}
+
+	if err = pem.Encode(f, &pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(key),
+	}); err != nil {
+		return cli.NewExitError(err, 1)
+	}
+
+	fmt.Printf("RSA key id: %s -- saved with PEM encoding to %s\n", keyid, out)
```
bbengfort, line 1166: Note that an RSA private key contains the public key; so we can extract the public key from the private key without having to write it separately.

pkg/gds/config/config.go
```diff
@@ -41,6 +41,12 @@ type AdminConfig struct {
 	Enabled  bool   `split_words:"true" default:"true"`
 	BindAddr string `split_words:"true" default:":4434"`
 	Mode     string `split_words:"true" default:"release"`
+
+	// TokenKeys are the paths to RSA JWT signing keys in PEM encoded format. The
+	// environment variable should be a comma separated list of keyid:path/to/key.pem
+	// Multiple keys are used in order to rotate keys regularly; keyids therefore must
+	// be sortable; in general we prefer to use ksuid for key ids.
+	TokenKeys map[string]string `split_words:"true" required:"true"`
```
bbengfort, line 49: Really quite happy that envconfig handles extracting maps from environment variables, very cool!

pkg/gds/config/config_test.go
```diff
@@ -21,6 +21,7 @@ var testEnv = map[string]string{
 	"GDS_ADMIN_ENABLED":              "true",
 	"GDS_ADMIN_BIND_ADDR":            ":444",
 	"GDS_ADMIN_MODE":                 "debug",
+	"GDS_ADMIN_TOKEN_KEYS":           "1y9fT85qWaIvAAORW7DKxtpz9FB:testdata/key1.pem,1y9fVjaUlsVdFFDUWlvRq2PLkw3:testdata/key2.pem",
```
bbengfort, line 24: This is an example of what an environment variable that is parsed to a map in envconfig should look like; basically a comma or semicolon separated list of key:value pairs.

pkg/gds/config/config_test.go
```diff
@@ -103,43 +105,54 @@ func TestConfig(t *testing.T) {
...
+			if key == envvar {
+				os.Unsetenv(key)
+			} else {
+				setEnv(key)
+			}
+		}
+
+		_, err := config.New()
+		require.Errorf(t, err, "expected %q to be required but no error occurred", envvar)
+	}
```
bbengfort, line 147: Changed the test so that it sets all environment variables to test that all the required variables are in fact required (the previous test only required one of the required environment variables to be required to pass).

cmd/gdsutil/main.go
```diff
@@ -359,6 +364,23 @@ func main() {
...
 		},
+		{
+			Name:     "admin:tokenkey",
+			Usage:    "generate an RSA token key pair and ksuid for JWT token signing",
+			Category: "admin",
+			Action:   generateTokenKey,
+			Flags: []cli.Flag{
+				cli.StringFlag{
+					Name:  "o, out",
+					Usage: "path to write keys out to (optional, will be saved as ksuid.pem by default)",
```
rebeccabilbro, line 375: if the outfile provided is the same file as the one to which the older keys were written, does this append a new key or overwrite the contents of the existing file?

pkg/gds/config/config.go
rebeccabilbro: This is neat! 

So the picture I have in my head is that keys live in a file called `ksuid.pem` (or similar), to which we append a new key every 3 months. 

`envconfig` (routinely? continuously? on reboot?) looks in that file and computes the 2 (?) most recent keys; the newest one, which we use for client token generation and verification, and the second-to-newest one, which we can use for client token verification.

do I have that right?

pkg/gds/config/config_test.go
rebeccabilbro: Ah, this is making me think that I was wrong in thinking we were appending to a single `ksuid.pem` file, and that we should be uniquely naming each pem so that we can instead we give `envconfig` a list of the two most recent pem files we created.  I suppose that means we have to remember the name of the 3-months ago pem file (so that both it and the new one will be included). Do we need to worry about deleting the >3 months ago pems?

rebeccabilbro (approved):
Neat that key rotation is made so straightforward with `envconfig` @bbengfort! 

I'll admit I'm not sure I've understand the mechanics (I added a few specific questions inline), so I might need a bit more explanation to understand how to ensure we are internally using the new key generation mechanism correctly without overwriting or losing track of the prior key record. 

Also, this is probably a 🎃  question, but "key rotation" is actually more akin to "key incrementation" or "key progression", right? Because we will never return or rotate back to a key that is older than 3 months?

PR #95, opened 2021-09-15
fix: remove language dropdown

elysee15:


web/gds-admin-ui/src/components/LanguageDropdown/index.js
```diff
@@ -1,76 +0,0 @@
...
-    };
-
-    return (
-        <Dropdown show={dropdownOpen} onToggle={toggleDropdown}>
-            <Dropdown.Toggle
-                variant="link"
-                id="dropdown-languages"
-                as={Link}
-                to="#"
-                onClick={toggleDropdown}
```
bbengfort, line 54: Perhaps we could get rid of this for now, or comment it out.

web/gds-admin-ui/src/components/LanguageDropdown/index.js
```diff
@@ -1,76 +0,0 @@
...
-                    {Languages.map((lang, i) => {
-                        return (
-                            <Link to="/" className="dropdown-item notify-item" key={i + '-lang'}>
-                                <img src={lang.flag} alt={lang.name} className="me-1" height="12" />{' '}
-                                <span className="align-middle">{lang.name}</span>
-                            </Link>
-                        );
-                    })}
-                </div>
-            </Dropdown.Menu>
```
bbengfort, line 71: Let's also comment out the Dropdown.Menu 

What happens when you comment out things in JSX? Does it get rendered as an html comment or is it omitted from the final bundle?

bbengfort (changes requested):
I'd still like the enFlag there - and we'll want to have the language drop down in the future; so would you mind adding back the component? I was thinking we could simply remove the onclick and the div map.

web/gds-admin-ui/src/components/LanguageDropdown/index.js
elysee15: Yes it is omitted 

```
 output: {
              ecma: 5,
              comments: false,
              // Turned on because emoji and regex is not minified properly using default
              // https://github.com/facebook/create-react-app/issues/2488
              ascii_only: true,
            },
```

That's the ouput in the of TerserPlugin ( js minifier ) webpack config

web/gds-admin-ui/src/components/LanguageDropdown/index.js
elysee15: I'll put in comment

web/gds-admin-ui/src/components/LanguageDropdown/index.js
bbengfort: Excellent, thank you for looking into that!

bbengfort (approved):
Awesome thank you!

PR #96, opened 2021-09-16
GDS Admin API: VASP List Endpoint

bbengfort:
Implements a paginated VASP list endpoint that can be used to fetch
snippet information about a VASP for GDS Admin UI list and summary
views. Future work will allow filtering and other query params on this
endpoint.

Fixes SC567


pkg/gds/admin.go
```diff
@@ -134,12 +134,92 @@ func (s *Admin) setupRoutes() (err error) {
...
+				snippet.VerifiedContacts = append(snippet.VerifiedContacts, key)
+			}
+
+			// Append to list in reply
+			out.VASPs = append(out.VASPs, snippet)
+		}
+	}
+
+	if err = iter.Error(); err != nil {
+		log.Warn().Err(err).Msg("could iterate over vasps in store")
```
tl6kk, line 214: Maybe here it should be could **not** iterate?

tl6kk (approved):
Everything looks good to me other than the small error in the messaging. I'll go ahead and approve this.

pkg/gds/admin.go
bbengfort: good catch!

PR #98, opened 2021-09-16
GDS Admin API: Filter VASP list by status

bbengfort:
Adds a query param to the vasp list endpoint to filter by verification
status. The verification status must match one of the protocol buffer
verification status strings, but is case and whitespace insensitive.

Fixes SC-577

### DIFF
--- a/cmd/gds/main.go
+++ b/cmd/gds/main.go
@@ -292,6 +292,10 @@ func main() {
 					Name:  "s, page-size",
 					Usage: "specify the number of items per page",
 				},
+				cli.StringFlag{
+					Name:  "S, status",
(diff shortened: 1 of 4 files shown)


rebeccabilbro (approved):
Looks good; glad that we can filter by status now! Thank you for the live code review @bbengfort 😄 

Per our discussion, I like the idea of opening a story for next sprint to work on creating a pseudoindex for ldb that would allow us to query by status a bit more efficiently as the number of VASPS continues to increase &mdash; I also think this would be a great thing for our engineering team to do collaboratively, since it's an opportunity to think deeply about different database implementations!

PR #101, opened 2021-09-17
GDS Admin API: VASP Detail Endpoint

bbengfort:
This endpoint provides a mechanism to retrieve the details for a VASP
that are required for the Admin UI - including some computed values and
extra information that are not standard in the pb.VASP record.

Fixes SC-571

TODO:
- [x] create endpoint stub in client and server
- [x] define vasp detail record
- [x] implement admin api endpoint

### DIFF
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -136,6 +136,7 @@ func (s *Admin) setupRoutes() (err error) {
 	v2 := s.router.Group("/v2")
 	v2.GET("/status", s.Status)
 	v2.GET("/vasps", s.ListVASPs)
+	v2.GET("/vasps/:vaspID", s.RetrieveVASP)
 	v2.POST("/vasps/:vaspID/review", s.Review)
 	v2.POST("/vasps/:vaspID/resend", s.Resend)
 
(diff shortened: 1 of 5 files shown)


pkg/gds/admin.go
```diff
@@ -239,6 +242,68 @@ func (s *Admin) ListVASPs(c *gin.Context) {
...
+		return
+	}
+
+	// Create the response to send back
+	out = &admin.RetrieveVASPReply{
+		VerifiedContacts: models.VerifiedContacts(vasp),
+		Traveler:         models.IsTraveler(vasp),
+	}
+	if out.Name, err = vasp.Name(); err != nil {
+		log.Warn().Err(err).Msg("could not get VASP name")
```
rebeccabilbro, line 268: Could we bump this up to an error? If a record got into GDS and doesn't have any name at all, that seems bad

pkg/gds/admin/v2/client_test.go
```diff
@@ -165,6 +165,34 @@ func TestListVASPs(t *testing.T) {
 	require.Equal(t, fixture.Count, out.Count)
 }
 
+func TestRetrieveVASP(t *testing.T) {
+	fixture := &admin.RetrieveVASPReply{}
```
rebeccabilbro, line 169: I think it's a good idea to just use this dummy reply here, since an actual VASP takes up so many lines. Maybe before this line we could have an inline comment with a tinyurl link to the VASP definition from the protocol buffers: https://github.com/trisacrypto/trisa/blob/main/proto/trisa/gds/models/v1beta1/models.proto#L20-L91

rebeccabilbro (approved):
Looks good @bbengfort, thanks for the live review! Just noted inline 2 things that came up in our discussion.

PR #106, opened 2021-09-22
fix: vasp to 2 details column view

elysee15:
https://www.awesomescreenshot.com/video/5333433?key=3c9e266b80d5914f69cb58d35a5b6f
Capture d’écran 2021-09-22 à 15 41 28
Capture d’écran 2021-09-22 à 15 42 29
fe


bbengfort (approved):
I think two columns does make the detail view much more readable; thank you so much for changing the columns and headers as requested!

PR #110, opened 2021-09-24
Better Verified Contacts Handling

bbengfort:
The VASP List endpoint required both verified and unverified contacts to
display badges correctly in the UI. This PR adapts the response to
return a map of contact name to verified bool rather than just a list of
verified contacts.

Unblocks SC-594

### DIFF
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -359,11 +359,7 @@ func (s *Admin) ListVASPs(c *gin.Context) {
 			snippet.Name, _ = vasp.Name()
 
 			// Add verified contacts to snippet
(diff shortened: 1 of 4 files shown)


pdeziel (approved):
Looks good to me, would be nice if we could reference the contact type strings from a single part of the code it's probably out of the scope of this PR.

bbengfort:
@pdeziel you're definitely right - there is a lot of one-off contact handling code that could be simplified in the `models` package. For example, it's very common to loop through all the contacts, but you have to do that one by one. I'll create a story in GDS enhancements for us to do a better job of this.

PR #111, opened 2021-09-24
feat: update vasps list

elysee15:
Capture d’écran 2021-09-24 à 20 25 55

Fixes SC-599


bbengfort (approved):
Well done, I see everything from SC-599 was completed!

Issue #44, opened 2021-06-07
Admin authentication system


rebeccabilbro:
We would like to have a formal authentication system for admins to review registration requests and manage the directory. We need to establish what the formal review process will be.

Subtasks:
- [ ] Update `models.proto` with review of metadata from [Admin API](https://github.com/trisacrypto/directory/blob/ec2c97dfb7cdf8df746be1b0c8851daebf0dacff/proto/gds/models/v1/models.proto#L47-L52)
- [ ] Replace [admin verification token](https://github.com/trisacrypto/directory/blob/main/pkg/gds/server.go#L504-L508) with authentication

PR #113, opened 2021-09-24
Implements Double Cookie Middleware

bbengfort:
Implements the double cookie middleware for CSRF protection in the Admin
API. The CSRF protection relies on a cookie that is HttpOnly, e.g. one
that cannot be read in JavaScript and one that can be. The client
responsibility is to read the cookie that can be read and submit it in
an X-CSRF-TOKEN header with the request.

TODO:

- [x] Finish up tests (needs https test server for Secure cookies) 
- [x] Implement login and refresh stub endpoints that set the cookies

### DIFF
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -136,15 +136,35 @@ func (s *Admin) setupRoutes() (err error) {
 
 	// Add the v2 API routes
 	v2 := s.router.Group("/v2")
+	v2.POST("/authenticate", s.Authenticate)
+	v2.POST("/reauthenticate", admin.DoubleCookie(), s.Reauthenticate)
 	v2.GET("/status", s.Status)
 	v2.GET("/summary", s.Summary)
 	v2.GET("/autocomplete", s.Autocomplete)
 	v2.GET("/vasps", s.ListVASPs)
 	v2.GET("/vasps/:vaspID", s.RetrieveVASP)
-	v2.POST("/vasps/:vaspID/review", s.Review)
-	v2.POST("/vasps/:vaspID/resend", s.Resend)
+	v2.POST("/vasps/:vaspID/review", admin.DoubleCookie(), s.Review)
+	v2.POST("/vasps/:vaspID/resend", admin.DoubleCookie(), s.Resend)
 
 	return nil
+}
+
+func (s *Admin) Authenticate(c *gin.Context) {
+	if err := admin.SetDoubleCookieTokens(c, time.Now().Add(time.Minute*10).Unix()); err != nil {
+		c.JSON(http.StatusInternalServerError, admin.ErrorResponse("could not set cookies"))
+		return
+	}
+
+	c.JSON(http.StatusNotImplemented, admin.ErrorResponse("not implemented yet"))
+}
+
+func (s *Admin) Reauthenticate(c *gin.Context) {
+	if err := admin.SetDoubleCookieTokens(c, time.Now().Add(time.Minute*10).Unix()); err != nil {
+		c.JSON(http.StatusInternalServerError, admin.ErrorResponse("could not set cookies"))
+		return
+	}
+
+	c.JSON(http.StatusNotImplemented, admin.ErrorResponse("not implemented yet"))
 }
 
 // Summary provides aggregate statistics that describe the state of the GDS.
--- a/pkg/gds/admin/v2/csurf.go
+++ b/pkg/gds/admin/v2/csurf.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha256"
 	"encoding/base64"
 	"net/http"
+	"net/url"
 	"time"
 
 	"github.com/gin-gonic/gin"
@@ -36,6 +37,13 @@ func DoubleCookie() gin.HandlerFunc {
 		}
 
 		header := c.GetHeader(CSRFHeader)
+		if header, err = url.QueryUnescape(header); err != nil {
+			log.Warn().Err(err).Msg("could not unescape csrf token")
+			c.JSON(http.StatusBadRequest, ErrorResponse(err))
+			c.Abort()
+			return
+		}
+
 		if cookie != header {
 			log.Warn().Bool("header_exists", header != "").Bool("cookie_exists", cookie != "").Msg("csrf token cookie/header mismatch")
 			c.JSON(http.StatusForbidden, ErrorResponse(ErrCSRFVerification))
--- a/pkg/gds/admin/v2/csurf.go
+++ b/pkg/gds/admin/v2/csurf.go
@@ -1,0 +1,92 @@
+package admin
+
+import (
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/base64"
+	"net/http"
+	"time"
+
+	"github.com/gin-gonic/gin"
+	"github.com/rs/zerolog/log"
+)
+
+// Parameters and names for double-cookie submit CSRF protection
+const (
+	CSRFCookie          = "csrf_token"
+	CSRFReferenceCookie = "csrf_reference_token"
+	CSRFHeader          = "X-CSRF-TOKEN"
+)
+
+// DoubleCookie is Cross Site Request Forgery (CSRF/XSRF) protection middleware that
+// checks the presence of an X-CSRF-TOKEN header containing a cryptographically random
+// token that matches a token contained in the CSRF-TOKEN cookie in the request.
+// Because of the same-origin policy, an attacker cannot access the cookies or scripts
+// of the safe site therefore it will not be able to guess what token to put in the
+// request, thereby protecting from CSRF attacks. Note that Double Cookie CSRF
+// protection requires TLS to prevent MITM attacks.
+func DoubleCookie() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		cookie, err := c.Cookie(CSRFReferenceCookie)
+		if err != nil {
+			log.Warn().Err(err).Msg("no csrf token cookie in request")
+			c.JSON(http.StatusForbidden, ErrorResponse(ErrCSRFVerification))
+			c.Abort()
+			return
+		}
+
+		header := c.GetHeader(CSRFHeader)
+		if cookie != header {
+			log.Warn().Bool("header_exists", header != "").Bool("cookie_exists", cookie != "").Msg("csrf token cookie/header mismatch")
+			c.JSON(http.StatusForbidden, ErrorResponse(ErrCSRFVerification))
+			c.Abort()
+			return
+		}
+
+		c.Next()
+	}
+}
+
+// SetDoubleCookieTokens is a helper function to set cookies on a gin-request.
+func SetDoubleCookieTokens(c *gin.Context, nbf int64) error {
+	// Generate the CSRF token
+	token, err := GenerateCSRFToken()
+	if err != nil {
+		return err
+	}
+
+	// Compute max age from the not before unix timestamp of the access token.
+	maxAge := int((time.Until(time.Unix(nbf, 0))).Seconds())
+
+	// Set the reference cookie
+	c.SetCookie(CSRFReferenceCookie, token, maxAge, "/", "", true, true)
+
+	// Set the csrf token cookie
+	c.SetCookie(CSRFCookie, token, maxAge, "/", "", true, false)
+	return nil
+}
+
+// This is random seed that is meant to present an additional barrier to cryptanalysis
+// and is unique to each process in the network.
+var seed []byte
+
+func GenerateCSRFToken() (_ string, err error) {
+	// If the process seed is not generated, generate it now.
+	if len(seed) == 0 {
+		seed = make([]byte, 16)
+		if _, err = rand.Read(seed); err != nil {
+			return "", err
+		}
+	}
+
+	nonce := make([]byte, 32)
+	if _, err = rand.Read(nonce); err != nil {
+		return "", err
+	}
+
+	sig := sha256.New()
+	sig.Write(seed)
+	sig.Write(nonce)
+
+	return base64.URLEncoding.EncodeToString(sig.Sum(nil)), nil
+}
--- a/pkg/gds/admin/v2/csurf_test.go
+++ b/pkg/gds/admin/v2/csurf_test.go
@@ -1,0 +1,111 @@
+package admin_test
+
+import (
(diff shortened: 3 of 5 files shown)


pkg/gds/secrets/token.go
```diff
@@ -8,15 +8,15 @@ import (
 
 var (
 	chars        = []rune("ABCDEFGHIJKLMNPQRSTUVWXYZabcdefghjkmnpqrstuvwxyz1234567890")
-	specialChars = []rune("#$%&()*+-<=>?@[]^_{|}~")
+	specialChars = []rune("#%&()*+-<=>?@[]^_{}")
```
bbengfort, line 11: The `$` and `|` characters require escape in bash so I removed them for my sanity

pkg/gds/secrets/token.go
```diff
@@ -8,15 +8,15 @@ import (
...
 )
 
 // CreateToken creates a variable length random token that can be used for passwords or API keys.
 func CreateToken(length int) string {
 	random := rand.New(rand.NewSource(time.Now().UnixNano()))
 	var b strings.Builder
 	for i := 0; i < length; i++ {
-		if rand.Float64() <= 0.85 {
+		if random.Float64() <= 0.90 {
```
bbengfort, line 19: And also reduced the number of special chars :-)

pkg/gds/secrets/token.go
```diff
@@ -8,15 +8,15 @@ import (
 
 var (
 	chars        = []rune("ABCDEFGHIJKLMNPQRSTUVWXYZabcdefghjkmnpqrstuvwxyz1234567890")
-	specialChars = []rune("#$%&()*+-<=>?@[]^_{|}~")
+	specialChars = []rune("#%&()*+-<=>?@[]^_{}")
 )
 
 // CreateToken creates a variable length random token that can be used for passwords or API keys.
 func CreateToken(length int) string {
```
bbengfort, line 15: Was thinking about using this or similar to generate the CSRF token, but it is not crypto random secure; it's going to be removed soon anyway in favor of admin logins.

pkg/gds/admin/v2/csurf_test.go
```diff
@@ -0,0 +1,111 @@
...
+	data, err = readJSON(rep)
+	require.NoError(t, err)
+	require.NotContains(t, data, "error")
+	require.Contains(t, data, "success")
+	require.True(t, data["success"].(bool))
+
+	// Check that we have access to the cookie now
+	ep, _ := url.Parse(server.URL)
+	cookies = client.Jar.Cookies(ep)
+	require.Len(t, cookies, 1)
```
bbengfort, line 83: This is currently failing because the test server is http but the cookies are marked as `Secure`, meaning they're only available over https … the cookiejar is doing the right thing by not including them in the Jar. I have to either figure out how to do TLS with a test server (https://blog.dnsimple.com/2017/08/how-to-test-golang-https-services/) or fake the cookie … which is probably not a good thing to do.

bbengfort:
@rebeccabilbro I've marked the story as complete in Sprint 4 and created a story for review in Sprint 5.

pkg/gds/admin/v2/csurf_test.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	// Create the test server with the CSRF protected router
+	server := httptest.NewServer(router)
+
+	// Attempt to make a request to the server that is not CSRF protected
+	req, err := http.NewRequest(http.MethodPost, server.URL+"/action", nil)
+	require.NoError(t, err)
+
+	// Ensure the request is Forbidden
```
rebeccabilbro, line 48: Is it customary to say the request is Forbidden rather than that the response is Forbidden? Just trying to get the hang of front end syntax

pkg/gds/admin/v2/csurf_test.go
```diff
@@ -0,0 +1,175 @@
...
+	rep, err := client.Do(req)
+	require.NoError(t, err)
+	require.Equal(t, http.StatusForbidden, rep.StatusCode)
+
+	// Check the data in the response
+	data, err := readJSON(rep)
+	require.NoError(t, err)
+	require.Contains(t, data, "error")
+	require.Contains(t, data, "success")
```
rebeccabilbro, line 57: Can you help me understand why it's an error and a success here?

pkg/gds/admin/v2/csurf_test.go
```diff
@@ -0,0 +1,175 @@
...
+	rep, err = client.Do(req)
+	require.NoError(t, err)
+	require.Equal(t, http.StatusForbidden, rep.StatusCode)
+
+	// Check the data in the response
+	data, err = readJSON(rep)
+	require.NoError(t, err)
+	require.Contains(t, data, "error")
+	require.Contains(t, data, "success")
+	require.Equal(t, admin.ErrCSRFVerification.Error(), data["error"].(string))
```
rebeccabilbro, line 104: Is it important that this error is indistinguishable from the one on line 58? E.g. we want them to know they got it wrong, but not how they got it wrong?

pkg/gds/admin/v2/csurf_test.go
```diff
@@ -0,0 +1,175 @@
...
```
rebeccabilbro, line 122: Same question here as with lines 58 and 104

pkg/gds/admin/v2/csurf_test.go
```diff
@@ -0,0 +1,175 @@
...
+	rep, err = client.Do(req)
+	require.NoError(t, err)
+	require.Equal(t, http.StatusCreated, rep.StatusCode)
+
+	// Check the data in the response
+	data, err = readJSON(rep)
+	require.NoError(t, err)
+	require.NotContains(t, data, "error")
+	require.Contains(t, data, "success")
+	require.True(t, data["success"].(bool))
```
rebeccabilbro, line 147: If something weird caused the tests to take longer than 10 min, could this fail due to cookie expiration?

rebeccabilbro (approved):
This looks good @bbengfort, and having read the documentation you prepared on CSRF made the review a lot easier; thank you for that! We should remember to send a thank you to Will also 😉 

I did have a few questions (see inline), nothing blocking but would definitely help me understand how the pieces come together 🥧

rebeccabilbro:
Also noting that this will probably resolve #44?

pkg/gds/admin.go
```diff
@@ -136,17 +136,37 @@ func (s *Admin) setupRoutes() (err error) {
...
+
+func (s *Admin) Reauthenticate(c *gin.Context) {
+	if err := admin.SetDoubleCookieTokens(c, time.Now().Add(time.Minute*10).Unix()); err != nil {
+		c.JSON(http.StatusInternalServerError, admin.ErrorResponse("could not set cookies"))
+		return
+	}
+
+	c.JSON(http.StatusNotImplemented, admin.ErrorResponse("not implemented yet"))
+}
+
```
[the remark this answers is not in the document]
bbengfort: Great question - we do; it's SC-343 and I've updated that story to mention these endpoint stubs. Probably I should have done SC-343 first, but I had all that double cookie research done and I didn't want to forget it!

pdeziel (approved):
Looks good, seems like the implementation was fairly straightforward.

pkg/gds/admin/v2/csurf.go
```diff
@@ -0,0 +1,100 @@
...
+		cookie, err := c.Cookie(CSRFReferenceCookie)
+		if err != nil {
+			log.Warn().Err(err).Msg("no csrf token cookie in request")
+			c.JSON(http.StatusForbidden, ErrorResponse(ErrCSRFVerification))
+			c.Abort()
+			return
+		}
+
+		header := c.GetHeader(CSRFHeader)
+		if header, err = url.QueryUnescape(header); err != nil {
```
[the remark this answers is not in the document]
bbengfort: Apologizes in advances for the somewhat lengthy answer, but I wanted to make sure I captured everything to make sure I didn't make a mistake in the code. 

In order to protect URLs so that they can be properly parsed, certain characters are reserved, e.g. `/`, `:`, `.`, `&`, `=`, `?`, whitespace, etc. URL Query Escaping is a somewhat unique way of ensuring a URL is parsed correctly when these characters need to be used in a URL (usually in the query params, e.g. the part that follows the question mark in  `https://example.com/path/to/document.html?key=foo&otherkey=bar`). For historical reasons, unsafe characters are replaced by a "%" and two hexadecimal digits. 

Cookies need to be URL safe*, so many frameworks and browsers URL escape their values. In our case, the token is a base64 encoded string which means it contains `a-z`, `A-Z`, and `0-9`; however that's only 62 characters, two additional characters are needed, which are `+` and `/` in a standard base64 implementation** (which need to be escaped). Base64 encoding also uses a padding character***, `=` if the byte array is not a multiple of 3 bytes, and this character also needs to be escaped. 

Basically what that means is that if we have a token that looks like this: `a-base64-encoded_string=`; it's possible it will be received by the server as `a%2Fbase64%2Fencoded%2Bstring%3D`. The `%` encoding needs to be converted back into unescaped characters in order for us to perform our token comparison otherwise it will fail. 

`*` - this is only partially true and actually not true in our case, but that doesn't stop browsers or HTTP clients from URL escaping cookies; I noticed it using the Go http client in the test cases.

`**` - we're using a URL-safe base64 encoding which uses `_` and `-` as the extra characters which don't need to be escaped; however the `=` padding character is still involved. 

`***` - technically there is something called `RawURLEncoding` that omits padding, but we need a cryptographic equality on our bytes array.

pkg/gds/admin.go
```diff
@@ -136,17 +136,37 @@ func (s *Admin) setupRoutes() (err error) {
...
-	v2.POST("/vasps/:vaspID/review", s.Review)
-	v2.POST("/vasps/:vaspID/resend", s.Resend)
+	v2.POST("/vasps/:vaspID/review", admin.DoubleCookie(), s.Review)
+	v2.POST("/vasps/:vaspID/resend", admin.DoubleCookie(), s.Resend)
 
 	return nil
 }
 
+func (s *Admin) Authenticate(c *gin.Context) {
+	if err := admin.SetDoubleCookieTokens(c, time.Now().Add(time.Minute*10).Unix()); err != nil {
```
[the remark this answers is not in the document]
bbengfort: Excellent question - I think the cookies should have the same expiration as the refresh token since the Authenticate and Reauthenticate methods are the things that set the cookies.  When these methods create the tokens they will use the `NotBefore` property and pass that to the `SetDoubleCookieTokens` function. The configuration for the duration of the tokens is currently a global variable set here: https://github.com/trisacrypto/directory/blob/main/pkg/gds/tokens/tokens.go#L17-L22

pkg/gds/admin/v2/csurf.go
```diff
@@ -0,0 +1,100 @@
...
+// DoubleCookie is Cross Site Request Forgery (CSRF/XSRF) protection middleware that
+// checks the presence of an X-CSRF-TOKEN header containing a cryptographically random
+// token that matches a token contained in the CSRF-TOKEN cookie in the request.
+// Because of the same-origin policy, an attacker cannot access the cookies or scripts
+// of the safe site therefore it will not be able to guess what token to put in the
+// request, thereby protecting from CSRF attacks. Note that Double Cookie CSRF
+// protection requires TLS to prevent MITM attacks.
+func DoubleCookie() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		cookie, err := c.Cookie(CSRFReferenceCookie)
```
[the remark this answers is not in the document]
bbengfort: It is reading the http Request to find a `Cookie` header with the specified name (`csrf_reference_token`) and parsing it into an `*http.Cookie` struct; validating the correctness of the cookie with respect to it's params (such as `Secure`, `MaxAge`, etc.) then returning its value as a string or an error if the Cookie is not found or not valid. If valid, the value of `cookie` should be the token the server specified in `SetCookie` during Authenticate or Reauthenticate. 

I this does bring up an important point - authentication middleware should happen before double cookie middleware since "please login" is a better error then "invalid csrf token".

pkg/gds/admin/v2/csurf_test.go
bbengfort: I suppose my comment is somewhat non-standard. I mean to say that the response from the server to this request has a 403 Forbidden status code and status.

pkg/gds/admin/v2/csurf_test.go
bbengfort: The json response looks like this:

```json
{
    "success": false, 
    "error": "csrf token validation failed"
}
```

So here I'm testing that the data contains the "success" key and then on line 59 that it is false. 

The `success` bool is very much a throwback to some old school Ben development, which I would be happy to chat about during a happy hour with pleasant cocktails :-)

pkg/gds/admin/v2/csurf_test.go
bbengfort: This is definitely a question I go back and forth on. I think that the best practice is that you do want the error to be only as informative as possible for the front end to make a decision (e.g. redirect to login in this case); but not so informative that an attacker use the messages to analyze for vulnerabilities. The idea is that the the logs contain debugging information not the response itself. 

However, better front end errors help debug on the client side and also help find the log messages that may contain more information. The question is what does the front-end do? Does it just console.log the error in the front end or does it display the text to the user? 

Potentially another happy hour discussion?

pkg/gds/admin/v2/csurf.go
```diff
@@ -0,0 +1,100 @@
...
+
+		header := c.GetHeader(CSRFHeader)
+		if header, err = url.QueryUnescape(header); err != nil {
+			log.Warn().Err(err).Msg("could not unescape csrf token")
+			c.JSON(http.StatusBadRequest, ErrorResponse(err))
+			c.Abort()
+			return
+		}
+
+		if cookie != header {
```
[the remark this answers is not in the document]
bbengfort: Great question, I just did some Googling and StackOverflow says:

> You can't set a cookie with an empty string as it will delete the cookie. From the docs: If the value argument is an empty string, or FALSE, and all other arguments match a previous call to setcookie, then the cookie with the specified name will be deleted from the remote client.

If the cookie doesn't exist then the err on L32 will go off. Header definitely can be empty string; I think the `cookie_exists` bool was my just trying to add symmetry to `header_exists` even though it will always be true if we get to this point.

PR #115, opened 2021-09-27
Feat: Connect VASP Detail view to API

elysee15:
Capture d’écran 2021-09-27 à 19 04 56
Capture d’écran 2021-09-27 à 19 05 05
Capture d’écran 2021-09-27 à 19 05 18

Fixes SC-602


bbengfort (approved):
@elysee15 excellent progress; thank you for making those updates. 

I noticed in the screenshots that the route has changed: the vasp detail view is at the url `/vasps/:uuid` instead of `/vasps-summary/:uuid/details` but I don't see that change in the PR, was that in the last PR? 

In SC-603 we should consider what components we should create for this detail form. Would you mind adding a comment to that story with a list of suggestions? For example based on this PR it seems like a `<Contact />` component makes sense to render the technical, legal, administrative, and billing contacts in the `<ContactInfos />` component and to not render contacts that are not in the data. Perhaps a component for the `Issuer` and `Subject` fields of the certificate? Or one for the different `Name` data structures?

elysee15:
@bbengfort  About the route I modified it in the last PR

PR #116, opened 2021-09-27
VASP Verification state change log

pdeziel:
This PR adds some infrastructure for keeping track of VASP verification state changes, by adding a new struct to GDSExtraData and two helper methods: `GetVerificationChangeLog` and `AppendVerificationChangeLog`. The ultimate goal of this is to support an Admin API endpoint which returns a time series representation of VASP verification status.

TODO:
- [x] Add more test cases


proto/gds/models/v1/models.proto
```diff
@@ -50,6 +50,28 @@ message GDSExtraData {
     // Temporary: verification token for light weight authentication for verification
     // TODO: replace with admin API that uses authentication
     string admin_verification_token = 1;
+
+    // Change log which represents previous verification state changes made to a VASP
+    repeated VerificationChangeLogEntry verification_change_log = 2;
```
bbengfort, line 55: What would you think about calling this `AuditLog`? Although primarily this will be used for state changes, I think it would also be used for actions that do not change the state. In the example on line 69 - "resend emails" would look like:

{
  "timestamp": "2021-09-28T07:23:43Z",
  "previous_state:" "EMAIL_VERIFICATION",
  "current_state": "EMAIL_VERIFICATION",
  "description": "resent verification emails to legal and billing contacts",
  "source": "tinker@trisa.io"
}

E.g. we would record all actions relative to the state of the review for the purposes of ensuring that the VASP state was maintained correctly.

proto/gds/models/v1/models.proto
```diff
@@ -50,6 +50,28 @@ message GDSExtraData {
...
+
+// VerificationChangeLogEntry contains information about a single verification state
+// change for a VASP.
+message VerificationChangeLogEntry {
+    // RFC3339 timestamp
+    string timestamp = 1;
+
+    // Previous verification state (handled internally) and current verification state
+    uint32 previous_state = 2;
+    uint32 current_state = 3;
```
bbengfort, line 66: Storing these states as a `uint32` would certainly work, but would require additional parsing by the application. I recommend we simply store these as a `trisa.gds.models.v1beta1.VerificationState`; at the top of this protobuf file add an import to `import "trisa/gds/models/v1beta1/models.proto";` -- the `go generate` [command](https://github.com/trisacrypto/directory/blob/main/pkg/gds/models/v1/generate.go#L3) appears already setup to include the TRISA files. 

See https://github.com/trisacrypto/directory/blob/main/proto/gds/admin/v1/admin.proto for an example of the use of an imported VerificationState

pkg/gds/models/v1/models.go
```diff
@@ -45,6 +45,51 @@ func SetAdminVerificationToken(vasp *pb.VASP, token string) (err error) {
...
+	}
+	return extra.GetVerificationChangeLog(), nil
+}
+
+// Append a VerificationChangeLogEntry to the extra data on the VASP record.
+func AppendVerificationChangeLog(vasp *pb.VASP, entry *VerificationChangeLogEntry) (err error) {
+	// Validate current state.
+	if entry.CurrentState > uint32(pb.VerificationState_ERRORED) {
+		return fmt.Errorf("cannot set verification state to unsupported value %d", entry.CurrentState)
+	}
```
bbengfort, line 68: Nice! Very thoughtful

pkg/gds/models/v1/models.go
```diff
@@ -45,6 +45,51 @@ func SetAdminVerificationToken(vasp *pb.VASP, token string) (err error) {
...
+
+	// Unmarshal previous extra data.
+	extra := &GDSExtraData{}
+	if vasp.Extra != nil {
+		if err = vasp.Extra.UnmarshalTo(extra); err != nil {
+			return fmt.Errorf("could not deserialize previous extra: %s", err)
+		}
+	}
```
bbengfort, line 76: If `vasp.Extra == nil` you'll need to `make` the array of audit logs; e.g. 

```go
} else {
    extra.VerificationChangeLog = make([]*VerificationChangeLogEntry, 0, 1) 
}
```

I've made the slice with a capacity of 1 since you know you'll be appending something shortly.

pkg/gds/models/v1/models.go
```diff
@@ -45,6 +45,51 @@ func SetAdminVerificationToken(vasp *pb.VASP, token string) (err error) {
...
+
+	// Set the previous state for the new entry.
+	if len(extra.VerificationChangeLog) > 0 {
+		entry.PreviousState = extra.VerificationChangeLog[len(extra.VerificationChangeLog)-1].CurrentState
+	}
```
bbengfort, line 81: This is a very smart way to determine the previous state. I am concerned that if the state is changed without an audit entry by mistake, then this will create an error - so I would want the caller to ensure the entry is correct and includes the previous state with the input. So perhaps we could make this optional? e.g.

```go
if entry.PreviousState == 0 && len(extra.VerificationChangeLog) > 0 {}
``` 

That way this will not overwrite the caller's entry.

pkg/gds/models/v1/models_test.go
```diff
@@ -26,6 +27,32 @@ func TestVASPExtra(t *testing.T) {
...
 	require.Equal(t, "pontoonboatz", token)
+
+	// Attempt to append to verification change log
+	entry := &VerificationChangeLogEntry{
+		Timestamp:    time.Now().Format(time.RFC3339),
+		CurrentState: uint32(pb.VerificationState_VERIFIED),
+		Description:  "automated",
+		Source:       "pontoon@boatz.com",
+	}
+	err = AppendVerificationChangeLog(vasp, entry)
```
bbengfort, line 38: Did this test not panic without the `make([]*VerificationLogEntry, 0, 1)`?

pkg/gds/models/v1/models_test.go
```diff
@@ -26,6 +27,32 @@ func TestVASPExtra(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Should be able to fetch the new change log
+	changeLog, err := GetVerificationChangeLog(vasp)
+	require.NoError(t, err)
+	require.Len(t, changeLog, 1)
+	//TODO: Is there a better way to compare protobuf objects?
```
bbengfort, line 45: If you want to make sure the protocol buffers have the same serial representation (e.g. all the fields are equal) you can use [`proto.Equal`](https://pkg.go.dev/google.golang.org/protobuf/proto#Equal)

bbengfort (commented):
Looks good so far! What were you thinking in terms of extra tests? 

I'd also recommend implementing an `UpdateState(vasp *pb.VASP, state *pb.VerificationState)` method - this function would have access to the previous state of the vasp before changing it's current state and it could automatically create the log entry.

proto/gds/models/v1/models.proto
pdeziel: Good suggestion, I think that name fits better.

proto/gds/models/v1/models.proto
pdeziel: Thanks, I like the idea of using the actual VerificationState.

pdeziel:
@bbengfort I've addressed your comments and added some test cases, so the PR is ready for a second look.

pkg/gds/models/v1/models_test.go
bbengfort: Neat - I learned something new!

pdeziel:
Also, I'm planning on adding the `UpdateState` function in another PR since there's a separate story for it already.

bbengfort (approved):
Looks great, thank you for making those changes! The tests are very well done - excellent coverage!

bbengfort:
@pdeziel your commits don't appear to be signed; do you have GPG keys setup in GitHub?

PR #118, opened 2021-09-28
Implements Authentication and Reauthentication endpoints

bbengfort:
Implements Authentication and Reauthentication endpoints.

Authentication requires a Google Identity Services token, which is verified using the [`idtoken`](https://pkg.go.dev/google.golang.org/api/idtoken#Validate) package. If it is a valid Google token, then access and refresh tokens are issued. 

Reauthentication requires the access token in the header, even if it is expired and the refresh token submitted in the body. Once the access and refresh tokens are compared and verified new tokens are returned. 

Both Authentication and Reauthentication requires double cookie CSRF protection since they are POST endpoints that modify state on the server. Therefore I've also added a `GET /v2/authenticate` method that is handled by `ProtectAuthenticate` - this function sets the cookies on the front-end. The browser must do a GET request to this endpoint before rendering the login button component to set the cookies. It can then POST the client credentials. Note that this is because we're using a javascript callback; if we were using a Google Redirect, Google itself would POST double cookies for verification. See https://developers.google.com/identity/gsi/web/guides/verify-google-id-token for more. 

The authentication and reauthentication endpoints reset the double cookies to expire when the refresh token does. 

TODO: 
- [x] test to see if this actually works with Google credentials
- [ ] review audience and subject verification pieces
- [ ] not exactly sure how to write tests for this
- [x] perform authorized domains check

### DIFF
--- a/containers/docker-compose.yaml
+++ b/containers/docker-compose.yaml
@@ -84,7 +84,7 @@ services:
       args:
         REACT_APP_GDS_API_ENDPOINT: http://localhost:4434/
         REACT_APP_GDS_IS_TESTNET: "true"
-        REACT_APP_GDS_CLIENT_ID: "TODO"
+        REACT_APP_GDS_CLIENT_ID: ${REACT_APP_GDS_CLIENT_ID}
     image: trisa/gds-admin-ui
     ports:
       - 3001:80
--- a/containers/docker-compose.yaml
+++ b/containers/docker-compose.yaml
@@ -24,6 +24,7 @@ services:
       - GDS_ADMIN_ENABLED=true
       - GDS_ADMIN_BIND_ADDR=:4434
       - GDS_ADMIN_TOKEN_KEYS="1y9jUjVqRqRJaiKsNqrZmQkhTqe:/data/creds/current_token_key.pem,1y9jTDgrljWl0iVUOQBcCev7rBG:/data/creds/rotated_token_key.pem"
+      - GDS_ADMIN_AUDIENCE
       - GDS_REPLICA_ENABLED=false
       - GDS_REPLICA_BIND_ADDR=:4435
       - GDS_REPLICA_PID=8
@@ -82,6 +83,7 @@ services:
       args:
         REACT_APP_GDS_API_ENDPOINT: http://localhost:4434/
         REACT_APP_GDS_IS_TESTNET: "true"
+        REACT_APP_GDS_CLIENT_ID: "TODO"
     image: trisa/gds-admin-ui
     ports:
       - 3001:80
--- a/containers/docker-compose.yaml
+++ b/containers/docker-compose.yaml
@@ -25,6 +25,7 @@ services:
       - GDS_ADMIN_BIND_ADDR=:4434
       - GDS_ADMIN_TOKEN_KEYS="1y9jUjVqRqRJaiKsNqrZmQkhTqe:/data/creds/current_token_key.pem,1y9jTDgrljWl0iVUOQBcCev7rBG:/data/creds/rotated_token_key.pem"
       - GDS_ADMIN_AUDIENCE
+      - GDS_ADMIN_AUTHORIZED_DOMAINS="trisa.io,rotational.io"
       - GDS_REPLICA_ENABLED=false
       - GDS_REPLICA_BIND_ADDR=:4435
       - GDS_REPLICA_PID=8
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -13,7 +13,9 @@ import (
 	ginzerolog "github.com/dn365/gin-zerolog"
 	"github.com/gin-contrib/cors"
 	"github.com/gin-gonic/gin"
+	"github.com/golang-jwt/jwt/v4"
 	"github.com/rs/zerolog/log"
+	"google.golang.org/api/idtoken"
 	"google.golang.org/protobuf/encoding/protojson"
 
 	"github.com/trisacrypto/directory/pkg"
@@ -21,15 +23,22 @@ import (
 	"github.com/trisacrypto/directory/pkg/gds/config"
 	"github.com/trisacrypto/directory/pkg/gds/models/v1"
 	"github.com/trisacrypto/directory/pkg/gds/store"
+	"github.com/trisacrypto/directory/pkg/gds/tokens"
 	pb "github.com/trisacrypto/trisa/pkg/trisa/gds/models/v1beta1"
 )
 
 // NewAdmin creates a new GDS admin server derived from a parent Service.
 func NewAdmin(svc *Service) (a *Admin, err error) {
+	// Define the base admin server
 	a = &Admin{
 		svc:  svc,
 		conf: &svc.conf.Admin,
 		db:   svc.db,
+	}
+
+	// Create the token manager
+	if a.tokens, err = tokens.New(a.conf.TokenKeys); err != nil {
+		return nil, err
 	}
 
 	// Create the router
@@ -58,12 +67,13 @@ func NewAdmin(svc *Service) (a *Admin, err error) {
 // performing secure commands with authentication.
 type Admin struct {
 	sync.RWMutex
-	svc     *Service            // The parent Service the admin server uses to interact with other components
-	srv     *http.Server        // The HTTP server that listens on its own independent port
-	conf    *config.AdminConfig // The admin server specific configuration (alias to s.svc.conf.Admin)
-	db      store.Store         // Database connection for loading objects (alias to s.svc.db)
-	router  *gin.Engine         // The HTTP handler and associated middleware
-	healthy bool                // application state of the server
+	svc     *Service             // The parent Service the admin server uses to interact with other components
+	srv     *http.Server         // The HTTP server that listens on its own independent port
+	conf    *config.AdminConfig  // The admin server specific configuration (alias to s.svc.conf.Admin)
+	tokens  *tokens.TokenManager // A token manager that signs JWT tokens with RSA keys
+	db      store.Store          // Database connection for loading objects (alias to s.svc.db)
+	router  *gin.Engine          // The HTTP handler and associated middleware
+	healthy bool                 // application state of the server
 }
 
 // Serve GRPC requests on the specified address.
@@ -136,7 +146,8 @@ func (s *Admin) setupRoutes() (err error) {
 
 	// Add the v2 API routes
 	v2 := s.router.Group("/v2")
-	v2.POST("/authenticate", s.Authenticate)
+	v2.GET("/authenticate", s.ProtectAuthenticate)
+	v2.POST("/authenticate", admin.DoubleCookie(), s.Authenticate)
 	v2.POST("/reauthenticate", admin.DoubleCookie(), s.Reauthenticate)
 	v2.GET("/status", s.Status)
 	v2.GET("/summary", s.Summary)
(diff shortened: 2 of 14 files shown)


containers/docker-compose.yaml
```diff
@@ -82,6 +83,7 @@ services:
       args:
         REACT_APP_GDS_API_ENDPOINT: http://localhost:4434/
         REACT_APP_GDS_IS_TESTNET: "true"
+        REACT_APP_GDS_CLIENT_ID: "TODO"
```
bbengfort, line 86: I need to Google how to set an environment variable as a build arg in docker compose.

pkg/gds/admin.go
```diff
@@ -149,22 +160,193 @@ func (s *Admin) setupRoutes() (err error) {
 	return nil
 }
 
+// Set the maximum age of authentication protection cookies.
+const protectAuthenticateMaxAge = time.Minute * 10
```
bbengfort, line 164: The time that we allow people to finish logging in - open to discussion.

pkg/gds/admin.go
```diff
@@ -149,22 +160,193 @@ func (s *Admin) setupRoutes() (err error) {
...
+		c.JSON(http.StatusUnauthorized, admin.ErrorResponse("invalid credentials"))
+		return
+	}
+
+	// Validate the credential with Google
+	if claims, err = idtoken.Validate(c.Request.Context(), in.Credential, s.conf.Audience); err != nil {
+		log.Warn().Err(err).Msg("invalid credentials used for authentication")
+		c.JSON(http.StatusUnauthorized, admin.ErrorResponse("invalid credentials"))
+		return
+	}
```
bbengfort, line 212: This function ensures that the token is valid, actually from Google, not expired, and that it matches our Google Cloud project credentials.

bbengfort: Still TODO: validate our own claims; e.g. is this a user with a valid admin domain?

pkg/gds/admin.go
```diff
@@ -149,22 +160,224 @@ func (s *Admin) setupRoutes() (err error) {
...
+		c.JSON(http.StatusInternalServerError, admin.ErrorResponse("could not authenticate with credentials"))
+		return
+	}
+
+	// Refresh the double cookies for CSRF protection while using the access/refresh tokens
+	expiresAt := refreshToken.Claims.(*tokens.Claims).ExpiresAt
+	if err := admin.SetDoubleCookieTokens(c, expiresAt); err != nil {
 		c.JSON(http.StatusInternalServerError, admin.ErrorResponse("could not set cookies"))
 		return
 	}
```
pdeziel, line 377: Could potentially wrap this into its own function (`createAuthReply`?) since it looks like the `Authenticate` and `Reauthenticate` both do this after validation.

web/gds-admin-ui/src/pages/account/Login.js
```diff
@@ -8,6 +8,10 @@ import SignWithGoogle from '../../components/Auth/SignWithGoogle';
 import { loginUser } from '../../redux/auth/actions';
 import config from "../../config";
 
+// HACK: temporary code, do not commit!
+import { APICore } from "../../helpers/api/apiCore";
+const api = new APICore();
+
```
pdeziel, line 14: Did you mean to include this in the PR?

pdeziel (commented):
Looks good so far. Thanks for making the diagram, it really helped me understand the authentication process better!

web/gds-admin-ui/src/pages/account/Login.js
bbengfort: Great catch, I did not!

pkg/gds/admin.go
bbengfort: That's a good idea - the only tricky bit is the requirement to get the `refreshToken.Claims` before setting the token cookies, but I could have that function return the timestamp as well.

pkg/gds/admin.go
bbengfort: The Google JWT token contains an `aud` value (JWT for "Audience") that is from the API credentials set up step in the Google Project. Both the front-end and the backend need to know this variable in order to engage the OAuth process. Audience here just verifies that the front end and the backend are using the same credentials, it does not provide any extra security.

pkg/gds/admin.go
bbengfort: I don't think there is a convention for this; or at least I haven't found one. 10 min is as good a guess as any I suppose.

pkg/gds/admin.go
```diff
@@ -149,22 +160,224 @@ func (s *Admin) setupRoutes() (err error) {
...
-	c.JSON(http.StatusNotImplemented, admin.ErrorResponse("not implemented yet"))
+	// Return successful authentication!
+	c.JSON(http.StatusOK, out)
 }
 
+func (s *Admin) checkAuthorizedDomain(claims *idtoken.Payload) error {
+	// Fetch the claim from the idtoken payload
+	domain, ok := claims.Claims["hd"]
+	if !ok {
+		return errors.New("no hd claim to verify authorized domain with")
```
rebeccabilbro, line 262: Did we ever figure out what "hd" stands for?

pkg/gds/admin.go
```diff
@@ -149,22 +160,224 @@ func (s *Admin) setupRoutes() (err error) {
...
+
+	// Search the authorized domains, if found return nil
+	for _, authorized := range s.conf.AuthorizedDomains {
+		if domains == authorized {
+			// Found an authorized domain!
+			return nil
+		}
+	}
+
+	return fmt.Errorf("%s is not in the configured authorized domains", domains)
```
rebeccabilbro, line 279: Is the idea that if this error comes up, the UI should consume it and prompt the user to log in with a different google account?

pkg/gds/admin.go
```diff
@@ -149,22 +160,224 @@ func (s *Admin) setupRoutes() (err error) {
...
+	// signed correctly by the token server.
+	if accessClaims, err = s.tokens.Parse(tks); err != nil {
+		log.Warn().Err(err).Msg("reauthenticate called with invalid access token")
+		c.JSON(http.StatusUnauthorized, admin.ErrorResponse("request is not authorized"))
+		return
+	}
+
+	// Parse incoming JSON data from the client request (contains refresh token)
+	in = new(admin.AuthRequest)
+	if err = c.ShouldBind(&in); err != nil {
```
rebeccabilbro, line 318: What is the role of "binding" here in the context of the auth/auth flow diagram you shared?

pkg/gds/admin.go
```diff
@@ -149,22 +160,224 @@ func (s *Admin) setupRoutes() (err error) {
...
+		log.Warn().Msg("mismatched access and refresh token pair")
+		c.JSON(http.StatusUnauthorized, admin.ErrorResponse("invalid credentials"))
+		return
+	}
+
+	// At this point we've validated the reauthentication and are ready to reissue tokens
+	// Create the access and refresh tokens from the claims
+	if accessToken, err = s.tokens.CreateAccessToken(accessClaims); err != nil {
+		log.Error().Err(err).Msg("could not create access token")
+		c.JSON(http.StatusInternalServerError, admin.ErrorResponse("could not authenticate with credentials"))
```
rebeccabilbro, line 349: What are some of the scenarios that would cause reauthentication to fail on line 347 if the access and refresh tokens comparison succeeds on 339?

Do we want our log messages on lines 348, 354, 362, 367 to reflect that this is a reauthentication fail vs a first-time authentication fail?

pkg/gds/admin/v2/auth.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"errors"
+	"strings"
+
+	"github.com/gin-gonic/gin"
+)
+
+const (
+	bearer        = "Bearer "
```
rebeccabilbro, line 11: Just checking that the extra whitespace char was intentional

pkg/gds/tokens/tokens_test.go
```diff
@@ -195,6 +195,61 @@ func (s *TokenTestSuite) TestKeyRotation() {
...
+	// Modify claims to be expired
+	claims := accessToken.Claims.(*tokens.Claims)
+	claims.IssuedAt = time.Unix(claims.IssuedAt, 0).Add(-24 * time.Hour).Unix()
+	claims.ExpiresAt = time.Unix(claims.ExpiresAt, 0).Add(-24 * time.Hour).Unix()
+	claims.NotBefore = time.Unix(claims.NotBefore, 0).Add(-24 * time.Hour).Unix()
+	accessToken.Claims = claims
+
+	// Create signed token
+	tks, err := tm.Sign(accessToken)
+	require.NoError(err, "could not create expired access token from cliams")
```
rebeccabilbro, line 226: cliams => claims

web/gds-admin-ui/src/pages/account/Login.js
rebeccabilbro, line 14: Flagging this for you 😉

web/gds-admin-ui/src/pages/account/Login.js
```diff
@@ -18,7 +22,11 @@ const Login = (): React$Element<any> => {
 
     const handleCredentialResponse = (response) => {
         if (response.credential) {
-            dispatch(loginUser(response.credential))
+            // HACK: temporary code, do not commit!
+            console.log(response.credential);
+            const rep = api.create("/login", {credential: response.credential});
+            console.log(rep);
+            dispatch(loginUser(response.credential));
```
rebeccabilbro, line 29: And this 😉

containers/docker-compose.yaml
rebeccabilbro: This might help? https://stackoverflow.com/a/52430444/6552250

rebeccabilbro (approved):
This looks good @bbengfort , thank you for the walkthrough this morning and the diagram for explanation; I had a few small questions (inline), but I don't see any reason not to merge this in and move on to the next step of integrating into the Admin UI!

containers/docker-compose.yaml
bbengfort: That did it - thanks!

pkg/gds/config/config.go
```diff
@@ -38,15 +38,17 @@ type GDSConfig struct {
 }
 
 type AdminConfig struct {
-	Enabled  bool   `split_words:"true" default:"true"`
-	BindAddr string `split_words:"true" default:":4434"`
-	Mode     string `split_words:"true" default:"release"`
+	Enabled           bool     `split_words:"true" default:"true"`
+	BindAddr          string   `split_words:"true" default:":4434"`
+	Mode              string   `split_words:"true" default:"release"`
+	Audience          string   `split_words:"true"`
```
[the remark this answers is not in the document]
bbengfort: Yes!

pkg/gds/admin.go
```diff
@@ -126,17 +136,18 @@ func (s *Admin) setupRoutes() (err error) {
 	// Add CORS configuration
 	// TODO: configure origins from the environment rather than hard-coding
 	s.router.Use(cors.New(cors.Config{
-		AllowAllOrigins:  true,
+		AllowOrigins:     []string{"http://localhost", "http://localhost:3000", "http://localhost:3001"},
```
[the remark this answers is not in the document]
bbengfort: SC-1059 is created

pkg/gds/admin.go
bbengfort: Just looked, it turns out it stands for "hosted domain": https://cloud.google.com/iap/docs/signed-headers-howto#verifying_the_jwt_payload

pkg/gds/admin.go
bbengfort: Yes, e.g. if you accidentally signed in with your @gmail.com account.

pkg/gds/admin.go
bbengfort: In the gin framework, binding is populating the `admin.AuthRequest` struct from the request; e.g. reading the raw JSON from the request body and unmarshaling. So this is what the server is doing to read the `{credential: [Google JWT credential]}` payload.

pkg/gds/admin.go
bbengfort: If we get to line 347 then something is wrong on the server; e.g. the token RSA keys are bad, something is wrong with the configuration, or a developer error occurred. 348, 354, 362, and 667 return an 500 error (something went really wrong) instead of a 401 error (you're not authorized) and the log level increases from `Warn()` to `Error()`, and hopefully we'll receive the Slack alert and fix it! 

So to answer your question - it would be a pretty narrow edge case, for example the user authenticates then we restart the server with a new configuration and then the user attempts to reauthenticate. However, your point about messaging which method we're in is a good one as that may help debugging - so I'll changing the wording!

pkg/gds/admin/v2/auth.go
bbengfort: It is intentional in this case, but good catch!

web/gds-admin-ui/src/pages/account/Login.js
bbengfort: Thank you!

PR #119, opened 2021-09-29
VASP update state change refactor

pdeziel:
This adds a new function `UpdateVerificationStatus` which updates the verification state of a VASP and appends an entry to the audit log to record the state change. State changes have been refactored to use this new function instead of just assigning a new value to `VASP.VerificationStatus`.

### DIFF
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -753,16 +753,34 @@ func (s *Admin) Review(c *gin.Context) {
 		return
 	}
 
+	// Retrieve user claims for access to provided user info
+	var claims *tokens.Claims
+	value, exists := c.Get(admin.UserClaims)
+	if exists && value != nil {
+		var ok bool
+		claims, ok = value.(*tokens.Claims)
+		if !ok {
+			err = fmt.Errorf("claims is an incorrect type, expecting *tokens.Claims found %T", value)
+			log.Error().Err(err).Msg("could not retrieve user claims")
+			c.JSON(http.StatusInternalServerError, admin.ErrorResponse("unable to retrieve user info"))
+			return
+		}
+	} else {
+		log.Error().Err(fmt.Errorf("no user claims in context")).Msg("could not retrieve user claims")
+		c.JSON(http.StatusInternalServerError, admin.ErrorResponse("unable to retrieve user info"))
+		return
+	}
+
 	// Accept or reject the request
 	out = &admin.ReviewReply{}
 	if in.Accept {
-		if out.Message, err = s.acceptRegistration(vasp); err != nil {
+		if out.Message, err = s.acceptRegistration(vasp, claims); err != nil {
 			log.Error().Err(err).Msg("could not accept VASP registration")
 			c.JSON(http.StatusInternalServerError, admin.ErrorResponse("unable to accept VASP registration request"))
 			return
 		}
 	} else {
-		if out.Message, err = s.rejectRegistration(vasp, in.RejectReason); err != nil {
+		if out.Message, err = s.rejectRegistration(vasp, in.RejectReason, claims); err != nil {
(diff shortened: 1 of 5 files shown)


pkg/gds/admin.go
```diff
@@ -561,7 +561,10 @@ func (s *Admin) acceptRegistration(vasp *pb.VASP) (msg string, err error) {
 		return "", err
 	}
 	vasp.VerifiedOn = time.Now().Format(time.RFC3339)
-	vasp.VerificationStatus = pb.VerificationState_REVIEWED
+	// TODO: Replace "email" in source parameter with user email address.
```
bbengfort, line 564: I'm working on the authentication middleware right now that will add the user email to the request context. Would you prefer to hold off on finishing this PR until that's merged and then rebase with it to implement it, or create a story to do this later and merge this PR without waiting for the middleware?

pkg/gds/gds.go
```diff
@@ -113,9 +113,13 @@ func (s *GDS) Register(ctx context.Context, in *api.RegisterRequest) (out *api.R
 		VaspCategories:      in.VaspCategories,
 		EstablishedOn:       in.EstablishedOn,
 		Trixo:               in.Trixo,
-		VerificationStatus:  pb.VerificationState_SUBMITTED,
+		VerificationStatus:  pb.VerificationState_NO_VERIFICATION,
```
bbengfort, line 116: This is also the zero value so this entire line could be safely omitted, but I like that you've added it here explicitly for clarity.

pkg/gds/models/v1/models.go
```diff
@@ -45,6 +46,27 @@ func SetAdminVerificationToken(vasp *pb.VASP, token string) (err error) {
...
+		CurrentState:  state,
+		Description:   description,
+		Source:        source,
+	}
+	if err := AppendAuditLog(vasp, entry); err != nil {
+		return err
+	}
+
+	// Set the new state on the VASP.
+	vasp.VerificationStatus = state
```
bbengfort, line 65: I have a slight preference that the verification status gets changed before the audit log is appended, just in case we typo the previous state guard in `AppendAuditLog` but it's not a big deal.

bbengfort (approved):
Well done! Thank you for so thoroughly searching through the code to find all the VASP state changes! I really liked the descriptions you added and you were spot on with "email" vs "automated" as the source. 

There is one place that I would also like to add an audit trail; but you wouldn't have been able to find it easily since it doesn't actually change the VASP state. I'd like to track when emails are "resent" to the users:

https://github.com/trisacrypto/directory/blob/main/pkg/gds/admin.go#L901

My thought is that we would add a different audit message to each one of the cases in the `switch` for the email to send. The previous and current states will be the same. 

Note that many of the state changes imply emails being sent; however if it's easier, we could add auditing to the email manager so we know exactly what emails were sent. If this is your preference, then let's create a story for that and do it in a different PR.

pkg/gds/admin.go
pdeziel: Seems like a simple enough change so I'll hold off on merging this PR until the middleware one is done.

pdeziel:
@bbengfort Thanks for getting to the review so quickly! I'll hold onto this PR for now until I can integrate the user email. As for your other question, I'm leaning towards creating another story for the email audit log and handling that in another PR.

pdeziel:
@bbengfort I've rebased and added the email from the context, mind taking a quick look to verify that the rebase is correct and the claims error handling is ok?

bbengfort (approved):
Thank you for holding off until you had access to the user claims from the context! I rebased this with main since there was a merge conflict; would you mind taking a look to make sure nothing got dropped? Feel free to merge when you're ready.

PR #120, opened 2021-09-29
Authorization Middleware

bbengfort:
Adds middleware that checks the Bearer JWT token from the Authorization
header before permitting the request to continue.

TODO:

- [x] Add middleware to all endpoints but the `authenticate` and `reauthenticate` endpoints 
- [x] Determine if authorization should happen before double cookie verification
- [x] Add claims to the context so that the request has access to the user email address for auditing
- [x] Create unit tests for the middleware 
- [ ] Verify the endpoints generally work (we'll need to work on integration tests in the next epic)

### DIFF
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -132,6 +132,7 @@ func (s *Admin) setupRoutes() (err error) {
 	// Application Middleware
 	s.router.Use(ginzerolog.Logger("gin"))
 	s.router.Use(gin.Recovery())
+	s.router.Use(s.Available())
 
 	// Add CORS configuration
 	// TODO: configure origins from the environment rather than hard-coding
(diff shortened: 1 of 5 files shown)


pkg/gds/admin.go
```diff
@@ -132,6 +132,7 @@ func (s *Admin) setupRoutes() (err error) {
 	// Application Middleware
 	s.router.Use(ginzerolog.Logger("gin"))
 	s.router.Use(gin.Recovery())
+	s.router.Use(s.Available())
```
pdeziel, line 135: Any particular reason why this was moved? Does the order of these statements matter or is it mainly cosmetic?

pdeziel (approved):
Looks good to me, are there any other security measures we are planning to add to the authorization middleware?

pkg/gds/admin.go
bbengfort: I think it actually does matter; the order of the middleware application determines the order middleware is applied.

bbengfort:
Thanks @pdeziel! We will probably continue to refine our security mechanisms but for now this should be enough to get us to an MVP!

PR #121, opened 2021-09-29
fix: connect dashboard reviews summary to filtered VASP list API

elysee15:
I just modified the request endpoint that get Vasps with `pending review` status to use `GET v2/vasps?status=review+pending`


bbengfort (approved):
Very nice work indeed. I presume that for the VASP list functionality the query is omitted so it returns all VASPs. Well done!

PR #122, opened 2021-09-29
fix: hide Mock Timeseries Graph if Production Build

elysee15:


bbengfort (approved):
Excellent - we'll make sure that we set the correct environment variables when we deploy the MVP. Please rebase and then we'll merge!

bbengfort (approved):
Thank you - this will help us deploy the MVP today with a little less noise on the front end! We can then release the completed time series with the connection to the API all at once!

PR #123, opened 2021-09-29
CertificateRequest state change audit log

pdeziel:
This PR is similar to #119 but for CertificateRequest objects. An audit log is added to the CertificateRequest object which is meant to capture all the state changes made while a certificate request is being processed as well as some metadata.

The `UpdateCertificateRequestStatus` function is added which changes the status of a CertificateRequest and appends a new entry to the audit log. I have tried to replace all the current instances of setting the status on CertificateRequests with the new function, but it's possible I missed some.


bbengfort (approved):
Well done, this seems like a pretty straight forward PR after #119 and #116 - I appreciate your attention to detail about which functions are returning errors and which are simply stopping; sorry about the inconsistency there. The tests were also very thorough, which I appreciate as well! 

It looks like there is one place where we'd like the email context added; so I suppose you'll update the PR with that before merging?

pkg/gds/admin.go
```diff
@@ -778,7 +778,10 @@ func (s *Admin) acceptRegistration(vasp *pb.VASP) (msg string, err error) {
 	for careqs.Next() {
 		req := careqs.CertReq()
 		if req != nil && req.Vasp == vasp.Id && req.Status == models.CertificateRequestState_INITIALIZED {
-			req.Status = models.CertificateRequestState_READY_TO_SUBMIT
+			// TODO: Replace "email" in the source parameter with user email address.
```
bbengfort, line 781: @pdeziel I'd like to merge this in to help me with SC-909; would you mind creating a story to make this small update?

PR #125, opened 2021-09-30
Implement Allowed Origins configuration

bbengfort:
Implements a configuration for CORS control of the Admin API. This will
allow us to more flexbily deploy the Admin UI on multiple domains while
restricting access necessary for credentials.


pdeziel (approved):
Looks good to me.

PR #127, opened 2021-10-01
feat: implement login functionality

elysee15:


web/gds-admin-ui/src/components/Auth/SignWithGoogle.js
```diff
@@ -5,21 +5,18 @@ import React from 'react'
...
+        if (window.google) {
+            window.google.accounts.id.initialize({
+                client_id: clientId,
+                callback: loginResponse,
+                text,
+            });
+            window.google.accounts.id.renderButton(
+                document.getElementById("buttonDiv"),
+                { theme: "filled_blue", size: "large", margin: "0 auto" }
+            );
```
bbengfort, line 17: does the removal of the `onload` event allow the button to be rendered only after the `GET /v2/authenticate` has been completed?

web/gds-admin-ui/src/helpers/api/apiCore.js
```diff
@@ -1,10 +1,13 @@
 import axios from 'axios';
 import { defaultEndpointPrefix } from '../../utils';
+import jwtDecode from 'jwt-decode'
 
-// content type
 axios.defaults.headers.post['Content-Type'] = 'application/json';
 axios.defaults.baseURL = defaultEndpointPrefix();
 
+axios.defaults.withCredentials = true
```
bbengfort, line 8: It might be helpful to add a comment here:

```
// withCredentials ensures that axios fetch includes `HttpOnly` cookies in the request for CSRF protectioin
```

web/gds-admin-ui/src/helpers/api/apiCore.js
```diff
@@ -37,17 +40,24 @@ axios.interceptors.response.use(
...
-    if (token) axios.defaults.headers.common['Authorization'] = 'JWT ' + token;
+    if (token) axios.defaults.headers.common['Authorization'] = 'Bearer ' + token;
     else delete axios.defaults.headers.common['Authorization'];
 };
 
+const setCookie = (cookie) => {
+    if (cookie) {
+        axios.defaults.headers['X-CSRF-Token'] = cookie
+    }
+}
```
bbengfort, line 57: Using the pattern from `setAuthorization` may I ask why this isn't:

```javascript
const setCookie = (cookie) => {
    if (cookie) axios.defaults.headers.common['X-CSRF-Token'] = cookie;
    else delete axios.defaults.headers.common['X-CSRF-Token'];
}
```

e.g. what is `headers.common`? and do we need the functionality to delete the cookie?

web/gds-admin-ui/src/helpers/api/apiCore.js
```diff
@@ -37,17 +40,24 @@ axios.interceptors.response.use(
...
+
+const AUTH_SESSION_KEY = '__SESSION_TOKEN__';
```
bbengfort, line 59: Just so I'm clear, this is to store the access token in local storage? What is the difference between `sessionStorage` and `localStorage`?

web/gds-admin-ui/src/helpers/api/apiCore.js
```diff
@@ -167,6 +177,23 @@ class APICore {
...
+        if (!user) {
+            return false;
+        }
+
+        const decodedAccessToken = jwtDecode(user.access_token);
+        console.log(jwtDecode(user.refresh_token))
+        const currentTime = Date.now() / 1000;
+        if (decodedAccessToken.exp < currentTime) {
+            console.warn('access token expired');
+            return false;
```
bbengfort, line 192: In the `reauthenticate` PR - will this function also check the refresh token?

web/gds-admin-ui/src/components/Auth/SignWithGoogle.js
elysee15: Yes, it does. I removed it because the google doesn't render correctly and I have to reload the page

web/gds-admin-ui/src/pages/account/Login.js
```diff
@@ -7,19 +7,50 @@ import AccountLayout from './AccountLayout';
...
+const api = new APICore()
 
 const Login = (): React$Element<any> => {
     const dispatch = useDispatch();
     const { userIsLoggedIn, user } = useSelector((state) => ({
         userIsLoggedIn: state.Auth.userIsLoggedIn,
         user: state.Auth.user
     }))
+    const [authenticated, setAuthenticated] = React.useState(false)
+    const isMounted = React.useRef(true)
```
bbengfort, line 22: Could you explain what `isMounted` is doing here please? E.g. why is this `useRef` instead of `useState(false)`?

web/gds-admin-ui/src/pages/account/Login.js
```diff
@@ -7,19 +7,50 @@ import AccountLayout from './AccountLayout';
...
 
+const api = new APICore()
 
 const Login = (): React$Element<any> => {
     const dispatch = useDispatch();
     const { userIsLoggedIn, user } = useSelector((state) => ({
         userIsLoggedIn: state.Auth.userIsLoggedIn,
         user: state.Auth.user
     }))
+    const [authenticated, setAuthenticated] = React.useState(false)
```
bbengfort, line 21: `authenticated` is not really the right verb here; I'd prefer to use `protected`, `csrfProtected`, or `doubleCookieProtected` for this state; otherwise folks who are reading this code will be confused about what is happening with the `GET` request.

bbengfort (approved):
@elysee15 this looks very good, thank you for your hard work on this PR. I had a couple of clarifying questions if you wouldn't mind - just so I understand the login workflow completely. 

One other question - where does the profile picture and name get extracted for the header bar? Is that still extracted from the Google credential or is it from the access token?

web/gds-admin-ui/src/helpers/api/apiCore.js
elysee15: `headers.common` is to set `axios` default instance
I can do `setCookie('')` when I want to make a request without cookie but it will remove the CSRF protection

web/gds-admin-ui/src/helpers/api/apiCore.js
elysee15: the `sessionStorage` is only available in the current tab you are using and will be automatically deleted after closing the tab.
the `localStorage` will keep the data until you delete it yourself.

web/gds-admin-ui/src/helpers/api/apiCore.js
elysee15: Yes

web/gds-admin-ui/src/pages/account/Login.js
elysee15: `isMounted` is a pattern in React to avoid `memory leaks` due to side effects like data fetching, setting up a subscription, and manually changing the DOM React components.
`useRef` unlike local variables that `useState` takes, React makes sure same reference is returned during each render

web/gds-admin-ui/src/pages/account/Login.js
elysee15: Yes, I had some doubt about it

web/gds-admin-ui/src/pages/account/Login.js
elysee15: > @elysee15 this looks very good, thank you for your hard work on this PR. I had a couple of clarifying questions if you wouldn't mind - just so I understand the login workflow completely.
> 
> One other question - where does the profile picture and name get extracted for the header bar? Is that still extracted from the Google credential or is it from the access token?

I take all this from the access token you give me after authentication

web/gds-admin-ui/src/helpers/api/apiCore.js
bbengfort: Perfect, thank you for the clarification.

web/gds-admin-ui/src/helpers/api/apiCore.js
bbengfort: Ok, let's keep it for now, but I think we may need the functionality in the future. Adding the csrf_token to all requests shouldn't hurt anything.

web/gds-admin-ui/src/pages/account/Login.js
```diff
@@ -7,19 +7,50 @@ import AccountLayout from './AccountLayout';
...
 
+const api = new APICore()
 
 const Login = (): React$Element<any> => {
     const dispatch = useDispatch();
     const { userIsLoggedIn, user } = useSelector((state) => ({
         userIsLoggedIn: state.Auth.userIsLoggedIn,
         user: state.Auth.user
     }))
+    const [csrfProtected, setCsrfProtected] = React.useState(false)
```
bbengfort, line 21: Thank you for making this change!

bbengfort (approved):

PR #128, opened 2021-10-01
[WIP] Admin timeline API

pdeziel:
This implements an endpoint for the Admin API which returns time series data in the form of JSON which contains a record for each week. Each record contains the week in YYYY-MM-DD format, the total number of VASPs, and counts of the VerificationState changes broken down by status. The value of this endpoint is that it can be used to render time series graphs on the frontend.

TODO

- [ ] Make start/end date configurable in the request
- [ ] Figure out what the route should be called, right now it's /v2/timeline
- [ ] The VASP count is not always going to be correct (e.g., if the state of a VASP doesn't change for a week then it won't be counted)...should we just remove it from the output?

### DIFF
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -989,23 +988,27 @@ func (s *Admin) Timeline(c *gin.Context) {
 		c.JSON(http.StatusBadRequest, admin.ErrorResponse(fmt.Errorf("invalid start date: %s", startTime.Format(time.RFC3339))))
 	}
 
-	// Initialize the record structs
-	type weekRecord struct {
-		VASPs         map[string]bool
-		Registrations map[string]int
-	}
+	// Initialize required counting structs
 	numWeeks := int(endTime.Sub(startTime).Hours()/24/7) + 1
-	weeks := make([]*weekRecord, 0, numWeeks)
+	vaspCounts := make([]map[string]bool, 0, numWeeks)
+	out := &admin.TimelineReply{
+		Weeks: make([]admin.TimelineRecord, 0, numWeeks),
+	}
+	weekTime := startTime
 	for i := 0; i < numWeeks; i++ {
-		record := &weekRecord{
-			VASPs:         make(map[string]bool),
+		weekDate := weekTime.Format(timeFormat)
+		record := admin.TimelineRecord{
+			Week:          weekDate,
+			VASPsCount:    0,
 			Registrations: make(map[string]int),
 		}
 		var s int32
 		for s = 0; s <= int32(pb.VerificationState_ERRORED); s++ {
 			record.Registrations[pb.VerificationState_name[s]] = 0
 		}
-		weeks[i] = record
+		out.Weeks[i] = record
+		vaspCounts[i] = make(map[string]bool)
+		weekTime = weekTime.Add(time.Hour * 24 * 7)
 	}
 
 	// Iterate over the VASPs and count registrations
@@ -1036,11 +1039,16 @@ func (s *Admin) Timeline(c *gin.Context) {
 			}
 			weekNum := int(timestamp.Sub(startTime).Hours() / 24 / 7)
 
-			// Mark VASP if we haven't seen it before
-			if _, exists := weeks[weekNum].VASPs[vasp.Id]; !exists {
-				weeks[weekNum].VASPs[vasp.Id] = true
+			// Count VASP if we haven't seen it before
+			if _, exists := vaspCounts[weekNum][vasp.Id]; !exists {
+				vaspCounts[weekNum][vasp.Id] = true
+				out.Weeks[weekNum].VASPsCount++
 			}
-			weeks[weekNum].Registrations[vasp.VerificationStatus.String()]++
+
+			// Count registration state if it changed
+			if entry.PreviousState != entry.CurrentState {
+				out.Weeks[weekNum].Registrations[entry.CurrentState.String()]++
+			}
 		}
 	}
 
@@ -1050,27 +1058,6 @@ func (s *Admin) Timeline(c *gin.Context) {
 		return
 	}
 
-	// Construct timeline reply
-	out := &admin.TimelineReply{
-		Weeks: make([]admin.TimelineRecord, 0, numWeeks),
-	}
-	weekTime := startTime
-	for i := 0; i < numWeeks; i++ {
-		weekDate := weekTime.Format(timeFormat)
-		out.Weeks = append(out.Weeks, admin.TimelineRecord{
-			Week:          weekDate,
-			VASPsCount:    len(weeks[i].VASPs),
-			Registrations: weeks[i].Registrations,
-		})
-		weekTime = weekTime.Add(time.Hour * 24 * 7)
-	}
-
-	// Sort output by week
-	sort.Slice(out.Weeks, func(i, j int) bool {
-		first, _ := time.Parse(timeFormat, out.Weeks[i].Week)
-		second, _ := time.Parse(timeFormat, out.Weeks[j].Week)
-		return first.Before(second)
-	})
 	c.JSON(http.StatusOK, out)
 }
 
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -1020,14 +1020,17 @@ func (s *Admin) Timeline(c *gin.Context) {
 		}
 
 		// Get VASP audit log
-		if auditLog, err := models.GetAuditLog(vasp); err != nil {
+		var auditLog []*models.AuditLogEntry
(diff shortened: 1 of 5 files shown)


pkg/gds/admin/v2/client.go
```diff
@@ -224,6 +224,21 @@ func (s APIv2) Resend(ctx context.Context, in *ResendRequest) (out *ResendReply,
 	return out, nil
 }
 
+func (s APIv2) Timeline(ctx context.Context) (out *TimelineReply, err error) {
```
tl6kk, line 227: Not sure if here we want `(s *APIv2)` instead? I remember I had a similar comment from one of the localew reviews, that using a pointer would save us some space.

tl6kk (approved):
The code looks very clear to me 👍🏻  

I only have a small inline comment.

pkg/gds/admin/v2/client.go
pdeziel: Good question, I was trying to follow the existing format for the other function definitions in the file but it looks like @bbengfort is adding the * in his new [PR](https://github.com/trisacrypto/directory/pull/129#discussion_r720226494) anyway.

bbengfort: In a previous PR I did have to update the receiver to a pointer. You can think about the `(s APIv2)` as similar to `self` in Python, it's the "instance" (no classes in Go, but the metaphor is good) referred to by the method. If you have a pointer to the instance, you can modify the instance; without the pointer, you have a copy passed by value, which means no modifications will take effect.

See the go example here: https://play.golang.org/p/v35S-ZNbiSX

bbengfort:
@pdeziel I merged main to make the PR easier to review.

pkg/gds/admin.go
```diff
@@ -1006,6 +1007,133 @@ func (s *Admin) Resend(c *gin.Context) {
...
+	// Get request parameters
+	in = new(admin.ReviewTimelineParams)
+	if err = c.ShouldBindQuery(&in); err != nil {
+		log.Warn().Err(err).Msg("could not bind request with query params")
+		c.JSON(http.StatusBadRequest, admin.ErrorResponse(err))
+		return
+	}
+
+	// Default value for start date
+	earliestTime = time.Date(2020, time.January, 1, 0, 0, 0, 0, time.Local)
```
bbengfort, line 1035: Times are stored in the database as UTC - perhaps it would be better to use `time.UTC` for the tzinfo in this date?

pkg/gds/admin.go
```diff
@@ -1006,6 +1007,133 @@ func (s *Admin) Resend(c *gin.Context) {
...
+	if in.Start != "" {
+		// Parse start date
+		if startTime, err = time.Parse(timeFormat, in.Start); err != nil {
+			log.Warn().Err(err).Msg("could not parse start date")
+			c.JSON(http.StatusBadRequest, admin.ErrorResponse(fmt.Errorf("invalid start date: %s", in.Start)))
+		}
+		// Hard limit on the earliest date to avoid making the server do unnecessary work
+		if startTime.Before(earliestTime) {
+			startTime = earliestTime
+		}
```
bbengfort, line 1045: I totally understand what you're getting at with this hard limit, but since you're defaulting to `earliestTime` and the user is submitting the request for this time, it's probably best not restrict them artificially -- even if the user submits epoch time you're still talking about only ~3500 weeks, which isn't terrible as a for loop. Perhaps you could `log.Warn()` if the number of weeks is > 260 (e.g. five years)?

pkg/gds/admin.go
```diff
@@ -1006,6 +1007,133 @@ func (s *Admin) Resend(c *gin.Context) {
...
+	} else {
+		startTime = earliestTime
```
bbengfort, line 1047: I would recommend that the default startTime should be 1 year ago; e.g. only fetch the past year of timeline data - that will make things much easier to display on the front-end. As written, this function will continue to return more and more data as time goes on. 

This brings up another issue - the weeks are currently aligned with the difference of the end and start times. If the end time is a Thursday and the start time is a Tuesday, then the weeks will be 7 day periods Wednesday-Wednesday, which will probably introduce confusion on the front end. Instead, I propose we use the `startTime.ISOWeek()` method to get the year and week number of the start time and iterate forward until we get to the `endTime.ISOWeek()`. Here is a discussion of how to get the week start date from the ISO week: https://xferion.com/golang-reverse-isoweek-get-the-date-of-the-first-day-of-iso-week/

So what does this mean in the case where the start time is a Thursday and the end time is a Tuesday? We could do a few things:

1. Just use start/end to select the week, aligning to the nearest week that includes the specified dates
2. Change start and end params to simply specifying the number of weeks to return 
3. Filter audit records by date so they only fall into the specified time period even for partial weeks

I recommend option 1

pkg/gds/admin.go
```diff
@@ -1006,6 +1007,133 @@ func (s *Admin) Resend(c *gin.Context) {
...
+		weekDate := weekTime.Format(timeFormat)
+		record := admin.ReviewTimelineRecord{
+			Week:          weekDate,
+			VASPsCount:    0,
+			Registrations: make(map[string]int),
+		}
+		var s int32
+		for s = 0; s <= int32(pb.VerificationState_ERRORED); s++ {
+			record.Registrations[pb.VerificationState_name[s]] = 0
+		}
```
bbengfort, line 1082: Just to clarify - the default value of the map is already 0 so if you access a key that doesn't exist in the map it will return 0 without an error. Addition also works, e.g.

```go
counts := make(map[string]int)
counts["foo"]++
counts["bar"]++
counts["other"] += 11
fmt.Println(counts)
```

Prints: `map[bar:1 foo:1 other:11]`

Therefore this section is not necessary for initialization, but is necessary to ensure that all states are represented with a default count of 0  (otherwise the JSON would simply omit the state) - is that correct? If so, I think a comment to that effect would be very helpful to prevent well-intending future contributors from removing this section. 

Alternatively we could talk to @elysee15 to see if he can handle a "missing" state with a count of 0 - this would reduce the amount of data transferred, only sending states with counts > 0. That would be better, but only if the front-end graph can handle that easily.

pkg/gds/admin.go
```diff
@@ -1006,6 +1007,133 @@ func (s *Admin) Resend(c *gin.Context) {
...
+			// Count VASP if we haven't seen it before
+			if _, exists := vaspCounts[weekNum][vasp.Id]; !exists {
+				vaspCounts[weekNum][vasp.Id] = true
+				out.Weeks[weekNum].VASPsCount++
+			}
+
+			// Count registration state if it changed
+			if entry.PreviousState != entry.CurrentState {
+				out.Weeks[weekNum].Registrations[entry.CurrentState.String()]++
+			}
```
bbengfort, line 1124: Excellent guard!

bbengfort (approved):
@pdeziel looks good so far! I had a couple of comments that I hope are helpful, let me know if more discussion is required. 

The changing VASP counts is definitely an issue but because the endpoint is "reviews" perhaps we frame this as a "changes over time" data structure? E.g. instead of `vasps_count` we could rename the field to `vasps_updated`. This actually makes it easier to do a cumulative rollup of the timeseries - which we could do on the server with a parameter or on the front end.

pkg/gds/admin.go
pdeziel: Agreed, that makes more sense.

pkg/gds/admin.go
pdeziel: My thinking was that the earliest date a user can technically request is 0000-01-01 which is ~100,000 weeks. Maybe a solution could be to change the hard limit to the epoch time or return an error if the given date is earlier than the epoch?

pkg/gds/admin.go
pdeziel: Option 1 sounds good, I agree that it's better if the weeks are more predictable for displaying on the front end.

pkg/gds/admin.go
pdeziel: I think that could work, depending on how easy it is to parse on the front-end. @elysee15 Does the front-end know about the possible verification states?

pkg/gds/admin.go
bbengfort: Ok, I'm on board with ensuring that it's later than epoch time - that's an understandable time limit that users will understand naturally.

pdeziel:
@bbengfort Mind taking a second look before I merge? I implemented the ISOWeek alignment scheme following your suggestion.

pkg/gds/admin.go
```diff
@@ -1006,6 +1008,147 @@ func (s *Admin) Resend(c *gin.Context) {
...
+	out = &admin.ReviewTimelineReply{
+		Weeks: make([]admin.ReviewTimelineRecord, 0, 1),
+	}
+
+	// Iterate over the weeks and record the week start dates
+	for {
+		var ok bool
+		if week, ok = weekIter.Next(); !ok {
+			break
+		}
```
bbengfort, line 1082: Much more readable!

PR #130, opened 2021-10-01
Review notes infrastructure

pdeziel:
This adds some infrastructure to support a review notes endpoint in the Admin API where notes can be created, updated, retrieved, and deleted. The data exists as a map in the `GDSExtraData` protobuf object and I've added some helper functions to support interacting with that map.


pkg/gds/models/v1/models.go
```diff
@@ -148,6 +148,131 @@ func UpdateCertificateRequestStatus(request *CertificateRequest, state Certifica
...
+	extra := &GDSExtraData{}
+	if vasp.Extra != nil {
+		if err = vasp.Extra.UnmarshalTo(extra); err != nil {
+			return fmt.Errorf("could not deserialize previous extra: %s", err)
+		}
+	}
+
+	if extra.ReviewNotes == nil {
+		extra.ReviewNotes = make(map[string]*ReviewNote)
+	}
```
rebeccabilbro, line 183: If `vasp.Extra` doesn't exist yet, might we get a nil pointer on line 181 because ReviewNotes doesn't exist yet also? I always forget how this works in Go, but I wonder if we need to do something like:

```golang
// Unmarshal previous extra data.
extra := &GDSExtraData{}
if vasp.Extra != nil {
	if err = vasp.Extra.UnmarshalTo(extra); err != nil {
		return fmt.Errorf("could not deserialize previous extra: %s", err)
	}
} else {
        extra.ReviewNotes = make(map[string]*ReviewNote)
}
```

pkg/gds/models/v1/models_test.go
```diff
@@ -141,6 +165,80 @@ func TestAuditLog(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Len(t, notes, 1)
+	require.Equal(t, "boats", notes["boats"].Id)
+	require.Equal(t, "", notes["boats"].Modified)
+	require.Equal(t, "pontoon@boatz.com", notes["boats"].Author)
+	require.Equal(t, "", notes["boats"].Editor)
+	require.Equal(t, "boats are cool", notes["boats"].Text)
+
+	// Attempt to update a note that doesn't exist
+	err = UpdateReviewNote(vasp, "jetskis", "admin@example.com", "jetskis are fun")
```
rebeccabilbro, line 199: love that we are sticking with the nautical theme 🛥️

rebeccabilbro (approved):
Looks good @pdeziel! 

I'm admittedly a bit out of sync with the GDS repo, so I did a quick review of the latest codebase and it looks like the note addition/modification/deletion functionality is super consistent with the audit log functionality (aside from the fact that it needs to be a map rather than an array). 

As I mentioned when we chatted, there does seem to be a teensy bit of variation across the Get methods in `models`, some of which return attrs of `extra` and others that return getters of `extra` (which I found defined in the compiled pb go code), but I'm chalking this up to us having multiple contributors with different styles 🙂 . 

I also had a tiny question about a possible nil pointer issue that I wanted you to take a quick look at just in case. Otherwise looks good to go!

pkg/gds/models/v1/models.go
pdeziel: I'm pretty sure Go lets us access extra.ReviewNotes, even though it's nil, because we have already instantiated a GDSExtraData on line 174. If we tried to access something inside ReviewNotes we would definitely get an error. I originally ran into a nil pointer exception which is why I added this check and the make.

PR #131, opened 2021-10-04
feat: Implement logout functionality

elysee15:
Fixes SC-582


web/gds-admin-ui/src/components/ProfileDropdown.js
```diff
@@ -12,17 +13,22 @@ type ProfileDropdownState = {
...
-     * toggle profile-dropdown
-     */
     const toggleDropdown = () => {
         setDropdownOpen(!dropdownOpen);
     };
 
+    const handleLogoutClick = () => {
+        setTimeout(() => {
+            api.deleteUserSession()
+        }, 1000)
```
bbengfort, line 29: Why the timeout of one second?

web/gds-admin-ui/src/helpers/api/auth.js
```diff
@@ -29,4 +24,8 @@ function forgotPasswordConfirm(params: any): any {
     return api.create(`${baseUrl}`, params);
 }
 
-export { login, logout, signup, forgotPassword, forgotPasswordConfirm };
+function postCredentials(credentials, params) {
+    return api.create('/authenticate', credentials, params)
+}
+
+export { login, signup, forgotPassword, forgotPasswordConfirm, postCredentials };
```
bbengfort, line 31: We can probably remove `signup`, `forgotPassword`, and `forgotPasswordConfirm`. How is `login` different from `postCredentials`? We should probably make that just one function.

web/gds-admin-ui/src/helpers/api/apiCore.js
```diff
@@ -188,17 +215,21 @@ class APICore {
             this.setLoggedInUser({ token, ...user, ...modifiedUser });
         }
     };
+
+    deleteUserSession = () => {
+        sessionStorage.removeItem(AUTH_SESSION_KEY)
+        window.location.href = '/login'
+    }
```
bbengfort, line 222: Should we also call `setAuthorization(null)` to remove the access token from the default common headers?

bbengfort (approved):
Approved pending the rebase with #127 - some questions, but this is looking very good!

web/gds-admin-ui/src/components/ProfileDropdown.js
elysee15: Just to put a latency because I found the logout to fast

web/gds-admin-ui/src/helpers/api/auth.js
elysee15: Okay

web/gds-admin-ui/src/helpers/api/apiCore.js
elysee15: alright

web/gds-admin-ui/src/components/ProfileDropdown.js
bbengfort: What happens when it is too fast? It seems like we'd want it to be pretty immediate, particularly because this is security related.

web/gds-admin-ui/src/helpers/api/index.js
```diff
@@ -1,4 +1,4 @@
 // @flow
-import { login, logout, signup, forgotPassword, forgotPasswordConfirm } from './auth';
+import { login, signup, forgotPassword, forgotPasswordConfirm } from './auth';
 
-export { login, logout, signup, forgotPassword, forgotPasswordConfirm };
+export { login, signup, forgotPassword, forgotPasswordConfirm };
```
bbengfort, line 4: @elysee15 do these also have to be removed?

bbengfort (approved):
@elysee15 I'm merging this in but it looks like `helpers/api/index.js` has some imports and exports that still need to be cleaned up if you could do that in the next PR.

PR #132, opened 2021-10-04
feat: implement user authorization

elysee15:


web/gds-admin-ui/src/pages/error/PageNotFound.js
```diff
@@ -45,7 +43,7 @@ const ErrorPageNotFound = (): React$Element<React$FragmentType> => {
                 </div>
             </div>
 
-            <footer className="footer footer-alt">2018 - 2021 © Hyper - Coderthemes.com</footer>
+            <footer className="footer footer-alt">Created and maintained by <a href="https://rotational.io">Rotational</a> Labs in partnership with <a href="https://ciphertrace.com">CipherTrace</a> on behalf of <a href="https://trisa.io">TRISA</a>.</footer>
```
bbengfort, line 46: Great catch! Thank you!

web/gds-admin-ui/src/routes/PrivateRoute.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { Route, Redirect } from 'react-router-dom';
+
+import { APICore } from '../helpers/api/apiCore';
+
+
+const PrivateRoute = ({ component: Component, ...rest }) => {
+    const api = new APICore();
```
bbengfort, line 8: In the future we should think about having simple state like the `user` and `isAuthenticated` stored in redux state rather than in session storage accessible only through the `APICore` functions. I don't believe that `user` and `isAuthenticated` are secure information, and they can be loaded from the session store using a reducer or an action. That might simplify this very common requirement to get access to the user state. What do you think? 

If you agree, please create a story in the backlog and we can work on it in the future!

bbengfort (approved):
I really like `PrivateRoute` and thank you for cleaning up the 404 page!

PR #133, opened 2021-10-04
fix: user profile

elysee15:
Capture d’écran 2021-10-04 à 09 18 20


bbengfort (approved):
@elysee15 looks great - thank you for cleaning up the search bar; glad that the user profile is working now with the new access token mechanism!

PR #136, opened 2021-10-05
feat: add refresh token

elysee15:


web/gds-admin-ui/src/helpers/api/apiCore.js
```diff
@@ -186,11 +199,29 @@ class APICore {
...
+                    }
+                }).then(res => {
+                    this.setLoggedInUser(res.data)
+                    setAuthorization(res.data.access_token)
+                }).catch((err) => {
+                    console.log('[/reauthenticate] error----', err)
+                })
+            }
+            return true;
+        } else {
```
bbengfort, line 222: This logic is not quite right:

```javascript
if (currentTime < decodedAccessToken.exp && currentTime > decodedAccessToken.nbf) {
    // The access token is valid -- we could just return true here
    // Alternatively, we could check if we're in that small window of time where we can reauthenticate when the access token is valid:
    if (isValidRefreshToken(user.refresh_token)) {
      const rep = await reauthenticate();
      return rep;
    }
    return true;
} else {
    // access token is invalid, check if we can reauthenticate
    if (isValidRefreshToken(user.refresh_token)) {
      const rep = await reauthenticate();
      return rep;
    }
    // neither the access nor the refresh token is valid any longer
   return false;
}
```

web/gds-admin-ui/src/helpers/api/apiCore.js
```diff
@@ -186,11 +199,29 @@ class APICore {
 
         const decodedAccessToken = jwtDecode(user.access_token);
         const currentTime = Date.now() / 1000;
-        if (decodedAccessToken.exp < currentTime) {
+        const csrfToken = getCookie('csrf_token');
+
+        if (currentTime < decodedAccessToken.exp && currentTime > decodedAccessToken.nbf) {
+            if (isValidRefreshToken(user.refresh_token)) {
```
bbengfort, line 205: The `refresh_token` is stored with the user - I didn't see where this was stored? Is the user also stored in the session state?

bbengfort (commented):
@elysee15 good start so far! I had a couple of comments.

Also, I was testing out the setTimeout for logout, and it was delaying far longer than a second for me, could we please remove the setTimeout and logout immediately?

web/gds-admin-ui/src/helpers/api/apiCore.js
elysee15: There concern is that I verified both tokens 
- The refresh token`nbf` is `35min` after token creation
- The refresh token expire at the same time with `access token` that is 1hour
Then I was expecting that the access token expire before the refresh token to be able to make a `reauthenticate` request if there is an `error`

web/gds-admin-ui/src/helpers/api/apiCore.js
elysee15: The user object `( access_token & refresh token )` is stored in the session

web/gds-admin-ui/src/layouts/Topbar.js
```diff
@@ -25,13 +29,22 @@ type TopbarProps = {
...
     const { user, loading } = useSelector(state => ({
         user: state.Auth.user,
         loading: state.Auth.loading
     }))
 
+    React.useEffect(() => {
+        const { access_token } = api.getLoggedInUser()
+        const decodedToken = access_token && jwtDecode(access_token)
+        dispatch(loginUserSuccess(AuthActionTypes.LOGIN_USER_SUCCESS, decodedToken))
+    }, [dispatch])
```
bbengfort, line 46: So this means that whenever the `Topbar` is rendered, the `getLoggedInUser` function may do reauthentication with the refresh token, so you're redispatching the successfully logged in user?

web/gds-admin-ui/src/helpers/api/apiCore.js
```diff
@@ -8,22 +8,25 @@ axios.defaults.baseURL = defaultEndpointPrefix();
...
             // window.location.href = '/not-found';
         } else if (error && error.response && error.response.status === 403) {
-            window.location.href = '/access-denied';
+            sessionStorage.removeItem(AUTH_SESSION_KEY)
+            window.location.href = '/login';
         } else {
             switch (error.response.status) {
                 case 401:
                     message = 'Invalid credentials';
+                    window.location.href = '/login'
```
bbengfort, line 29: is `sessionStorage.removeItem(AUTH_SESSION_KEY)` also required here?

web/gds-admin-ui/src/helpers/api/apiCore.js
```diff
@@ -186,11 +213,25 @@ class APICore {
...
+
+        const payload = {
+            credential: user.refresh_token
+        }
+
+        if (currentTime < decodedAccessToken.exp && currentTime > decodedAccessToken.nbf) {
+            // The access token is valid -- we could just return true here
+            // Alternatively, we could check if we're in that small window of time where we can reauthenticate when the access token is valid:
+            if (isValidRefreshToken(user.refresh_token)) {
+                this.reauthenticate(payload)
```
bbengfort, line 225: If this is successful, this should return true, but what if `reauthenticate` is not successful?

web/gds-admin-ui/src/helpers/api/apiCore.js
```diff
@@ -186,11 +213,25 @@ class APICore {
...
+            }
             return true;
+        } else {
+            // access token is invalid, check if we can reauthenticate
+            if (isValidRefreshToken(user.refresh_token)) {
+                this.reauthenticate(payload)
```
bbengfort, line 231: If this is successful, this function should return true; but it will return false on line 234.

web/gds-admin-ui/src/helpers/api/apiCore.js
elysee15: No, but I'm testing the way to fix an issue about redirection when user token expires

web/gds-admin-ui/src/helpers/api/apiCore.js
elysee15: No, it contain an async function

PR #137, opened 2021-10-05
email audit log

pdeziel:
This adds an audit log to the `GDSContactExtraData` which records emails sent to contacts. This is mostly useful for keeping track of when and what emails are resent to users. Currently the reason and email subject are recorded, but we could grab more fields from the email message if we want. Also, there is one spot in `SendReviewRequest` where we are not recording emails since they are sent to the TRISA admins instead of a contact. We could add a specialized audit log for those emails as well but I wasn't sure if that's something we are interested in tracking.


pkg/gds/models/v1/models.go
```diff
@@ -288,17 +288,70 @@ func GetContactVerification(contact *pb.Contact) (_ string, _ bool, err error) {
 	return extra.GetToken(), extra.GetVerified(), nil
 }
 
-// SetContactVerification token and verified status on the Contact record (completely
-// replaces the old record, which may not be ideal).
+// SetContactVerification token and verified status on the Contact record.
```
bbengfort, line 291: Good catch!

pkg/gds/models/v1/models_test.go
```diff
@@ -271,6 +271,25 @@ func TestContactExtra(t *testing.T) {
...
+	require.Len(t, emailLog, 1)
+	require.Equal(t, "verify_contact", emailLog[0].Reason)
+	require.Equal(t, "verification", emailLog[0].Subject)
+
+	// Should not overwrite contact verification
+	token, verified, err = GetContactVerification(contact)
+	require.NoError(t, err)
+	require.False(t, verified)
+	require.Equal(t, "12345", token)
+
```
bbengfort, line 292: Good tests!

pkg/gds/emails/client.go
```diff
@@ -298,9 +311,16 @@ func (m *EmailManager) SendDeliverCertificates(vasp *pb.VASP, path string) (sent
 				continue
 			}
 
+			sent++
+
+			if err = models.AppendEmailLog(contact, string(admin.ResendDeliverCerts), msg.Subject); err != nil {
+				log.Error().Err(err).Str("vasp", vasp.Id).Int("contact", idx).Msg("could not log deliver certs email")
+				nErrors++
+				continue
+			}
```
bbengfort, line 320: We're modifying the VASP record here, I'm assuming that at some point later in the function call the VASP is saved back to disk with the changes to its contacts?

bbengfort (commented):
Looks good so far! In [TestClientSendEmails](https://github.com/trisacrypto/directory/blob/main/pkg/gds/emails/client_test.go#L18) could we add a check to make sure that the email log of the VASP was updated and saved to disk please? I'll whisper you some email credentials.

PR #140, opened 2021-10-07
GDS Admin API: Extend VASP List for CSV Download

bbengfort:
Adds registered directory, verified on, and certificate serial number to
VASP list.

### DIFF
--- a/pkg/gds/admin/v2/api.go
+++ b/pkg/gds/admin/v2/api.go
@@ -97,13 +97,16 @@ type ListVASPsReply struct {
 
 // VASPSnippet provides summary information about a VASP.
 type VASPSnippet struct {
-	ID                 string          `json:"id"`
-	Name               string          `json:"name"`
-	CommonName         string          `json:"common_name"`
-	VerificationStatus string          `json:"verification_status,omitempty"`
-	LastUpdated        string          `json:"last_updated,omitempty"`
-	Traveler           bool            `json:"traveler"`
-	VerifiedContacts   map[string]bool `json:"verified_contacts"`
+	ID                  string          `json:"id"`
+	Name                string          `json:"name"`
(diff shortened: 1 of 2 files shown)


pkg/gds/admin/v2/api.go
```diff
@@ -97,13 +97,16 @@ type ListVASPsReply struct {
...
+	ID                  string          `json:"id"`
+	Name                string          `json:"name"`
+	CommonName          string          `json:"common_name"`
+	RegisteredDirectory string          `json:"registered_directory,omitempty"`
+	VerificationStatus  string          `json:"verification_status,omitempty"`
+	LastUpdated         string          `json:"last_updated,omitempty"`
+	VerifiedOn          string          `json:"verified_on,omitempty"`
+	Traveler            bool            `json:"traveler"`
+	CertificateSerial   string          `json:"certificate_serial_number,omitempty"`
+	VerifiedContacts    map[string]bool `json:"verified_contacts"`
```
pdeziel, line 109: For my own curiosity, does the order in which go struct fields are defined ever matter? Or are structs in go inherently unordered?

pdeziel (approved):
Looks good to me, I had a go question but it's unrelated to the approval of this PR.

pkg/gds/admin/v2/api.go
bbengfort: That's actually a good question - it turns out that the order does matter to the compiler. Depending on the data types described by the `struct` it's possible that two different structs can allocate different amounts of memory: https://medium.com/@felipedutratine/how-to-organize-the-go-struct-in-order-to-save-memory-c78afcf59ec2. 

I think that order may also matter in how the `reflect` package works. However, for our purposes, we can usually treat them as unordered.

PR #141, opened 2021-10-08
Add i18n wrappers and extract messages for translators

tl6kk:
This PR adds the necessary setups for using Lingui.js (e.g. the `.linguirc` file), and also adds in the `<Trans> text messages </Trans>` wrapper for all text messages to be translated. It also added i18n steps into the `README` file.

The translation files are in `locales/[locale]/messages.po`. The `zh` translation file is populated with random Chinese character just for testing purposes, to maybe render the Chinese site and check if all translations is in the correct place, before sending the translation files to translators.

### DIFF
--- a/web/gds-ui/README.md
+++ b/web/gds-ui/README.md
@@ -68,3 +68,44 @@ This section has moved here: [https://facebook.github.io/create-react-app/docs/d
 ### `yarn build` fails to minify
 
 This section has moved here: [https://facebook.github.io/create-react-app/docs/troubleshooting#npm-run-build-fails-to-minify](https://facebook.github.io/create-react-app/docs/troubleshooting#npm-run-build-fails-to-minify)
+
+## i8n steps
+
+### Install
+- Install `lingui/cli`, `@lingui/macro`, and `@lingui/react1
+```
+npm install --save-dev @lingui/cli @babel/core
+npm install --save-dev @lingui/macro babel-plugin-macros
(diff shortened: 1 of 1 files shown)


web/gds-ui/src/components/ivms101/Address.js
```diff
@@ -42,7 +44,7 @@ const Address = ({index, address, onChange, onDelete}) => {
 
   return (
     <>
-    <Form.Label>Address {index+1}</Form.Label>
+    <Form.Label><Trans>Address {index+1}</Trans></Form.Label>
```
bbengfort, line 47: This may be tricky … it's rendered as `Address 1` or `Address 2` - should the `<Trans>` tag include the `{index+1}`? How would that show up for the translator?

bbengfort (approved):
Well done, this was quite a lot of work! I'm very impressed! I had one question, but I think we can go ahead and merge this for now; I'll QA the english language site in case we need to do a push of the UI before translation is complete, but things are looking excellent so far!

web/gds-ui/src/components/ivms101/Address.js
tl6kk: Oops, this will probably not work with Trans. Both cases will be 
```
#: src/components/ivms101/Address.js:47
msgid "Address {0}"
msgstr "添加翻译"
```
in view of the translator. We might need to flatten the code a little bit to work with the translations.

PR #143, opened 2021-10-08
feat: get Timeline graph data from backend API

elysee15:


web/gds-admin-ui/src/helpers/api/apiCore.js
```diff
@@ -20,14 +20,12 @@ axios.interceptors.response.use(
...
-            console.log("errr-----", error)
-            setAuthorization(null)
-            window.location.href = '/login'
+            sessionStorage.removeItem(AUTH_SESSION_KEY)
+            window.location.href = '/login';
         } else {
             switch (error.response.status) {
                 case 401:
                     message = 'Invalid credentials';
-                    setAuthorization(null)
```
bbengfort, line 30: @elysee15 is this an error?

bbengfort (commented):
@elysee15 don't forget to remove the guard that we put in SC-830 so that the timeline graph displays!

bbengfort:
@elysee15 thank you for cleaning up the conditional render of the timeline!

PR #145, opened 2021-10-13
fix: vasps bagde doesn't appear for some statuses

elysee15:
Fixes SC-1299


web/gds-admin-ui/src/utils/index.js
```diff
@@ -62,7 +62,29 @@ function getCookie(name = '') {
...
+            return 'bg-danger'
+        case Status.APPEALED:
+            return 'bg-primary'
+        case Status.REVIEWED:
+        case Status.ISSUING_CERTIFICATE:
+            return 'bg-info'
+        default:
+            return undefined
+    }
+}
```
bbengfort, line 85: I suspect that we may want to use this function in other places as well since color to status mappings will happen frequently. That might mean that in the future we'll want to remove the `bg-` prefix (so that we can use `text-` prefix, for example); but I don't think we need to do that now, just wanted to call it out. 

Nice one!

web/gds-admin-ui/src/pages/app/lists/index.js
```diff
@@ -50,10 +51,7 @@ const StatusColumn = ({ row }) => {
     return (
         <React.Fragment>
             <span
-                className={classNames('badge', {
-                    'bg-success': row.original.verification_status === Status.VERIFIED,
-                    'bg-warning': row.original.verification_status === Status.SUBMITTED || row.original.verification_status === Status.PENDING_REVIEW,
-                })}>
+                className={classNames('badge', getStatusClassName(row.original.verification_status))}>
```
bbengfort, line 54: Before, was the badge uncolored and the text simply white, which is why we could not see it?

bbengfort (approved):
Excellent!

PR #147, opened 2021-10-14
sc-948 gdsutil profiles

pdeziel:
This upgrades gdsutil to use urfave v2 and adds a mechanism for managing profiles. Profiles basically consist of environment-like string variables that can be arbitrarily set, retrieved, and deleted. The user can switch between profiles in a manner similar to gcloud, and variable operations are always done on the current profile.

### DIFF
--- a/cmd/gdsutil/main.go
+++ b/cmd/gdsutil/main.go
@@ -545,7 +545,7 @@ func ldbKeys(c *cli.Context) (err error) {
 	}
 
 	if err = iter.Error(); err != nil {
-		return cli.NewExitError(err, 1)
+		return cli.Exit(err, 1)
 	}
 
 	return nil
@@ -559,10 +559,10 @@ func ldbGet(c *cli.Context) (err error) {
 		// Check that out is a directory
 		var info fs.FileInfo
 		if info, err = os.Stat(out); err != nil {
-			return cli.NewExitError("specify an existing, writeable directory to output files to", 1)
+			return cli.Exit("specify an existing, writeable directory to output files to", 1)
 		}
 		if !info.IsDir() {
-			return cli.NewExitError("specify a directory to write files out to", 1)
+			return cli.Exit("specify a directory to write files out to", 1)
 		}
 	}
 
@@ -570,12 +570,12 @@ func ldbGet(c *cli.Context) (err error) {
 	for _, keys := range c.Args().Slice() {
 		var key []byte
 		if key, err = wire.DecodeKey(keys, b64decode); err != nil {
-			return cli.NewExitError(fmt.Errorf("could not decode key: %s", err), 1)
+			return cli.Exit(fmt.Errorf("could not decode key: %s", err), 1)
 		}
 
 		var data []byte
 		if data, err = ldb.Get(key, nil); err != nil {
-			return cli.NewExitError(err, 1)
+			return cli.Exit(err, 1)
 		}
 
 		// Unmarshal the thing
@@ -588,15 +588,15 @@ func ldbGet(c *cli.Context) (err error) {
 		switch prefix {
 		case wire.NamespaceVASPs, wire.NamespaceCertReqs, wire.NamespaceReplicas:
 			if pbValue, err = wire.UnmarshalProto(prefix, data); err != nil {
-				return cli.NewExitError(err, 1)
+				return cli.Exit(err, 1)
 			}
 		case wire.NamespaceIndices:
 			if jsonValue, err = wire.UnmarshalIndex(data); err != nil {
-				return cli.NewExitError(err, 1)
+				return cli.Exit(err, 1)
 			}
 		case wire.NamespaceSequence:
 			if jsonValue, err = wire.UnmarshalSequence(data); err != nil {
-				return cli.NewExitError(err, 1)
+				return cli.Exit(err, 1)
 			}
 		default:
 			fmt.Fprintf(os.Stderr, "warning: cannot unmarshal unknown namespace %q, printing raw data\n", prefix)
@@ -607,7 +607,7 @@ func ldbGet(c *cli.Context) (err error) {
 		switch {
 		case jsonValue != nil:
 			if outdata, err = json.MarshalIndent(jsonValue, "", "  "); err != nil {
-				return cli.NewExitError(err, 1)
+				return cli.Exit(err, 1)
 			}
 		case pbValue != nil:
 			jsonpb := protojson.MarshalOptions{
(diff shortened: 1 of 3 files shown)


cmd/gdsutil/main.go
```diff
@@ -53,26 +55,29 @@ func main() {
...
 			Action:   ldbKeys,
 			Before:   openLevelDB,
 			Flags: []cli.Flag{
-				cli.StringFlag{
-					Name:   "d, db",
-					Usage:  "dsn to connect to trisa directory storage",
-					EnvVar: "GDS_DATABASE_URL",
+				&cli.StringFlag{
+					Name:    "d",
+					Aliases: []string{"db"},
```
bbengfort, line 68: I went the opposite direction with `Name: "db"` and `Aliases: []string{"d"}` - is it better to do it this way?

cmd/gdsutil/main.go
```diff
@@ -149,35 +162,95 @@ func main() {
...
-				cli.StringFlag{
-					Name:   "u, replica-endpoint",
-					Usage:  "the url to connect the directory replica client",
-					Value:  "replica.vaspdirectory.net:443",
-					EnvVar: "TRISA_DIRECTORY_REPLICA_URL",
+				&cli.StringFlag{
+					Name:    "u",
+					Aliases: []string{"replica-endpoint"},
+					Usage:   "the url to connect the directory replica client",
+					Value:   "replica.vaspdirectory.net:443",
```
bbengfort, line 246: If the Value is set here, it will be difficult to tell if the user specified it on the command line or if the value should come from the profile; I removed it from the `gds` CLI.

cmd/gds/main.go
```diff
@@ -52,8 +52,8 @@ func main() {
 				EnvVars: []string{"TRISA_DIRECTORY_PROFILE", "GDS_PROFILE"},
 			},
 			&cli.StringFlag{
-				Name:    "endpoint",
-				Aliases: []string{"e"},
+				Name:    "directory-endpoint",
+				Aliases: []string{"d"},
```
bbengfort, line 56: Could we also keep the `e` endpoint, e.g. `Aliases: []string{"d", "e"}`? It's so ingrained that I use -e for endpoint that it will probably trip me up. Also I've noticed that there is no `GlobalString` in urfave v2 anymore, which means that if an flag conflicts with a subcommand flag, e.g. `gds -d foo cmd -d bar` then "bar" will be the value taking. `-d` is commonly used for database, directory, etc.

cmd/gds/main.go
```diff
@@ -1033,19 +1033,19 @@ func adminDeleteNote(c *cli.Context) (err error) {
 func manageProfiles(c *cli.Context) (err error) {
 	// Handle list and then exit
 	if c.Bool("list") {
-		var p profiles.Profiles
+		var p *profiles.Profiles
```
bbengfort, line 1036: Good catch! I might have missed the change from a `map` to ` struct` here!

cmd/gdsutil/main.go
bbengfort: Note that we should still remove the Value here in favor of the profiles.

cmd/gdsutil/main.go
```diff
@@ -188,16 +240,23 @@ func main() {
...
-				cli.StringFlag{
-					Name:   "u, replica-endpoint",
-					Usage:  "the url to connect the directory replica client",
-					Value:  "replica.vaspdirectory.net:443",
-					EnvVar: "TRISA_DIRECTORY_REPLICA_URL",
+				&cli.StringFlag{
+					Name:    "replica-endpoint",
+					Aliases: []string{"u"},
+					Usage:   "the url to connect the directory replica client",
+					Value:   "replica.vaspdirectory.net:443",
```
bbengfort, line 247: This value should also be removed in favor of the profile.

cmd/gdsutil/main.go
```diff
@@ -208,20 +267,28 @@ func main() {
...
```
bbengfort, line 274: And here!

cmd/gdsutil/main.go
```diff
@@ -973,7 +1135,7 @@ func registerExport(c *cli.Context) (err error) {
 	pbForm := &api.RegisterRequest{
 		Entity:           vasp.Entity,
 		Contacts:         vasp.Contacts,
-		TrisaEndpoint:    c.String("endpoint"),
+		TrisaEndpoint:    c.String("admin-endpoint"),
```
bbengfort, line 1138: The RegisterRequest should use the `directory-endpoint` I believe.

pkg/gds/client/profiles.go
```diff
@@ -11,60 +11,139 @@ import (
...
 		}
 
-		p = make(Profiles)
 		if err = yaml.Unmarshal(data, &p); err != nil {
 			return nil, fmt.Errorf("could not unmarshal profiles: %s", err)
 		}
 
+		if p.Version != profileVersion {
+			return nil, fmt.Errorf("invalid profile version %s, expected %s", p.Version, profileVersion)
+		}
```
bbengfort, line 132: Good check!

pkg/gds/client/profiles.go
```diff
@@ -75,69 +154,51 @@ func LoadActive(c *cli.Context) (p *Profile, err error) {
...
-	folders := cfgd.QueryFolders(configdir.Global)
-	if len(folders) == 0 {
-		return errors.New("could not find user configuration directory")
+// Save the profiles to disk in the specified configuration folder. If the configuration
+// folder is nil, the configuration folder is located and created if it doesn't exist.
+func (p Profiles) Save(folder *configdir.Config) (err error) {
+	if folder == nil {
+		if folder, err = GetProfilesFolder(); err != nil {
+			return fmt.Errorf("could not find profiles folder: %s", err)
+		}
```
bbengfort, line 192: Because the search order starts with the local directory, will this create the config folder in the CWD? I think the default should be in the user folder, but let me know if I'm missing something here.

bbengfort (approved):
@pdeziel looks good so far! There are a couple of minor things that I caught that need to be switched and I had a couple of questions, but after that we should be good to merge. Thanks for taking this one on, I know it was a big one!

pkg/gds/client/profiles.go
pdeziel: You're right, that's the current behavior. I think defaulting to the user folder does make more sense, so maybe I'll make the write precedence user -> global -> local and then use the library's local -> user -> global search order when looking for the config file so that the user can still create an overriding config in the CWD.

PR #148, opened 2021-10-14
WIP: GDS CLI upgrade to urfave.v2 and profiles

bbengfort:


pkg/gds/client/profiles.go
```diff
@@ -0,0 +1,179 @@
...
+	return "", errors.New("no profiles are available")
+}
+
+// Load the profiles from disk if they're available.
+func Load() (p Profiles, err error) {
+	folder := cfgd.QueryFolderContainsFile(profileYAML)
+	if folder != nil {
+		var data []byte
+		if data, err = folder.ReadFile(profileYAML); err != nil {
+			return nil, fmt.Errorf("could not read %s: %s", profileYAML, err)
```
pdeziel, line 47: We could probably print out the full path here for debugging purposes.

pkg/gds/client/profiles.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		return p, nil
+	}
+	return nil, errors.New("no profiles are available")
+}
+
+// LoadActive is a shorthand for Load() then Active() and finally Update()
+func LoadActive(c *cli.Context) (p *Profile, err error) {
+	var profiles Profiles
```
pdeziel, line 62: I can see myself getting tripped up with the similarity between `Profile` and `Profiles`, maybe `Profiles` should be renamed to something like `ProfileMap`, `ProfileManifest`, or `ProfileManager`? Same goes for the filenames.

pkg/gds/client/profiles.go
```diff
@@ -0,0 +1,179 @@
...
+	if err = p.Update(c); err != nil {
+		return nil, err
+	}
+
+	return p, nil
+}
+
+// Active returns the profile with the specified name or the active profile if no name
+// is specified. If multiple profiles are marked active it returns the first active
+// profile with no ordering gurantees. If no profiles are marked active and there is
```
pdeziel, line 80:
```suggestion
// profile with no ordering guarantees. If no profiles are marked active and there is
```

pkg/gds/client/profiles.go
```diff
@@ -0,0 +1,179 @@
...
+
+	// Save the configuration to the first folder
+	if err = folders[0].WriteFile(profileYAML, data); err != nil {
+		return fmt.Errorf("could not write profiles to disk: %s", err)
+	}
+
+	return nil
+}
+
+// Install creates a default profile and saves it to disk, overwriting the previous contents.
```
pdeziel, line 147:
```suggestion
// Install creates default profiles and saves them to disk, overwriting the previous contents.
```

cmd/gds/main.go
```diff
@@ -888,92 +976,171 @@ func adminUpdateNote(c *cli.Context) (err error) {
...
+		if err = p.SetActive(name); err != nil {
+			return cli.Exit(err, 1)
+		}
+		fmt.Printf("profile %q is now active\n", name)
+		return nil
+	}
+
+	// Handle show named or active profile
+	if c.Args().Len() > 1 {
+		return cli.Exit("specify ony a single profile to print", 1)
```
pdeziel, line 1093:
```suggestion
		return cli.Exit("specify only a single profile to print", 1)
```

cmd/gds/main.go
```diff
@@ -37,392 +38,479 @@ func main() {
...
 				},
-				cli.StringFlag{
-					Name:  "n, name",
-					Usage: "the name of the note to delete",
+			},
+			{
+				Name:      "profile",
+				Aliases:   []string{"config"},
+				Usage:     "view and manage profiles to configure client with",
+				UsageText: "gds profile [name]\n   gds profile --activate [name]\n   gds profile --list\n   gds profile --path\n   gds profile --install",
```
pdeziel, line 483: Is the idea that the user edits the configuration file themselves? Would there be any benefit to also allowing the user to set individual config settings through the CLI?

pdeziel (commented):
Looks pretty good so far, I had a few comments mostly relating to naming.

PR #151, opened 2021-10-15
Fetch peer identity from gRPC connection context

pdeziel:
This adds an interceptor for the Replication Server which fetches peer identity information from the peer certificate supplied in the TLS authentication process and adds it to the context so that it can be used in the gRPC service functions. Note that this requires mTLS to be implemented for this to do anything useful.

### DIFF
--- a/pkg/gds/interceptor.go
+++ b/pkg/gds/interceptor.go
@@ -48,7 +48,7 @@ func (s *Service) replicaInterceptor(ctx context.Context, in interface{}, info *
 	start := time.Now()
 
 	// Check if we're in maintenance mode - status method should still return a full response.
-	if s.conf.Maintenance && info.FullMethod != "/trisa.gds.api.v1beta1.TRISADirectory/Status" {
+	if s.conf.Maintenance {
 		return nil, status.Error(codes.Unavailable, "the GDS service is currently in maintenance mode")
 	}
 
@@ -102,8 +102,9 @@ func peerFromTLS(ctx context.Context) (info *PeerInfo, err error) {
 			info = &PeerInfo{
 				Name:        &c[0].Subject,
 				DNSNames:    c[0].DNSNames,
-				IPAddresses: c[0].IPAddresses,
+				IPAddresses: []net.IP{net.ParseIP(gp.Addr.String())},
 			}
+			info.IPAddresses = append(info.IPAddresses, c[0].IPAddresses...)
 			return info, nil
 		}
 	}


pkg/gds/interceptor.go
```diff
@@ -26,3 +42,71 @@ func (s *Service) serverInterceptor(ctx context.Context, in interface{}, info *g
...
+	if s.conf.Maintenance && info.FullMethod != "/trisa.gds.api.v1beta1.TRISADirectory/Status" {
+		return nil, status.Error(codes.Unavailable, "the GDS service is currently in maintenance mode")
+	}
+
+	// Fetch peer information from the TLS info.
+	var peer *PeerInfo
+	if peer, err = peerFromTLS(ctx); err != nil {
+		log.Error().Err(err).Msg("unable to retrieve remote peer info")
+		// TODO: Uncomment when mTLS is added
+		//return nil, status.Error(codes.Unauthenticated, "unable to retrieve authenticated peer information")
```
bbengfort, line 60: Perfect!

pkg/gds/interceptor.go
```diff
@@ -26,3 +42,71 @@ func (s *Service) serverInterceptor(ctx context.Context, in interface{}, info *g
...
+
+	// Search for a valid peer certificate.
+	for _, c := range chains {
+		// Certificate chain should contain at least the peer certificate and the CA.
+		if len(c) >= 2 {
+			// The peer certificate is always first.
+			info = &PeerInfo{
+				Name:        &c[0].Subject,
+				DNSNames:    c[0].DNSNames,
+				IPAddresses: c[0].IPAddresses,
```
bbengfort, line 105: Could we perhaps also add the `Addr` from the `gp` to the start of the list of IP addresses: https://pkg.go.dev/google.golang.org/grpc/peer#Peer? 

It's possible that we will have IP addresses on the certificate, but it's usually not recommended since IP addresses can change and then you have to issue new certs. I think we should keep it here just in case it is an mTLS mechanism we use, but having the `Addr` on the list as backup would be good.

pkg/gds/interceptor.go
bbengfort: When we add mTLS are there any other checks we need to do to ensure mTLS is valid?

pkg/gds/interceptor.go
```diff
@@ -26,3 +42,72 @@ func (s *Service) serverInterceptor(ctx context.Context, in interface{}, info *g
...
+
+	// Search for a valid peer certificate.
+	for _, c := range chains {
+		// Certificate chain should contain at least the peer certificate and the CA.
+		if len(c) >= 2 {
+			// The peer certificate is always first.
+			info = &PeerInfo{
+				Name:        &c[0].Subject,
+				DNSNames:    c[0].DNSNames,
+				IPAddresses: []net.IP{net.ParseIP(gp.Addr.String())},
```
bbengfort, line 105: Fascinating that `net.ParseIP` returns nil if `gp.Addr.String()` is not a valid IP

bbengfort (approved):

PR #152, opened 2021-10-18
Adds Admin Review Link to Notification Email

bbengfort:
Adds a link to the admin UI to the notification email sent to TRISA admins when a registration is ready for review.


pkg/gds/emails/client.go
```diff
@@ -26,6 +26,14 @@ func New(conf config.EmailConfig) (m *EmailManager, err error) {
 		client: sendgrid.NewSendClient(conf.SendGridAPIKey),
 	}
 
+	// Warn if email configuration isn't complete and will produce partial emails.
+	if conf.VerifyContactBaseURL == "" || conf.AdminReviewBaseURL == "" {
+		log.Warn().
+			Bool("missing_verify_contact_base_url", conf.VerifyContactBaseURL == "").
+			Bool("missing_admin_review_base_url", conf.AdminReviewBaseURL == "").
+			Msg("partial email configuration, some emails may not include links")
+	}
```
bbengfort, line 35: When the email client is initialized, that's when we're issuing warnings about partial configs now. Note that the configs are still optional, but the admins should be alerted.

pkg/gds/emails/emails_test.go
```diff
@@ -48,31 +48,43 @@ func TestVerifyContactURL(t *testing.T) {
...
 	}
-	link, err := url.Parse(data.VerifyContactURL())
-	require.NoError(t, err)
-	require.Equal(t, "https", link.Scheme)
-	require.Equal(t, "vaspdirectory.net", link.Host)
-	require.Equal(t, "/verify-contact", link.Path)
-	params := link.Query()
-	require.Equal(t, data.Token, params.Get("token"))
-	require.Equal(t, data.VID, params.Get("vaspID"))
+	require.Empty(t, data.VerifyContactURL(), "if no base url is provided, VerifyContactURL() should return empty string")
```
pdeziel, line 51: I didn't realize you could add a debug string to the require, that's pretty useful!

containers/docker-compose.yaml
```diff
@@ -49,7 +52,7 @@ services:
       - GDS_BACKUP_KEEP=1
       - GOOGLE_APPLICATION_CREDENTIALS
       - GOOGLE_PROJECT_NAME
-      - GDS_SECRETS_TESTING
+      - GDS_SECRETS_TESTING=true
```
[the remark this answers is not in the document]
bbengfort: `GDS_SECRETS_TESTING` uses the mock secrets manager rather than google secret manager, which I think is what we want when we're running things locally. One of `GOOGLE_APPLICATION_CREDENTIALS` or `GDS_SECRETS_TESTING` needs to be set. Perhaps it is better to just leave this to the environment though; I'll change it back.

PR #153, opened 2021-10-18
fix: Show 404 page on VASP not found.

elysee15:
Screen Shot 2021-10-18 at 22 30 31


tl6kk (approved):
Tested with an invalid VASP ID and looks great from my end! The return home button also successfully brings me back to the dashboard.

PR #154, opened 2021-10-18
feat: Implement review speed Pie chart with summary endpoint

elysee15:
Fixes SC-1304
Screen Shot 2021-10-18 at 23 27 12

### DIFF
--- a/web/gds-admin-ui/src/pages/app/dashboard/Status.js
+++ b/web/gds-admin-ui/src/pages/app/dashboard/Status.js
@@ -1,18 +1,23 @@
 import React from 'react';
 import { Doughnut } from 'react-chartjs-2';
 import { Card, Row, Col } from 'react-bootstrap';
-import { capitalizeFirstLetter, convertCountsToPercentages } from '../../../utils';
+import { capitalizeFirstLetter, getRatios } from '../../../utils';
+import { Status as STATUS } from '../../../constants';
 
 const Status = ({ statuses }) => {
     const colors = ['#0acf97', '#727cf5', '#fa5c7c'];
 
-    const getDonutChartData = () => {
+    const statusRatios = () => {
         if (statuses && typeof statuses === "object") {
-            return Object.values(convertCountsToPercentages(statuses))
+            return getRatios(statuses)
         }
 
-        return []
+        return {}
     }
+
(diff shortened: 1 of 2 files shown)


web/gds-admin-ui/src/utils/index.js
```diff
@@ -43,7 +43,7 @@ function formatDisplayedData(target) {
 
 const convertCountsToPercentages = (data) => {
     const total = Object.values(data).reduce((acc, x) => acc + x);
-    return Object.fromEntries(Object.entries(data).map(([k, v]) => [k, v / total]));
+    return Object.fromEntries(Object.entries(data).map(([k, v]) => [k, (v / total).toFixed(1)]));
```
bbengfort, line 46: You'll need to update this to:

```
return Object.fromEntries(Object.entries(data).map(([k, v]) => [k, ((v / total)* 100.0).toFixed(1)]));
```

In order to get the `%` rather than the ratio.

bbengfort: However, the donut chart might require the ratios so you might have to do this computation second.

bbengfort (commented):
Great start! I had a note that the donut chart data and display percentages are mismatched. I have to get on a call now, but if you can fix it that'd be great; otherwise I'll be back to make suggestions shortly.

web/gds-admin-ui/src/pages/app/dashboard/Status.js
```diff
@@ -2,18 +2,22 @@ import React from 'react';
...
-            return Object.values(convertCountsToPercentages(statuses))
+            return convertCountsToPercentages(statuses)
         }
 
-        return []
+        return {}
     }
 
+    const getDonutChartData = () => Object.values(statusesPercentages())
+
```
bbengfort, line 19: @elysee15 what I recommend is that we do the following:

```javascript
const getDonutChartData = () => Object.values(statusesPercentages())
const statusPercents = () => Object.fromEntries(statusPercentages().map((key, val) => [key, val*100.0]))
```

Then use `statusPercents` in the `<span>`, e.g. `<span>{statusPercents()[STATUS.PENDING_REVIEW] + '%'}</span>

bbengfort (approved):
Thank you for making those changes!

PR #155, opened 2021-10-19
sc-288 Mock SendGrid for email testing

pdeziel:
This creates a mock for SendGrid which does some email message validation, writes to an in-memory store, and returns a REST response. The way it's intended to be used is to call `emails.NewMock()` instead of `emails.New()`, which creates a mocked EmailManager which contains a mocked SendGrid client that contains a running list of sent emails that tests can verify against.

One thing that could be added is writing the email data out to disk on `Send`, although I wasn't sure if that actually makes the integration tests easier.


pkg/gds/emails/mock.go
```diff
@@ -0,0 +1,124 @@
...
+
+	return m, nil
+}
+
+type mockEmailManager struct {
+	conf         config.EmailConfig
+	Client       *mockSendGridClient
+	serviceEmail *mail.Address
+	adminsEmail  *mail.Address
+}
```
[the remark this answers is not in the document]
bbengfort: It's a good start, the trick is to use an interface like you suggest. 

```go
type EmailClient interface {
   Send(email *sgmail.SGMailV3) (*rest.Response, error)
}
```

Then the email manager is defined as:

```go
// EmailManager allows the server to send rich emails using the SendGrid service.
type EmailManager struct {
	conf         config.EmailConfig
	client       EmailClient
	serviceEmail *mail.Address
	adminsEmail  *mail.Address
}
```

At this point, both `mockSendGridClient` and `sendgrid.Client` implement the `EmailClient` interface since they both have the same `Send` method (if there are any other `sendgrid.Client` methods that the `EmailManager` uses, we'll have to add them to the interface and to the `mockSendGridClient` as well). 

Then in the `New` function, we'll use a configuration, e.g. `GDS_EMAIL_TESTING=true` so detect whether or not to create a `mockSendGridClient` in the `EmailManager` or a `sendgrid.Client`. The interface allows us to use either interchangeably. 

Because only the calls to `sendgrid.Client` are mocked, we can test all of the `EmailManager` code without actually executing a SendGrid request.

bbengfort (commented):
Great start! I answered your question inline.

pkg/gds/emails/client.go
```diff
@@ -49,14 +51,19 @@ func New(conf config.EmailConfig) (m *EmailManager, err error) {
 // EmailManager allows the server to send rich emails using the SendGrid service.
 type EmailManager struct {
 	conf         config.EmailConfig
-	client       *sendgrid.Client
+	Client       EmailClient
```
bbengfort, line 54: The reason you've made this property public (capitalized) is so that the tests can access the client in order to make assertions? If so, it would be my preference to keep this private and then have a getter method at the bottom of the file: 

```go
// Client returns either the mock or the sendgrid client. This method is primarily used for testing. 
func (m *EmailManager) Client() EmailClient { 
    return m.client 
}
``` 

This ensures that the `EmailClient` can't be modified externally and is a pattern I've adopted in other places, e.g. with `Routes()` on the admin api server.

pkg/gds/emails/client.go
```diff
@@ -49,14 +51,19 @@ func New(conf config.EmailConfig) (m *EmailManager, err error) {
...
+type EmailClient interface {
+	Send(email *sgmail.SGMailV3) (*rest.Response, error)
+}
+
 func (m *EmailManager) Send(message *sgmail.SGMailV3) (err error) {
 	var rep *rest.Response
-	if rep, err = m.client.Send(message); err != nil {
+	if rep, err = m.Client.Send(message); err != nil {
 		return err
 	}
```
bbengfort, line 68: This turned out to be pretty straight forward, I was worried there were other `sendgrid.Client` messages I was missing!

pkg/gds/emails/emails_test.go
```diff
@@ -90,18 +90,19 @@ func TestSendEmails(t *testing.T) {
...
 	// emails - this is to test the rendering of the emails with data only; it does not
 	// go through any of the server workflow for generating tokens, etc.
 	var conf config.EmailConfig
 	err := envconfig.Process("gds", &conf)
 	require.NoError(t, err)
 
-	// This test sends emails from the serviceEmail using SendGrid to the adminsEmail
+	if !conf.EmailTesting {
+		t.Skip("skip generate and send emails test")
+	}
```
bbengfort, line 102: The way this test worked before was to actually send the emails and we used it both to verify that the `EmailManager` was working correctly and that the emails looked the way they should after being rendered from the templates. 

With the mock in place I don't think there is any reason to trigger an email send (though I guess we could keep the original test function in place called `TestSendLiveEmails` and only use it if something is wrong with email sending). Because we're not sending emails in the tests anymore, I propose that we take more control of the configuration and manually construct an `EmailConfig` without `godotenv.Load()` or a skip flag. E.g. 

```
func TestSendEmails(t *testing.T) {
    conf := config.EmailConfig{
        AdminEmail: "admin@example.com",
        ...
        Testing: true,
    }

    email, err := emails.New(conf)
    require.NoError(t, err)
}
```

So that removes quite a few lines of code! 

Also, I don't know how you feel about the 1 giant test to do all the testing for all the different emails. If it's annoying you, you could consider using a [test suite](https://pkg.go.dev/github.com/stretchr/testify/suite) here - the mock email client makes this a bit easer:

```go
import (
    "testing"
    "github.com/stretchr/testify/suite"
)

type EmailTestSuite struct {
    suite.Suite
    conf config.EmailConfig
}

func (suite *EmailTestSuite) SetupSuite() {
    suite.conf = config.EmailConfig{
        AdminEmail: "admin@example.com",
        …
        Testing: true,
    }
}

func (suite *EmailTestSuite) AfterTest(suiteName, testName string) {
    emails.PurgeMockEmails()
}

func (suite *EmailTestSuite) TestVerifyContactEmail() {
    require := suite.Require()
    email, err := emails.New(suite.conf)
    require.NoError(err)
    ...
}

func TestEmailSuite(t *testing.T) {
    suite.Run(t, &EmailTestSuite{})
}
```

This is totally optional though!

pkg/gds/emails/emails_test.go
```diff
@@ -111,23 +112,42 @@ func TestSendEmails(t *testing.T) {
...
 	dcdata := emails.DeliverCertsData{Name: receipient.Name, VID: "42", CommonName: "example.com", SerialNumber: "1234abcdef56789", Endpoint: "trisa.example.com:443"}
 	msg, err = emails.DeliverCertsEmail(sender.Name, sender.Address, receipient.Name, receipient.Address, "testdata/foo.zip", dcdata)
 	require.NoError(t, err)
 	require.NoError(t, email.Send(msg))
+	require.Len(t, emails.MockEmails, 4)
+	expected, err = json.Marshal(msg)
+	require.NoError(t, err)
+	require.Equal(t, expected, emails.MockEmails[3])
+
+	t.Cleanup(emails.PurgeMockEmails)
```
bbengfort, line 152: Nice use of the cleanup function! I believe this should go at the top of the test function to guarantee it gets run though. I'm not 100% sure, but I suspect that the `require` functions execute a `t.FailNow()` which stops the execution of the test, so if any of the tests fail, the cleanup function won't be run since it's the last function call.

pkg/gds/emails/mock.go
```diff
@@ -0,0 +1,85 @@
...
+
+var MockEmails [][]byte
+
+func PurgeMockEmails() {
+	MockEmails = nil
+}
+
+type mockSendGridClient struct{}
+
+func (c *mockSendGridClient) Send(msg *sgmail.SGMailV3) (rep *rest.Response, err error) {
```
bbengfort, line 20: This is really truly excellent! My first assumption was that you went through the `sendgrid` code base to mimic the validation in the `Send` method of the actual client. However, when I went to that method to take a look, I realized how much more work you actually did! I found the `GetRequestBody` function call from the helpers, but my quick look did not see the chain to do all of the email address validation - really, really well done! 

And this was super important, if we didn't do any email validation then we wouldn't be able to capture possible errors that sendgrid might return to us - you've given us not only address validation but also created a mechanism with which we can validate other parts of the request if needed. Thank you!

pkg/gds/config/config.go
```diff
@@ -86,6 +86,7 @@ type EmailConfig struct {
 	DirectoryID          string `envconfig:"GDS_DIRECTORY_ID" default:"vaspdirectory.net"`
 	VerifyContactBaseURL string `envconfig:"GDS_VERIFY_CONTACT_URL" default:"https://vaspdirectory.net/verify-contact"`
 	AdminReviewBaseURL   string `envconfig:"GDS_ADMIN_REVIEW_URL" default:"https://admin.vaspdirectory.net/vasps/"`
+  EmailTesting         bool   `envconfig:"GDS_EMAIL_TESTING" required:"false"`
```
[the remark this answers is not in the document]
pdeziel: Oops, there as a merge conflict with main that I resolved through the GitHub UI, looks like that was what messed up the formatting.

pdeziel:
@bbengfort I decided to create a test suite for the send email tests following your suggestion, and made other modifications based on your comments.

PR #157, opened 2021-10-19
fix: timeline showing up blank

elysee15:
Screen Shot 2021-10-19 at 21 32 51


pdeziel (approved):
Looks good to me, thanks for making the change!

PR #158, opened 2021-10-20
fix: vertical overflow in pending reviews makes it hard to use controls

elysee15:
Fixes SC-1313

Screen Shot 2021-10-20 at 07 59 18


tl6kk (approved):
Thank you @elysee15. This looks great from my side and I'll go ahead and merge this in.

PR #163, opened 2021-10-22
VASP history reconstruction - DO NOT MERGE

pdeziel:
This code is temporary and supports some devops work that we have to do involving generating audit logs for VASPs that don't exist. The command I've created for this part of gdsutil and uses the GDS_DATABASE_URL (or database_url from the profile) to read the DB, automatically generate missing audit logs, and write them back to the DB. If VASPs already contain any audit log entries, then they are unmodified.

Usage: `gdsutil ldb:populate`

### DIFF
--- a/cmd/gdsutil/main.go
+++ b/cmd/gdsutil/main.go
@@ -58,6 +58,21 @@ func main() {
 	app.Before = loadProfile
 	app.Commands = []*cli.Command{
 		{
+			Name:     "ldb:populate",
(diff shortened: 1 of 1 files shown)


bbengfort (approved):
@pdeziel looks good so far, I'll create a story to execute these changes in the database next sprint and mark the story as complete for this sprint. 

There is one thing I want to discuss - and that's the REJECTED case; I suspect that's covered by the `case vasp.VerifiedOn` but want to make sure.

PR #177, opened 2021-10-29
sc-1616 implement Get RPC

pdeziel:
This implements the Get RPC on the trtl server for retrieving data from honu. This ended up being a bigger PR than I originally thought because I decided to create an interface that `honu.go` and `peers.go` could both use for interacting with the DB. I also added some CLI functionality so that I could test the RPC; the config situation is a bit of a mess right now but that will be the focus of a future PR.


cmd/trtl/main.go
```diff
@@ -268,8 +321,37 @@ func main() {
...
+	var conf config.Config
+	var db *honu.DB
+
+	replicaConfig := honuconfig.ReplicaConfig{
+		PID:    c.Uint64("pid"),
+		Region: c.String("region"),
+	}
+	if db, err = honu.Open(c.String("db"), replicaConfig); err != nil {
+		return cli.Exit(err, 1)
+	}
```
bbengfort, line 333: Could we please move this to `honu.New` -- all we want the CLI to only deal with configuration and execution, it shouldn't be responsible for setup, initialization, or runtime.

cmd/trtl/main.go
```diff
@@ -268,8 +321,37 @@ func main() {
...
+
 	fmt.Println("started the trtl data replication service")
+
+	// Block until the server is stopped.
+	if err = server.BlockUntilShutdown(); err != nil {
+		return cli.Exit(err, 1)
+	}
+	fmt.Println("server terminated")
```
bbengfort, line 354: The `Serve` method should block until shutdown; I'll pull your PR and make these updates.

bbengfort (commented):
@pdeziel very good start! I'm sorry you had to deal with the Store thing. My idea was that the store interface would be temporary until we could move the peers logic in, however, I can see that we're building up too many layers of redirection right now since we're trying to piece things together. I will move the peers code into your branch and firm up some of the comments I made; then you, @rebeccabilbro and I have a sync up on Monday and we can finalize the server model.

pkg/trtl/honu.go
```diff
@@ -11,15 +14,53 @@ import (
...
+					Region:  object.Version.Parent.Region,
+				},
+			},
+		}
+	} else {
+		// Just return the value for the given key.
+		var value []byte
+		log.Debug().Msg(string(in.Key))
+		if value, err = h.db.Get(in.Key); err != nil {
+			return nil, status.Error(codes.NotFound, err.Error())
```
bbengfort, line 57: What happens if the error is not a `NotFound` error?

pkg/trtl/honu.go
```diff
@@ -11,15 +14,53 @@ import (
...
+					Region:  object.Version.Region,
+				},
+				Parent: &pb.Version{
+					Pid:     object.Version.Parent.Pid,
+					Version: object.Version.Parent.Version,
+					Region:  object.Version.Parent.Region,
+				},
+			},
+		}
+	} else {
```
bbengfort, line 52: On line 33 you can `return &pb.GetReply` so that you don't need the `else` here.

pkg/trtl/honu.go
```diff
@@ -11,15 +14,53 @@ import (
...
+		// Just return the value for the given key.
+		var value []byte
+		log.Debug().Msg(string(in.Key))
```
bbengfort, line 55: I do like the idea of adding some log tracing for debugging, but having the log message here is inconsistent since only requests without `ReturnMeta` will log. What if we put the debug message at the top of the request? Also, I think the following format would help us debug things in Google Log explorer a bit easier:

```
log.Debug().Str("key", string(in.Key)).Bool("return_meta", in.Options.ReturnMeta).Msg("Trtl Get")
```

pkg/trtl/honu.go
```diff
@@ -11,15 +14,53 @@ import (
...
-func NewHonuService() *HonuService {
-	return &HonuService{}
+func NewHonuService(s *Server) (*HonuService, error) {
+	return &HonuService{parent: s, db: s.db}, nil
 }
 
 // Get is a unary request to retrieve a value for a key.
 func (h *HonuService) Get(ctx context.Context, in *pb.GetRequest) (out *pb.GetReply, err error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
+	if in.Options.ReturnMeta {
```
bbengfort, line 27: do we need to check if `in.Options` is nil in case of a `nil` panic? It could be that gRPC always gives back an empty struct here, but it would be good to check.

pkg/trtl/honu.go
```diff
@@ -11,15 +14,53 @@ import (
...
 }
 
-func NewHonuService() *HonuService {
-	return &HonuService{}
+func NewHonuService(s *Server) (*HonuService, error) {
+	return &HonuService{parent: s, db: s.db}, nil
 }
 
 // Get is a unary request to retrieve a value for a key.
 func (h *HonuService) Get(ctx context.Context, in *pb.GetRequest) (out *pb.GetReply, err error) {
```
bbengfort, line 26: We should check if a key is supplied, if not return `BadRequest`.

pkg/trtl/honu.go
bbengfort: Also, on other errors, we should probably do some logging. And we should debug log a not found.

pkg/trtl/peers.go
```diff
@@ -2,24 +2,45 @@ package trtl
...
+
+func (s *notImplementedStore) CreatePeer(peer *peers.Peer) (string, error) {
+	return "", errors.New("not implemented")
+}
+func (s *notImplementedStore) DeletePeer(id string) error   { return errors.New("not implemented") }
+func (s *notImplementedStore) ListPeers() iterator.Iterator { return nil }
+func (s *notImplementedStore) AllPeers() ([]*peers.Peer, error) {
+	return nil, errors.New("not implemented")
+}
+
```
bbengfort, line 31: Sorry, you went through all that work to create the store, but this is what I thought you had in mind for a temporary store. I didn't want to add too much redirection by having interfaces on top of interfaces. Let's create a story to move the Peer store code to this file so it can interact with the database directly.

cmd/trtl/main.go
```diff
@@ -307,6 +386,48 @@ func status(c *cli.Context) (err error) {
...
+			Key: key,
+			Options: &pb.Options{
+				ReturnMeta: c.Bool("meta"),
+			},
+		}
+		if resp, err = dbClient.Get(context.TODO(), req); err != nil {
+			return cli.Exit(err, 1)
+		}
+
+		// TODO: Make the output look nicer
```
bbengfort, line 409: Any reason you don't simply want to output the JSON representation of the request? I hesitate to print things nicely, because it makes it harder to pipe things into other commands, for example, if I wanted to save the data to a file, I might `trtl db:get mykey | base64 -d > myfile.blob` -- if we do want to print things nicely, then I propose that we have a `--format` flag that will allow us to select `pprint, json, raw, base64` etc that determines how the data will be output.

cmd/trtl/main.go
```diff
@@ -284,17 +352,28 @@ func validate(c *cli.Context) (err error) {
...
 
-// TODO: Replace with trtl client once implemented
-var trtlClient peers.PeerManagementClient
+var dbClient pb.TrtlClient
+var peersClient peers.PeerManagementClient
 
-// initClient starts a trtl client.
-func initClient(c *cli.Context) (err error) {
+// initDBClient starts a trtl client with a connection to a trtl database.
+func initDBClient(c *cli.Context) (err error) {
```
bbengfort, line 359: Let's discuss connection on Monday with @rebeccabilbro

pkg/trtl/trtl.go
```diff
@@ -17,40 +20,73 @@ import (
...
+	peers.RegisterPeerManagementServer(s.srv, s.peers)
+
+	// TODO: initialize the Replica service
+
 	return s, nil
 }
 
 // Serve gRPC requests on the specified bind address.
 func (t *Server) Serve() (err error) {
+	// TODO: change from enabled to maintenance mode
```
bbengfort, line 76: Let's discuss on Monday - and also let's discuss the tests.

bbengfort (commented):
@pdeziel I've made some updates to the PR to hopefully simplify things; take a look at the changes I made and let me know your thoughts. I've also added more specific comments to the `Get` RPC implementation.

pkg/trtl/honu.go
pdeziel: Oops, I didn't intend to leave that in the PR, I think moving it to the top of the request is a good idea though.

pkg/trtl/honu.go
pdeziel: Good call.

cmd/trtl/main.go
pdeziel: I agree, JSON is probably sufficient for the default case.

pkg/trtl/honu.go
```diff
@@ -23,14 +24,31 @@ func NewHonuService(s *Server) (*HonuService, error) {
...
+func (h *HonuService) Get(ctx context.Context, in *pb.GetRequest) (*pb.GetReply, error) {
+	var err error
+
+	if len(in.Key) == 0 {
+		log.Error().Msg("missing key in Trtl Get request")
+		return nil, status.Error(codes.InvalidArgument, "key must be provided in Get request")
+	}
+	if in.Options == nil {
+		log.Error().Msg("missing options in Trtl Get request")
+		return nil, status.Error(codes.InvalidArgument, "options must be provided in Get request")
```
bbengfort, line 36: I'm not sure I agree that options should be required - but perhaps we could revisit and discuss?

bbengfort (approved):
@pdeziel thank you for making those changes and adding the TODO for checking the honu not found error.

PR #180, opened 2021-10-31
Test fixtures

rebeccabilbro:
The PR adds a script to `fixtures` for test data generation and adds the gzipped data to `pkg/gds/testdata` for use in testing. 

`fakerize.py` is a very simple script is designed to create synthetic records for testing purposes, since true VASP records contain sensitive information such as cryptographic signatures and PII. Currently the script is able to create synthetic VASP records, but could be expanded to create synthetic certificate requests, etc. If it is something we find ourselves using a lot, we could convert it to a simple CLI app, but that seemed excessive for the current requirements. The records are synthesized using templates available in the `templates` folder as well as some 3rd party Python libraries [`Faker`](https://faker.readthedocs.io/en/master/) and [`lorem`](https://pypi.org/project/lorem/). Ensure you have install the requirements with `pip install -r requirements.txt`. Then run `python fakerize.py`. To modify the features of the synthetic VASPS, such as adding a new state, modify the `FAKE_VASPS` and add the new state to `STATE_CHANGES`.

The test data currently includes only synthetic VASPs, but they demonstrate what I believe to the full range of VASP states, including synthetic notes and audit logs. 

The only thing not included was email logs at the user level, since we do not yet have a template for these. Note that I had to modify `.gitignore` to commit to `fixtures`; I recommend reverting it to the previous state after merging.


.gitignore
```diff
@@ -15,5 +15,10 @@
 # vendor/
 
 # Local configuration
-fixtures/
+fixtures/backups
+fixtures/certs
+fixtures/creds
+fixtures/db
+fixtures/db.old
```
bbengfort, line 22: I agree that we should revert this to its previous state after merging; I feel like it's easier to maintain globally ignored directories with specific exceptions. I think `!fixtures/datagen` would not ignore the datagen file in fixtures, even if `fixtures/` was ignored, though I'm not 100% sure about the syntax. We could also use `git add -f fixtures/myfile` to force add a file even if it is in .gitignore, if that's easier.

fixtures/datagen/README.md
```diff
@@ -0,0 +1,6 @@
+## README
```
bbengfort, line 1: Thank you for adding the README! I know our future selves will be happy about this in a few months.

fixtures/datagen/fakerize.py
```diff
@@ -0,0 +1,446 @@
...
+def fake_legal_name(vasp):
+    """
+    Given a string representing a VASP's name, return a valid
+    dictionary for the faked name identifiers
+    """
+    return {
+        "name_identifiers": [
+            {
+                "legal_person_name": vasp,
+                "legal_person_name_identifier_type": "LEGAL_PERSON_NAME_TYPE_CODE_SHRT",
```
bbengfort, line 131: Could we change this to `LEGAL_PERSON_NAME_TYPE_CODE_LEGL` please? Sorry I didn't catch this before, the legal name is required, and it affects some of the name finding algorithms we use for the front end.

bbengfort (approved):
@rebeccabilbro well done, I know this was a lot of detailed work, and the synthetic data looks excellent! I just had one minor comment, feel free to merge when you're ready!

PR #181, opened 2021-11-01
fix routing and nil map

pdeziel:
This fixes two issues @elysee15 discovered with the review notes endpoint.

1. The notes endpoint was configured at `v2/:vaspID/notes`, it should be `v2/vasps/:vaspID/notes`.
2. The review notes handlers run into a nil exception when a VASP contains no review notes.


bbengfort (approved):
@pdeziel well done on sorting this out! Sorry we didn't catch it in the review of the original notes PR, hopefully this will help unblock @elysee15.

PR #183, opened 2021-11-02
fix: differentiate headers color in vasp details

elysee15:
Fixes SC-1776
Screen Shot 2021-11-04 at 11 36 05


tl6kk:
Hi @elysee15, I guess you are trying to highlight the structure of the detail page with this PR. For that purpose, I'll just go ahead and list a few proposals for changes so that we could capture all those details once for all. I'm using this image as a reference: 
GDS Admin UI - Detail Box

So in the CoinFlip screenshot you included here:
- [ ] From the top level, maybe we could make the `CoinFlip` as the color black?
- [ ] Then on the next level, the `Business details`, `TRISA details`, and `Contacts` should also be color black. (The two `details` better be `Details` too)
- [ ] Then the `Name identifiers`, `National identification`, `Country of registration`, `Customer number`, `Business categories`, `Established On`, `Address(es)`, and `Business Address` should be color grey but in bold (the same as what the `Business Address` currently is). And could we capitalize the first letter in all these words (e.g. `identifiers` to `Identifiers`, and we could leave `of` as `of`) ?
- [ ] The `ID`, `Common name`, `Endpoint`, and `Registered directory` are already in good shape, but maybe we could change `name` into `Name`, and `directory` into `Directory`; also `ID` into `VASP ID`. 
- [ ] Then the `Issued by`, `National Identification Type`, and `LEIX` could be **not** bold. And all the regular texts like these, such as `(US) by authority IRS` and `4957 Oakton St` should be in the same font and size.
- [ ] The `Legal`, `Short`, `Tax ID`, and `Business Entity` badge should also be in the same font as all other regular texts. Also not bold.
- [ ] And then at the end, the `First Listed` `Last Updated` and `Verified On` should be in black, in bold, and the dates below should also be in black but not bold.

tl6kk (approved):

PR #184, opened 2021-11-02
fix: registrations statuses donut chart doesn't show proper colors

elysee15:
Screen Shot 2021-11-02 at 16 37 47
Fixes SC-1787


tl6kk:
I have 1 Rejected and 2 Verified in my database, and here is what I got from the UI
Screen Shot 2021-11-03 at 5 19 43 PM

Regarding the color of the pie chart, is there a way to display the color correctly no matter how many types of statuses we have or don't have? We could maybe use the same color code as in the timeline chart:
Screen Shot 2021-11-03 at 5 26 03 PM
so that for example, Email Verified is always green and Issuing Certificate is always yellow.

Although the pie chart can display up to all 9 statuses, and then the icons below should always be 3: Pending, Rejected, and Verified. I think the color of the Pending as yellow, Rejected as red and Verified as blue is good as it is. This is how these 3 percentages can be calculated: 
Screen Shot 2021-11-03 at 5 31 05 PM
☝️ this is the requirements I got from the original story sc-1304. Please let me know if we have got updated requirements.

PR #185, opened 2021-11-02
fix uninitialized pointer

pdeziel:
This fixes an uninitialized pointer in `ListReviewNotes` which is the actual bug that @elysee15 is getting the 500 error from when calling the notes endpoint.


rebeccabilbro (approved):
Thank you for quickly identifying the bug and it's fix @pdeziel @elysee15!

PR #187, opened 2021-11-02
trtl test suite

pdeziel:
This is a work in progress PR for the trtl test suite. @rebeccabilbro and I put together an initial test suite that generates a leveldb in `testdata` and a test for the Get RPC. While this technically works it needs some more work:

- [x] Database should be gzipped into `testdata` and extracted for testing the suite
- [x] Figure out what the exact functionality of the --update flag should be and implement it
- [x] Fill out the test fixture DB with more peer and object data
- [x] Tests currently require manually setting the environment variables `TRTL_DATABASE_URL` `TRTL_PID` and `TRTL_REGION`


rebeccabilbro:
@bbengfort A couple questions that came up as @pdeziel and I were working through the requirements of the story:

- the `-update` use case was not clear to us for this case. For sure, some tests should require a fresh reload of the db from the golden files in the gzip, but it wasn't clear to us which tests should then generate new fixtures that we'd then want to add to the gzip. What is the desired effect of running `-update` on, say a test for `Get` or `Delete`?
- the requirements are for the fixtures to include some peers and some "data objects" -- are there some other specific types of objects (e.g. other than peers) you would like included in particular?

bbengfort:
@rebeccabilbro so that I can speak to the requirements, I'll address your questions in SC-1624

pkg/trtl/trtl_test.go
```diff
@@ -1,39 +1,323 @@
...
+// Note: This also generates a temporary directory which the suite teardown
+// should clean up.
+func generateDB(s *trtlTestSuite) {
+	require := s.Require()
+	db, err := leveldb.OpenFile(s.db, nil)
+	require.NoError(err)
+	defer db.Close()
+
+	loadFixtures(s)
+	fmt.Println(dbFixtures["alice"])
```
rebeccabilbro, line 84: print line can be removed?

pkg/trtl/trtl_test.go
```diff
@@ -1,39 +1,323 @@
...
+}
+
+func (s *trtlTestSuite) SetupSuite() {
+	var err error
+	require := s.Require()
+	s.gzip = filepath.Join("testdata", "db.tar.gz")
+	s.db, err = ioutil.TempDir("testdata", "db*")
+	require.NoError(err)
+
+	// Only regenerate the test database if requested or it doesn't exist.
```
rebeccabilbro, line 185: I wonder if we should also regenerate the test db if there's something wrong with it; e.g. it's become corrupted and can't be extracted or read from or something?

pkg/trtl/trtl_test.go
```diff
@@ -1,39 +1,323 @@
...
+	if _, err = os.Stat(s.gzip); *update || os.IsNotExist(err) {
+		generateDB(s)
```
rebeccabilbro, line 187: since there's a lot going on in the `generateDB` function, would you mind adding a comment here to remind us that it also calls `loadFixtures` under the hood?

pkg/trtl/trtl_test.go
```diff
@@ -1,39 +1,323 @@
...
-	_, err = client.Get(ctx, &pb.GetRequest{Key: []byte("foo")})
-	require.EqualError(t, err, "rpc error: code = Unimplemented desc = not implemented")
 
-	err = server.Shutdown()
-	require.NoError(t, err)
+	// Retrieve a value from a reserved namespace - should fail.
+	_, err = client.Get(ctx, &pb.GetRequest{
+		Namespace: "default",
+		Key:       []byte(object.Key),
+	})
```
rebeccabilbro, line 238: Thank you for adding test coverage for the new reserved namespaces we introduced in the last merge!

rebeccabilbro (approved):
This looks really good @pdeziel! Thank you for sticking with this as we figured things out together 🙂 . I have left a couple small inline comments, but I think this is pretty close to done. 

Insofar as your question about the redundancy of storing both the json file and the gzipped db in the repo, it seems like the right solution to me: the json file is the "source of truth" in that it is the one that we'll edit as GDS evolves, and as we come up with more edge cases we want to test. It's way easier to edit json! By contrast, the gzipped test db is what will allow us to test `trtl` functions (which expect to engage with a db not json). The `-update` flag is a way to synchronize the test db with any manual updates that have been made to our json. 

To that end, I was thinking that we might want to add a brief README to the `testdata` folder that explains the above so that we remember how to run the tests in the future 😅  We might also want to mention that the tests may *also* require updates as the protocol buffers change (e.g. if the definition of `object.Version` or `object.Object` should change in the protos). So we might also want to recommend that the test runner should always regenerate the protocol buffers before running the tests, to make sure we capture any breaking changes.

pkg/trtl/trtl_test.go
pdeziel: Yes, thanks for catching that.

pkg/trtl/trtl_test.go
pdeziel: Good idea, I can make that change.

PR #189, opened 2021-11-03
fix review note storage

pdeziel:
This fixes an issue where the review notes were not actually being stored in the database once the in-memory objects were updated.


bbengfort (approved):
This update makes sense to me - I didn't do any tests locally but since we have a story to add this to the integration tests, I'm comfortable moving forward.

PR #190, opened 2021-11-03
trtl handles namespaces, some are reserved

rebeccabilbro:
This PR enables trtl to handle namespaces, by prepending the namespace to the key before executing the honu requests. It also introduces three reserved namespaces; honu will return an error if a client attempts to access keys in any of these namespaces:
- "peers": used to store replication information
- "index": contains all index mappings, e.g. names, websites, etc.
- "sequence": to be used for backward compatibility

Note that this PR adds namespace fields to several message definitions in `trtl.proto`, which change the field ordering and will not be backwards compatible.  I went back and forth about whether to rename `prefix` to `namespace` in `IterRequest`, but I ended up leaving it as is, figuring that prefixes could include namespaces but might be more comprehensive? I'm happy to go back and change this to namespace though, if we feel it's effectively identical, since it's probably better to have fewer different ways of referencing the same thing:

```proto
message IterRequest {
    bytes prefix = 1;       // the prefix to range over, if nil all objects are returned
    Options options = 2;
}
```

There is some conflict with #187, due to both the proto file and also the Get test introduced in that PR, which does not yet anticipate the namespace prepending, but it should be fairly straightforward to resolve.


pkg/trtl/honu.go
```diff
@@ -27,6 +28,10 @@ func NewHonuService(s *Server) (*HonuService, error) {
 func (h *HonuService) Get(ctx context.Context, in *pb.GetRequest) (*pb.GetReply, error) {
 	var err error
 
+	if _, found := reservedNamespaces[string(in.Namespace)]; found {
+		log.Error().Msg("cannot use reserved namespace")
+		return nil, status.Error(codes.PermissionDenied, "cannot use reserved namespace")
+	}
```
bbengfort, line 34: For ease of managing logging errors, I'd recommend that we `log.Warn()` here rather than error - the user has supplied a bad request, but can change the namespace and try again with no action on our part, whereas errors are more severe alerts that require our attention.

Also, the `codes.PermissionDenied` was an interesting choice, and it makes sense. Generally I think of `PermissionDenied` as having a specific user context, e.g. user A is permission denied but user B is not, but I can see how this would make sense, permission is denied for all users to access this resource and the same request should not be tried again. What would you think about using `INVALID_ARGUMENT`, e.g. you can retry as long as you use a different namespace?

pkg/trtl/honu.go
```diff
@@ -35,17 +40,25 @@ func (h *HonuService) Get(ctx context.Context, in *pb.GetRequest) (*pb.GetReply,
...
 		return nil, status.Error(codes.InvalidArgument, "options must be provided in Get request")
 	}
-	log.Debug().Str("key", string(in.Key)).Bool("return_meta", in.Options.ReturnMeta).Msg("Trtl Get")
+
+	var key []byte
+	if len(in.Namespace) > 0 {
+		key = prepend(in.Namespace, in.Key)
+	} else {
+		key = in.Key
+	}
```
bbengfort, line 49: Should we allow requests without namespaces? On the one hand, I can see a use case for adding keys without a prefix; however it would also mean that we'd have to allow the user to iterate over all keys in the database including the reserved namespaces. 

One way that we could avoid this is to have a `"default"` namespace that automatically gets applied if no namespace is provided. We would add `"default"` to our list of reserved namespaces and then automatically apply it in the case where no namespace is specified, making it optional. Alternatively we could simply require a namespace and return an invalid argument if no namespace is specified. 

Thoughts? We could open a new story for this if it requires more discussion.

pkg/trtl/honu.go
```diff
@@ -109,3 +122,13 @@ func (h *HonuService) Cursor(in *pb.CursorRequest, stream pb.Trtl_CursorServer)
...
+
+// prepend the namespace to the key
+func prepend(namespace, key []byte) []byte {
+	return bytes.Join(
+		[][]byte{
+			namespace,
+			key,
+		}, []byte("::"),
+	)
+}
```
bbengfort, line 134: Thank you for maintaining the `::` separator between the namespace and the key!

proto/trtl/pb/v1/trtl.proto
```diff
@@ -29,7 +29,8 @@ service Trtl {
 
 message GetRequest {
     bytes key = 1;
-    Options options = 2;
+    bytes namespace = 2;
```
bbengfort, line 32: Having the `namespace` be a `bytes` object makes sense for parity with the key and for making key operations such as prepend easier. Note, however that we've identified the namespace in the [honu code as a string](https://github.com/rotationalio/honu/blob/main/proto/object/v1/object.proto#L21). We'll either have to change that field in Honu or convert the bytes into a string before saving it. 

The difference between `string` and `bytes` in code is that `string` is much more restrictive about what bytes can be included in the slice; namely a string can only contain "printable" bytes via the UTF-8 specification, whereas the byte slice can contain any byte value. Because strings in Go are UTF-8, it is fairly straight forward to cast a `string` to `[]byte` and vice versa. However, slices cannot be used as keys in a `map` since slices do not have equality defined. I recommend we choose our namespace type based on what is easiest for both the user and for us in the code.

bbengfort (approved):
@rebeccabilbro this looks great, thank you so much for adding namespaces to Trtl! I had one minor comment and then some other comments that might require discussion beyond this PR. We can open tickets for those things if necessary. 

As to the question of adding `namespace` to `IterRequest`, I think we should in addition to the `prefix`. You're absolutely correct that prefixes could include namespaces, but the user would then also have to know about the `::` separator. Additionally, if the prefix was totally generic, the user could also iterate over the reserved namespaces, which I'm not sure we want to have happen. This also gets into the question about whether namespaces should be required or not. 

If we define `IterRequest` with both the prefix and the namespace then the following are valid:

```go
// Iterate over all VASPs in the namamespace
req := &IterRequest{Prefix: nil, Namespace: []byte("vasps")}

// Iterate over all Certificate Requests that start with `1d`
req := &IterRequest{Prefix: []byte("1d"), Namespace: []byte("certreqs")}
```

I think this makes sense, though the first option would be the most common use case as you suggested. What do you think?

pkg/trtl/honu.go
rebeccabilbro: I figured we'd want to still have the option not to include namespaces?

pkg/trtl/honu.go
rebeccabilbro: Gotcha - it looked (below) like we were erring on the side of erroring rather than warning, so just trying to be consistent with what was already there. `PermissionDenied` was meant in the sense of "you do not have access to this namespace"

pkg/trtl/namespaces.go
```diff
@@ -0,0 +1,9 @@
+package trtl
+
+// Reserved namespaces that cannot be used by the caller since they are in use by trtl.
+// If necessary this can be moved to configuration in the future.
+var reservedNamespaces = map[string]struct{}{
+	"peers":    struct{}{},
+	"sequence": struct{}{},
+	"index":    struct{}{},
+}
```
pdeziel, line 9: Is it a common go pattern to populate value-less maps with anonymous structs or are we just leaving room for future namespace configuration?

pdeziel (approved):
Good PR, looks like it shouldn't be too difficult for me to merge this into the test suite branch.

pkg/trtl/honu.go
bbengfort: That's a good point, we should also change the errors below to `log.Warn()`. @pdeziel did we discuss that in the `Get` PR? I thought we had talked about not making `Options` required since it is a better user experience to provide `nil` rather than an empty object? 

Ok, I'm fine with sticking to `PermissionDenied` if you think that's what it should be.

pkg/trtl/honu.go
bbengfort: I think we have to answer the following questions:

1. Do we want users to have access to _all_ data in the database, including data from the reserved namespaces? If no, making namespace optional means having a `default` namespace, if yes, then namespace can be omitted.
2. Do we think of Trtl as a strict key/value store or something more like a document database (which is how GDS uses Trtl)? If the former, we should not require namespaces, if the latter, we should - it's a "collection" in Mongo terms.

pkg/trtl/namespaces.go
bbengfort: Great question, this is an idiomatic and lightweight mechanism to [store a set in Go](https://golangbyexample.com/set-implementation-in-golang/). The empty anonymous structs don't allocate any memory unlike a `bool` which allocates a bit.

pkg/trtl/honu.go
pdeziel: I remember seeing your comment about that but I think I missed it in the PR. I agree it's a better user experience to provide `nil` if you just want the value.

pkg/trtl/namespaces.go
pdeziel: Cool!

pkg/trtl/honu.go
rebeccabilbro: Trying to pick up the thread from above @bbengfort @pdeziel -- does that mean you guys want me to remove:

```golang
if in.Options == nil {
	log.Error().Msg("missing options in Trtl Get request")
	return nil, status.Error(codes.InvalidArgument, "options must be provided in Get request")
}
```

or change to Warn and `return nil, nil`?

PR #193, opened 2021-11-04
sc-1804 return note from CreateNote endpoint

pdeziel:
This changes the `CreateReviewNote` endpoint a bit so that it returns the newly created note instead of just the ID slug, which makes the review note handling on the front end a bit easier.

### DIFF
--- a/pkg/gds/admin/v2/client_test.go
+++ b/pkg/gds/admin/v2/client_test.go
@@ -412,7 +412,7 @@ func TestRetrieveVASP(t *testing.T) {
 }
 
 func TestCreateReviewNote(t *testing.T) {
-	fixture := &admin.CreateReviewNoteReply{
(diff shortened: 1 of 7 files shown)


pkg/gds/admin/v2/client_test.go
```diff
@@ -412,7 +412,7 @@ func TestRetrieveVASP(t *testing.T) {
 }
 
 func TestCreateReviewNote(t *testing.T) {
-	fixture := &admin.CreateReviewNoteReply{
+	fixture := &admin.ReviewNote{
```
bbengfort, line 415: Will we want the test to return a full review note too?

bbengfort (approved):
@pdeziel thank you for making this change, hopefully it unblocks @elysee15 on the front end!

PR #194, opened 2021-11-04
sc-1621 batch RPC

pdeziel:
This implements the batch RPC handler on the trtl server. I think the tricky part here is the error handling. The pattern I tried to follow was that if the server has a problem reading the stream, then it is considered a "fatal" error and the handler returns since it probably can't send back a reply. The other errors have to do with specific requests in the batch so the handler can continue, but the errors are recorded in the `BatchReply`.


bbengfort:
One thing that occurs to @rebeccabilbro and I as we're reviewing the protocol buffers, is that we'll have to add options to the `BatchRequest`, and expect the options in the first request if not all requests. There are two primary issues (one of which you may have solved, we haven't looked yet):

1. How do we determine when the batch is complete?
2. Should the batch be treated as a transaction?

For 2 - we'd have to implement some kind of manual locking, otherwise the caller would have to be aware that concurrent operations could happen between batch requests (the latter is more of a documentation issue). Note that GDS does not require transactions. E.g. see: [TODO](TODO).

pkg/trtl/honu.go
```diff
@@ -111,8 +112,74 @@ func (h *HonuService) Iter(ctx context.Context, in *pb.IterRequest) (out *pb.Ite
...
+	out := &pb.BatchReply{}
+	for {
+		// Read the next request from the stream.
+		in, err := stream.Recv()
+		if err == io.EOF {
+			return stream.SendAndClose(out)
+		}
+		if err != nil {
+			return status.Error(codes.Internal, err.Error())
+		}
```
bbengfort, line 126: These are the loop terminal conditions, and it appears that "successful" streams require the client to close the stream and send an `io.EOF` error. This pattern could work for detecting when the batch is complete, but I'm not sure how to implement it on the client side. Have you tested this functionality with a local server?

pkg/trtl/honu.go
```diff
@@ -111,8 +112,74 @@ func (h *HonuService) Iter(ctx context.Context, in *pb.IterRequest) (out *pb.Ite
 	return nil, status.Error(codes.Unimplemented, "not implemented")
 }
 
-func (h *HonuService) Batch(stream pb.Trtl_BatchServer) (err error) {
-	return status.Error(codes.Unimplemented, "not implemented")
+// Batch is a client-side streaming request to issue multiple commands, usually Put and Delete.
+func (h *HonuService) Batch(stream pb.Trtl_BatchServer) error {
+	out := &pb.BatchReply{}
```
bbengfort, line 117: There is a hidden place where errors might come from, and that's `stream.Context()` which returns a context object. That context has a `ctx.Done()` channel, which will inform you if the stream is closing, usually because of a deadline exceeded error. That error may not show up as a non-EOF error in `Recv`, so some experimentation is required to determine how we want to handle receiving messages and dealing with deadlines from the database. We can discuss this during our next sync.

pkg/trtl/honu.go
```diff
@@ -111,8 +112,74 @@ func (h *HonuService) Iter(ctx context.Context, in *pb.IterRequest) (out *pb.Ite
...
+				Error: "missing request field",
+			})
+			continue
+		}
+
+		// Process the request.
+		switch in.Request.(type) {
+		case *pb.BatchRequest_Put:
+			var reply *pb.PutReply
+			if reply, err = h.Put(stream.Context(), in.Request.(*pb.BatchRequest_Put).Put); err != nil || !reply.Success {
```
bbengfort, line 142: I think it's a bit simpler to replace `in.Request.(*pb.BatchRequest_Put).Put` with `in.GetPut()` - which should be available from the generated code. That method does essentially what you're doing (also with nil checks), but is a bit more readable. 

Also, I like that you're reusing the `h.Put` RPC call, but note that in the future we might have to break that call into two, an outer `h.Put` that handles thread safety and an inner `h.put` that can be called from either `Batch`, `Sync`, or `Put`.

pkg/trtl/honu.go
```diff
@@ -111,8 +112,74 @@ func (h *HonuService) Iter(ctx context.Context, in *pb.IterRequest) (out *pb.Ite
...
+				}
+				out.Errors = append(out.Errors, &pb.BatchReply_Error{
+					Id:    in.Id,
+					Error: errMsg,
+				})
+				continue
+			}
+		case *pb.BatchRequest_Delete:
+			var reply *pb.DeleteReply
+			if reply, err = h.Delete(stream.Context(), in.Request.(*pb.BatchRequest_Delete).Delete); err != nil || !reply.Success {
```
bbengfort, line 158: This would be `in.GetDelete()` see comment above.

pkg/trtl/honu.go
```diff
@@ -111,8 +112,74 @@ func (h *HonuService) Iter(ctx context.Context, in *pb.IterRequest) (out *pb.Ite
...
+		default:
+			out.Failed++
+			out.Errors = append(out.Errors, &pb.BatchReply_Error{
+				Id:    in.Id,
+				Error: "unknown request type",
+			})
+			continue
+		}
+
+		out.Successful++
```
bbengfort, line 181: Since you're relying on `continue` statements for this logic, it's probably wise to add an comment that warns future us to make sure the case handling has `continue` on failure logic.

bbengfort (approved):
I think this is a good prototype of the batch RPC - and demonstrates some of the questions that we'll need to answer for the next revision. We can ask these questions in story form and address them later, or we can discuss as a group in our next sync, let me know your preference. 

We will need to create tests for this RPC and make sure we understand how to send EOF so that the client gets a reply from the client code. Do you want to do that in a different PR/story or keep working on this PR?

pkg/trtl/honu.go
pdeziel: We do have a story for adding Batch and Cursor tests so I didn't want to overlap the scope of this too much, but I do think it's worth it to add a simple test to this PR to figure out what the client side would look like.

bbengfort: Excellent, thank you; and thank you for finding the `CloseAndRecv` function, I had no idea about that but it makes a lot of sense in client-side and server-side streaming!

pkg/trtl/trtl_test.go
```diff
@@ -341,3 +341,60 @@ func (s *trtlTestSuite) TestGet() {
...
+	reply, err := stream.CloseAndRecv()
+	require.NoError(err)
+	require.Equal(int64(len(requests)), reply.Operations)
+	require.Equal(int64(len(requests)), reply.Failed)
+	require.Equal(int64(0), reply.Successful)
+	require.Len(reply.Errors, len(requests))
+	for _, e := range reply.Errors {
+		require.Contains(requests, e.Id)
+		require.Equal(requests[e.Id].Id, e.Id)
+		require.Contains(e.Error, "not implemented")
```
bbengfort, line 398: I suppose this will change when we implement `Put`! But this demonstrates what is happening correctly now, thank you for including this code!

bbengfort (approved):
Thank you for addressing those comments and for figuring out how the terminal condition works!

PR #196, opened 2021-11-08
Load locale data for GDS-UI to avoid plural error msg

tl6kk:
This PR adds the `loadLocaleData` function to avoid this error message:

```
index.js:1 Plurals for locale en aren't loaded. Use i18n.loadLocaleData method to load plurals for specific locale. Using other plural rule as a fallback.
```

### DIFF
--- a/web/gds-ui/src/contexts/LanguageContext.js
+++ b/web/gds-ui/src/contexts/LanguageContext.js
@@ -5,7 +5,8 @@ import { messages as messagesEn } from '../locales/en/messages';
 import { messages as messagesDe } from '../locales/de/messages';
 import { messages as messagesFr } from '../locales/fr/messages';
 import { messages as messagesZh } from '../locales/zh/messages';
-import { en, de, fr, zh } from 'make-plural/plurals'
+import { messages as messagesJa } from '../locales/ja/messages';
+import { en, de, fr, zh, ja } from 'make-plural/plurals'
 
 
 const Context = React.createContext();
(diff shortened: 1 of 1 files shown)


web/gds-ui/src/contexts/LanguageContext.js
```diff
@@ -5,6 +5,7 @@ import { messages as messagesEn } from '../locales/en/messages';
 import { messages as messagesDe } from '../locales/de/messages';
 import { messages as messagesFr } from '../locales/fr/messages';
 import { messages as messagesZh } from '../locales/zh/messages';
+import { en, de, fr, zh } from 'make-plural/plurals'
```
rebeccabilbro, line 8: I guess we should also add `jp`?

web/gds-ui/src/contexts/LanguageContext.js
```diff
@@ -17,6 +18,12 @@ export const LanguageStore = ({ children }) => {
       fr: messagesFr,
       zh: messagesZh,
     });
+    i18n.loadLocaleData({
+      en: { plurals: en },
+      de: { plurals: de },
+      fr: { plurals: fr },
+      zh: { plurals: zh }
```
rebeccabilbro, line 25: should we add `jp` here too?

rebeccabilbro (approved):
This looks good @tl6kk &mdash; @bbengfort and I have confirmed that this fix does indeed make the error message go away. 

My main takeaway is that we aren't really using plurals very effectively yet in the GDS UI; for instance, we aren't using the `i18n.plural` call in our app as directed by [Lingui](https://lingui.js.org/guides/plurals.html). Perhaps as a follow-on story we could add a small addition to the [README](https://github.com/trisacrypto/directory/blob/main/web/gds-ui/README.md#i8n-steps) to provide instruction to future contributors to the GDS UI so that they know how to more effectively signal the use of pluralized forms for each of the languages we support?

PR #197, opened 2021-11-08
Translate GDS UI into Japanese

tl6kk:
This PR adds in the Japanese translations of the GDS UI into the `po` file, confirms with the `yarn extract` that no error was given and all translations can be picked up by lingui.js, and runs `yarn compile` to update the `ja/message.js` file.

Also reviewed the translations, this time it came back much better than our last round with the trisa doc : )


web/gds-ui/src/locales/de/messages.po
```diff
@@ -21,7 +21,7 @@ msgstr "Eine Komponente der <0>TRISA</0>-Architektur zur Einhaltung der Cryptocu
 msgid "A phonetic representation of the name by which the legal person is known."
```
tl6kk, line 21: Sorry about these changes in other languages added by the extract command.

rebeccabilbro (approved):
@tl6kk Thank you so much for reviewing the translation from Fiverr &mdash; I'm so relieved to hear that it went more smoothly this time. I see that lingui has made some small automated changes to the po files for the other languages, but it looks like it's just line number references, so nothing big. 

I think the only big thing missing from this PR is an update to the [LanguageSelect.js](https://github.com/trisacrypto/directory/blob/main/web/gds-ui/src/components/select/LanguageSelect.js) file, which needs to include Japanese as a choice in the dropdown menu to ensure that the translated text is available to users. Would you mind adding Japanese to the [current choices](https://github.com/trisacrypto/directory/blob/main/web/gds-ui/src/components/select/LanguageSelect.js#L6-L23)?

PR #199, opened 2021-11-09
GDS Admin API: Audit Log in VASP Detail

bbengfort:
This PR unblocks @elysee15 by actually sending audit log data in the VASP Detail endpoint.

### DIFF
--- a/pkg/utils/pb.go
+++ b/pkg/utils/pb.go
@@ -1,0 +1,34 @@
+package utils
+
+import (
+	"encoding/json"
+
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+// Rewire a protocol buffer message into a generic map[string]interface{} as an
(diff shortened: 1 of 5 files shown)


pkg/utils/pb.go
```diff
@@ -0,0 +1,34 @@
...
+	"encoding/json"
+
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+// Rewire a protocol buffer message into a generic map[string]interface{} as an
+// intermediate step before JSON or YAML marshalling. This is typically unnecessary work
+// and is used as a workaround for multi-protocol systems.
+func Rewire(m protoreflect.ProtoMessage) (out map[string]interface{}, err error) {
```
pdeziel, line 13: To clarify, is the reason we are using the `interface{}` type here so we can reuse this function for multiple types of protocol buffer messages, since we are just passing data back to the user?

pdeziel (approved):
Looks good to me.

PR #200, opened 2021-11-09
sc-1636 GDS test suite setup and teardown

pdeziel:
This adds the initial code for a GDS test suite, which involves maintaining a gzipped test database in a similar manner to what we have implemented for trtl. The major difference is that for GDS the test fixtures which are used to generate the "golden" database exist in separate files, so they have to be extracted from the `fakes.tgz` archive first. The test fixtures currently don't unmarshal properly into protocol buffers, so in this PR they are instead just parsed as generic JSON.


fixtures/datagen/fakerize.py
```diff
@@ -302,7 +302,7 @@ def make_verified(vasp, idx, template="fixtures/datagen/templates/verified.json"
     record["identity_certificate"]["subject"]["common_name"] = common_name + ".io"
     record["trisa_endpoint"] = common_name + ".io" + ":123"
     record["website"] = "https://" + common_name + ".io"
-    record["vasp_categories"] = [random.sample(VASP_CATEGORIES, random.randint(1, 4))]
+    record["vasp_categories"] = random.sample(VASP_CATEGORIES, random.randint(1, 4))
```
rebeccabilbro, line 305: Thank you for catching this error in the fakerizer!

pkg/utils/gzip.go
```diff
@@ -0,0 +1,113 @@
...
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+// ExtractGzip extracts a gzipped archive to the specified directory and
+// returns the path to the root level of the extracted archive.
+// Caller must ensure that the destination directory exists.
+func ExtractGzip(file, destDir string) (root string, err error) {
```
rebeccabilbro, line 15: Good idea of moving this out of the tests so that we could reuse it; sounds like something we'll leverage frequently given the testing pattern that we've adopted!

rebeccabilbro (approved):
Looks good @pdeziel and thank you for doing the live walkthrough this morning!

PR #203, opened 2021-11-12
sc-1526 admin API test coverage

pdeziel:
This adds tests for most of the Admin API functions. The goal for these was to test that the Admin API handlers are correctly parsing HTTP requests and returning the appropriate responses. This also fixes some bugs I found along the way:

- `ListVASPs` incorrectly orienting the pages with `page=1` and `page_size=1`
- `CreateReviewNote` failing to accept a custom note ID
- `ListReviewNotes` returning an empty note in the first entry (should fix sc-1814)


pkg/gds/admin_test.go
```diff
@@ -0,0 +1,939 @@
+package gds
```
pdeziel, line 1: I didn't want to put this in the same package as gds but I couldn't figure out a good way to replace the underlying database with the test fixture one because the `db` field is not exported. I'm definitely open to other solutions on how to deal with this.

pkg/gds/admin_test.go
```diff
@@ -0,0 +1,939 @@
...
+			ExpiresAt: time.Now().Unix(),
+		},
+	}
+	accessToken := jwt.NewWithClaims(jwt.SigningMethodRS256, claims)
+	refreshToken, err := a.tokens.CreateRefreshToken(accessToken)
+	require.NoError(err)
+	access, err := a.tokens.Sign(accessToken)
+	require.NoError(err)
+	refresh, err := a.tokens.Sign(refreshToken)
+	require.NoError(err)
```
bbengfort, line 177: It might be helpful to create a method to create credentials on demand, caching them for reuse since we'll need it for many tests. 

```go
func (s *gdsTestSuite) credentials(refresh bool) (ats, rts string, err error) {
	if refresh {
		// Delete cached access/refresh tokens 
		s.accessToken = ""
		s.refreshToken = ""
	}

	if s.accessToken == "" || s.refreshToken == "" {
		// Regenerate credentials
		creds := map[string]interface{}{
			"sub":     "102374163855881761273",
			"hd":      "rotational.io",
			"email":   "kate@rotational.io",
			"name":    "Kate Holland",
			"picture": "https://foo.googleusercontent.com/test!/Aoh14gJceTrUA",
		}
		accessToken, err := s.admin.tokens.CreateAccessToken(creds)
		if err != nil {
				return "", "", err
		}

		refreshToken, err := s.admin.tokens.CreateRefreshToken(accessToken)
		…

		s.accessToken = s.admin.tokens.Sign(accessToken)
		s.refreshToken = s.admin.tokens.Sign(refreshToken)
	}

	return s.accessToken, s.refreshToken, nil
}
```

pkg/gds/admin_test.go
```diff
@@ -0,0 +1,939 @@
...
+		"Authorization": "Bearer " + access,
+	}
+	res = s.doRequest(request, nil)
+	require.Equal(http.StatusOK, res.StatusCode)
+	// Double cookie tokens should be set
+	cookies := res.Cookies()
+	require.Len(cookies, 2)
+	for _, cookie := range cookies {
+		require.Equal(a.conf.CookieDomain, cookie.Domain)
+	}
```
bbengfort, line 241: Good test!

pkg/gds/admin_test.go
```diff
@@ -0,0 +1,939 @@
...
+}
+
+// Test that the Autocomplete endpoint returns the correct response.
+func (s *gdsTestSuite) TestAutocomplete() {
+	require := s.Require()
+	vasps := []string{
+		"d9da630e-41aa-11ec-9d29-acde48001122",
+		"d9efca14-41aa-11ec-9d29-acde48001122",
+	}
+	WriteVASPs(s, vasps)
```
bbengfort, line 285: Ok, I totally see why you're trying to limit the response back, but I am ok with us not testing exact responses. I'm perfectly happy with using `require.Len` to check we get the expected number back, and maybe a spot check with `require.Contains` a few things we're expecting. We want to be a bit flexible to handle changes in the future.

pkg/gds/admin_test.go
```diff
@@ -0,0 +1,939 @@
...
+	}
+	require.Equal(expected, actual)
+}
+
+// Test that the Autocomplete endpoint returns the correct response.
+func (s *gdsTestSuite) TestAutocomplete() {
+	require := s.Require()
+	vasps := []string{
+		"d9da630e-41aa-11ec-9d29-acde48001122",
+		"d9efca14-41aa-11ec-9d29-acde48001122",
```
bbengfort, line 283: Duplicate on purpose?

pkg/gds/admin_test.go
```diff
@@ -0,0 +1,939 @@
...
+	}
+	require.Equal(expected, actual)
+}
+
+// Test the ListVASPs endpoint.
+func (s *gdsTestSuite) TestListVASPs() {
+	require := s.Require()
+	vasps := []string{
+		"d9da630e-41aa-11ec-9d29-acde48001122",
+		"d9efca14-41aa-11ec-9d29-acde48001122",
```
bbengfort, line 316: Duplicate on purpose?

pkg/gds/admin_test.go
```diff
@@ -0,0 +1,939 @@
...
+	require.Len(actual.VASPs, 0)
+}
+
+// Test the RetrieveVASP endpoint.
+func (s *gdsTestSuite) TestRetrieveVASP() {
+	require := s.Require()
+	vasps := []string{
+		"d9da630e-41aa-11ec-9d29-acde48001122",
+	}
+	WriteVASPs(s, vasps)
```
bbengfort, line 419: Is this necessary for a detail endpoint? The size of the database shouldn't matter.

pkg/gds/admin_test.go
```diff
@@ -0,0 +1,939 @@
...
+	request.path = "/v2/vasps?page=2&page_size=1"
+	rep = s.doRequest(request, actual)
+	require.Equal(http.StatusOK, rep.StatusCode)
+	require.Equal(len(vasps), actual.Count)
+	require.Equal(2, actual.Page)
+	require.Equal(1, actual.PageSize)
+	require.Len(actual.VASPs, 1)
+	require.Equal(snippets[1], actual.VASPs[0])
+
+	request.path = "/v2/vasps?page=3&page_size=1"
```
bbengfort, line 404: Excellent series of tests. For future proofing we may want to consider a testTable with test case loop, just so it's easy to adapt. Perhaps we can create a story for refactoring?

bbengfort (commented):
@pdeziel this is excellent work - you've done really well creating a huge amount of test coverage for us, thank you so much! I've made a lot of comments, so I think we should take this in two passes. I stopped at `TestCreateReviewNote`. Let me know if you want to get on a call to discuss.

pkg/gds/gds_test.go
```diff
@@ -22,13 +23,30 @@ var (
...
-	db     string
-	golden string
+	fakes   string
+	db      string
+	golden  string
+	dbVASPs map[string]*pb.VASP
+}
+
+// Overwrite the test database with a subset of VASPs (for small-scale unit testing).
+func WriteVASPs(s *gdsTestSuite, ids []string) {
```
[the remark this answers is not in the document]
pdeziel: I like your idea of the small test fixture, the `WriteVASPs` method is a little hacky and redundant.

pkg/gds/gds_test.go
```diff
@@ -50,12 +70,12 @@ func loadFixtures(s *gdsTestSuite) {
...
-		if strings.HasPrefix(info.Name(), "vasps::") {
-			// TODO: Unmarshal into VASP object when the fixtures are fixed.
+		key := strings.TrimSuffix(info.Name(), ".json")
+		if strings.HasPrefix(key, "vasps::") {
 			vasp := &pb.VASP{}
 			err = protojson.Unmarshal(data, vasp)
 			require.NoError(err)
-			dbVASPs[info.Name()] = vasp
+			s.dbVASPs[strings.TrimPrefix(key, "vasps::")] = vasp
 			return nil
```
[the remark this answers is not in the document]
pdeziel: My focus for the PR was on VASPs but I think it does make sense to help our future selves out by generalizing the map to prefixed database objects.

pkg/gds/gds_test.go
```diff
@@ -1,4 +1,4 @@
-package gds_test
+package gds
```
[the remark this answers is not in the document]
pdeziel: This stems from `admin.go` needing to be in the gds package, so that I can inject the test fixture database into the Admin object. It seemed easier than trying to mock a service, but I definitely understand if we're drawing a hard line at separating the test code and real code into different packages, although we aren't following that rule 100% currently (see `pkg/gds/sectigo/creds_test.go`.

pkg/gds/gds_test.go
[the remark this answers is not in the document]
rebeccabilbro: @pdeziel @bbengfort just a heads up that #210 (introduces cert req data) proposes a modification to extract certs as well; if you prefer to modify the test here, lmk and I can deal with the merge conflict in my branch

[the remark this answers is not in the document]
bbengfort: Why don't we merge #210 first and then rebase into this PR? I think #210 is ready to go, but a bit more work is needed here.

pkg/gds/admin_test.go
pdeziel: They do look very similar, maybe we should be generating more different looking IDs in `fakerize.py` to reduce confusion in writing tests?

rebeccabilbro: @bbengfort @pdeziel - this is my fault; the test uuids are indeed all different, but agree they look super similar. I'm using [a function from `uuid` in the Python standard lib](https://docs.python.org/3/library/uuid.html#uuid.uuid1) to generate these uuids and by default they end up being more similar-looking if they're generated close to the same time.

bbengfort: @rebeccabilbro no worries at all - that's my fault for not noticing that they were actually different!

pkg/gds/admin_test.go
```diff
@@ -0,0 +1,939 @@
...
+	require.Equal(http.StatusUnauthorized, res.StatusCode)
+
+	// Invalid credential
+	request.in = &admin.AuthRequest{
+		Credential: "invalid",
+	}
+	res = s.doRequest(request, nil)
+	require.Equal(http.StatusUnauthorized, res.StatusCode)
+
+	// MAYBE: Can we mock the token validation to test the success path?
```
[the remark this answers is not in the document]
pdeziel: Do you have an example of generating a valid token with the mock? I'm getting stuck on `idtoken.Validate()` with this error:

`{"level":"warn","error":"idtoken: could not find matching cert keyId for the token provided","severity":"WARNING","time":"2021-11-17T20:15:03-06:00","message":"invalid credentials used for authentication"}`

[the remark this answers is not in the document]
pdeziel: I must be missing a piece, here is how I'm generating the access token for the Authenticate request:

```go
admin.conf = &config.AdminConfig{
	CookieDomain: "example.com",
	Oauth: config.OauthConfig{
		GoogleAudience:         "http://localhost",
		AuthorizedEmailDomains: []string{"example.com"},
	},
},
admin.tokens, err = tokens.MockTokenManager()
require.NoError(err)
creds := map[string]interface{}{
  "sub":     "102374163855881761273",
  "hd":      "example.com",
  "email":   "admin@example.com",
  "name":    "Jon Doe",
  "picture": "https://foo.googleusercontent.com/test!/Aoh14gJceTrUA",
}
accessToken, err := admin.tokens.CreateAccessToken(creds)
require.NoError(err)
access, err := admin.tokens.Sign(accessToken)
require.NoError(err)
request.in = &admin.AuthRequest{
  Credential: access,
}
```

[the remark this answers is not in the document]
bbengfort: Ah, gotcha, I realized what the problem is - when we submit credentials in an `Authenticate` request, it's actually supposed to be a Google credential … let's mark this test as TODO and create a story for it for later, it's out of scope for what we're trying to do here. When we do get to implementing this test, we'll have to mock the idtoken validator:

https://github.com/googleapis/google-api-go-client/blob/d50702e3711743974a9f359cf4d8b1b3eb46bbb4/idtoken/validate_test.go#L37

pkg/gds/gds_test.go
```diff
@@ -52,82 +74,136 @@ func loadFixtures(s *gdsTestSuite) {
...
 		return fmt.Errorf("unrecognized prefix for file: %s", info.Name())
 	})
 	require.NoError(err)
-	os.RemoveAll(root)
-	require.NoError(err)
+	require.NoError(os.RemoveAll(root))
+}
+
+// writeObj writes a protobuf object to the given database.
+func (s *gdsTestSuite) writeObj(db *leveldb.DB, key string, obj interface{}) {
```
bbengfort, line 99: @rebeccabilbro and I encountered an issue in the trtl test fixtures, where by circumventing the database insertion mechanisms, our fixtures and expected results did not match. To prevent that from happening to us here, I'm going to update this to directly use the store from the service.

pkg/gds/gds_test.go
```diff
@@ -52,82 +74,136 @@ func loadFixtures(s *gdsTestSuite) {
...
+
 	s.fakes = filepath.Join("testdata", "fakes.tgz")
-	s.golden = filepath.Join("testdata", "db.tgz")
-	s.db, err = ioutil.TempDir("testdata", "db-*")
+	s.dbGzip = filepath.Join("testdata", "db.tgz")
+	s.smallDBGzip = filepath.Join("testdata", "smalldb.tgz")
+	s.dbPath, err = ioutil.TempDir("testdata", "db-*")
+	require.NoError(err)
+	s.smallDBPath, err = ioutil.TempDir("testdata", "smalldb-*")
 	require.NoError(err)
```
bbengfort, line 159: Rather than loading all the fixtures at the beginning of the suite, I'm going to create individual cache functions for `loadSmallFixtures()`, `loadAllFixtures()`, and `loadEmptyFixtures()`, then pair those functions with `reset` counterparts in case the test modifies the database. This will prevent the need for the `BeforeTest` action, and allow us to run subsets of tests without having to load unnecessary fixtures. 

Also, I'm going to do a bit of reorganization if that's ok and try to group the fixture loading at the bottom; let me know if this makes sense or if you think we should create a new file for fixtures, e.g. `testdata.go` or `fixtures_test.go` or something like that.

pkg/gds/admin_test.go
```diff
@@ -0,0 +1,966 @@
...
+	}
+	return c, w
+}
+
+// doRequest is a helper function for making an admin API request and retrieving
+// the response.
+func (s *gdsTestSuite) doRequest(fn gin.HandlerFunc, c *gin.Context, w *httptest.ResponseRecorder, reply interface{}) (res *http.Response) {
+	require := s.Require()
+	// Call the admin function and return the HTTP response
+	fn(c)
```
bbengfort, line 92: Would it be ok if we called this `handler`?

pkg/gds/admin_test.go
```diff
@@ -0,0 +1,966 @@
...
+	path    string
+	headers map[string]string
+	params  map[string]string
+	in      interface{}
+	claims  *tokens.Claims
+}
+
+// makeRequest creates a new HTTP request and returns the gin context along with the
+// http.ResponseRecorder.
+func (s *gdsTestSuite) makeRequest(request *httpRequest) (*gin.Context, *httptest.ResponseRecorder) {
```
bbengfort, line 49: `makeRequest` and `doRequest` are excellent - well done!

pkg/gds/admin_test.go
```diff
@@ -0,0 +1,966 @@
...
+}
+
+// Test that we get a good response from ProtectAuthenticate.
+func (s *gdsTestSuite) TestProtectAuthenticate() {
+	require := s.Require()
+	a := s.initAdmin(s.dbPath)
+
+	request := &httpRequest{
+		method: http.MethodPost,
+		path:   "/v2/foo",
```
bbengfort, line 111: Isn't `ProtectAuthenticate` supposed to be a `GET /v2/authenticate`?

pkg/gds/admin_test.go
```diff
@@ -0,0 +1,966 @@
...
+
+	// Invalid credential
+	request.in = &admin.AuthRequest{
+		Credential: "invalid",
+	}
+	c, w = s.makeRequest(request)
+	res = s.doRequest(a.Authenticate, c, w, nil)
+	require.Equal(http.StatusUnauthorized, res.StatusCode)
+
+	// TODO: Test successful authentication path
```
bbengfort, line 151: Have we created a story for this?

pkg/gds/admin_test.go
```diff
@@ -0,0 +1,966 @@
...
+	a := s.initAdmin(s.dbPath)
+
+	request := &httpRequest{
+		method: http.MethodGet,
+		path:   "/v2/summary",
+	}
+	actual := &admin.SummaryReply{}
+	c, w := s.makeRequest(request)
+	rep := s.doRequest(a.Summary, c, w, actual)
+	require.Equal(http.StatusOK, rep.StatusCode)
```
bbengfort, line 258: Ah, I see - so we're circumventing the middleware by going directly to the handler. Authenticate and CSRF protect are never called. That makes a lot of sense. 

Could we create stories to test endpoints for authenticate and csrf protection required?

bbengfort (approved):
@pdeziel very, very well done - I went through all your admin tests in detail, and I have to say you covered a lot of cases and gave me the feeling of safety that we're testing things extremely well. I think you've also set us up to do more effective GDS testing in the future! 

I am going to make a couple of changes as noted in my comments below, then merge this so that we can get it into this sprint. However on Monday, would you please make sure you go through all the comments and create stories where necessary? 

Thank you so much for your hard work on this!

pkg/gds/admin_test.go
```diff
@@ -0,0 +1,966 @@
...
+		ID:                     vasps[0],
+		AdminVerificationToken: "foo",
+		Accept:                 false,
+	}
+	c, w = s.makeRequest(request)
+	rep = s.doRequest(a.Review, c, w, nil)
+	require.Equal(http.StatusBadRequest, rep.StatusCode)
+
+	// TODO: Add tests for accept and reject paths. This requires VASP fixtures with
+	// a non-empty adminVerificationToken.
```
bbengfort, line 785: Have we created a story for this?

pkg/gds/mock.go
```diff
@@ -0,0 +1,43 @@
+package gds
+
+import (
+	"github.com/trisacrypto/directory/pkg/gds/config"
+	"github.com/trisacrypto/directory/pkg/gds/emails"
+	"github.com/trisacrypto/directory/pkg/gds/store"
+	"github.com/trisacrypto/directory/pkg/gds/tokens"
+)
+
```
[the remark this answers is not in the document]
bbengfort: Actually I needed to do this with the small refactor I did this evening, so don't worry about a story for this.

pkg/gds/admin_test.go
pdeziel: Yes I think that makes more sense.

pkg/gds/admin_test.go
pdeziel: You're right, thanks for catching it.

pkg/gds/admin_test.go
pdeziel: I'll create one.

pkg/gds/admin_test.go
pdeziel: Sure, my goal for these specific tests was to make sure the logic within the handlers is correct when given a context, but I agree we do need to make sure that the authenticated and protected endpoints work from a more external perspective.

pkg/gds/admin_test.go
pdeziel: I'll create one shortly.

pkg/gds/gds_test.go
pdeziel: I think that organization makes sense for now, if we require more bells and whistles in the fixture loading then we can always move it to another file.

PR #205, opened 2021-11-15
mock sectigo server

pdeziel:
This adds a mock for the entire sectigo server rather than just the `client.Do` in the other PR. This allows tests by default to verify the success paths of the sectigo interactions in `sectigo.go` and also provides an option to replace the HTTP handlers with more error handling to test the error paths.


pkg/sectigo/sectigo_test.go
```diff
@@ -1,21 +1,223 @@
...
+	require.NotNil(t, rep)
+}
+
+func (s *SectigoTestSuite) processingInfo(t *testing.T) {
+	rep, err := s.api.ProcessingInfo(42)
+	require.NoError(t, err)
+	require.NotNil(t, rep)
+}
+
+func (s *SectigoTestSuite) download(t *testing.T) {
```
bbengfort, line 128: I suppose we will eventually have to modify this test to actually check the download, but we can get to that during integration tests.

bbengfort (approved):
@pdeziel looks good - well done! I know this one was a bear, but we did have to do some major refactoring; still - we're very well setup for integration tests now!

pkg/sectigo/sectigo_test.go
pdeziel: Thanks for catching this, I've added some lightweight checking for now.

PR #206, opened 2021-11-15
TRTL: configuration and logging

bbengfort:
This PR implements a robust trtl configuration and configures trtl logging for service startup. It also generalizes some of the logging utilities and adds tests for them. Comments inline with the code.

### DIFF
--- a/pkg/gds/config/config.go
+++ b/pkg/gds/config/config.go
@@ -10,16 +10,17 @@ import (
 	"github.com/kelseyhightower/envconfig"
 	"github.com/rs/zerolog"
 	"github.com/trisacrypto/directory/pkg/sectigo"
+	"github.com/trisacrypto/directory/pkg/utils/logger"
 )
 
 // Config uses envconfig to load required settings from the environment and validate
 // them in preparation for running the TRISA Global Directory Service.
 type Config struct {
-	DirectoryID string          `split_words:"true" default:"vaspdirectory.net"`
-	SecretKey   string          `split_words:"true" required:"true"`
-	Maintenance bool            `split_words:"true" default:"false"`
-	LogLevel    LogLevelDecoder `split_words:"true" default:"info"`
-	ConsoleLog  bool            `split_words:"true" default:"false"`
+	DirectoryID string              `split_words:"true" default:"vaspdirectory.net"`
+	SecretKey   string              `split_words:"true" required:"true"`
+	Maintenance bool                `split_words:"true" default:"false"`
+	LogLevel    logger.LevelDecoder `split_words:"true" default:"info"`
+	ConsoleLog  bool                `split_words:"true" default:"false"`
 	GDS         GDSConfig
 	Admin       AdminConfig
 	Database    DatabaseConfig
@@ -183,33 +184,6 @@ func (c SectigoConfig) Validate() error {
 func (c EmailConfig) Validate() error {
 	if c.AdminReviewBaseURL != "" && !strings.HasSuffix(c.AdminReviewBaseURL, "/") {
 		return errors.New("admin review base URL must end in a /")
-	}
-	return nil
-}
-
-// LogLevelDecoder deserializes the log level from a config string.
-type LogLevelDecoder zerolog.Level
-
-// Decode implements envconfig.Decoder
-func (ll *LogLevelDecoder) Decode(value string) error {
-	value = strings.TrimSpace(strings.ToLower(value))
-	switch value {
-	case "panic":
-		*ll = LogLevelDecoder(zerolog.PanicLevel)
-	case "fatal":
-		*ll = LogLevelDecoder(zerolog.FatalLevel)
-	case "error":
-		*ll = LogLevelDecoder(zerolog.ErrorLevel)
-	case "warn":
-		*ll = LogLevelDecoder(zerolog.WarnLevel)
-	case "info":
-		*ll = LogLevelDecoder(zerolog.InfoLevel)
-	case "debug":
-		*ll = LogLevelDecoder(zerolog.DebugLevel)
-	case "trace":
-		*ll = LogLevelDecoder(zerolog.TraceLevel)
-	default:
-		return fmt.Errorf("unknown log level %q", value)
 	}
 	return nil
 }
--- a/pkg/gds/service.go
+++ b/pkg/gds/service.go
@@ -16,9 +16,6 @@ import (
 )
 
 func init() {
-	zerolog.TimeFieldFormat = zerolog.TimeFormatUnix
-	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
-
 	// Initialize zerolog with GCP logging requirements
 	zerolog.TimeFieldFormat = time.RFC3339
 	zerolog.TimestampFieldName = logger.GCPFieldKeyTime
@@ -40,7 +37,7 @@ func New(conf config.Config) (s *Service, err error) {
 	}
 
 	// Set the global level
-	zerolog.SetGlobalLevel(zerolog.Level(conf.LogLevel))
+	zerolog.SetGlobalLevel(conf.GetLogLevel())
 
 	// Set human readable logging if specified
 	if conf.ConsoleLog {
--- a/pkg/trtl/config/config_test.go
+++ b/pkg/trtl/config/config_test.go
@@ -40,13 +40,13 @@ func TestConfig(t *testing.T) {
 	require.NoError(t, err)
 
 	// Test configuration set from the environment
-	require.Equal(t, true, conf.Enabled)
 	require.Equal(t, testEnv["TRTL_REPLICA_BIND_ADDR"], conf.BindAddr)
-	require.Equal(t, uint64(8), conf.PID)
-	require.Equal(t, testEnv["TRTL_REPLICA_NAME"], conf.Name)
-	require.Equal(t, testEnv["TRTL_REPLICA_REGION"], conf.Region)
-	require.Equal(t, 30*time.Minute, conf.GossipInterval)
-	require.Equal(t, 3*time.Minute, conf.GossipSigma)
+	require.Equal(t, true, conf.Replica.Enabled)
+	require.Equal(t, uint64(8), conf.Replica.PID)
+	require.Equal(t, testEnv["TRTL_REPLICA_NAME"], conf.Replica.Name)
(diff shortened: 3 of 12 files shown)


pkg/gds/config/config.go
```diff
@@ -10,16 +10,17 @@ import (
...
+	"github.com/trisacrypto/directory/pkg/utils/logger"
 )
 
 // Config uses envconfig to load required settings from the environment and validate
 // them in preparation for running the TRISA Global Directory Service.
 type Config struct {
-	DirectoryID string          `split_words:"true" default:"vaspdirectory.net"`
-	SecretKey   string          `split_words:"true" required:"true"`
-	Maintenance bool            `split_words:"true" default:"false"`
-	LogLevel    LogLevelDecoder `split_words:"true" default:"info"`
```
bbengfort, line 21: Moved the `LogLevelDecoder` to the `pkg/utils/logger` for use in both GDS and TRTL

pkg/gds/service.go
```diff
@@ -16,9 +16,6 @@ import (
 )
 
 func init() {
-	zerolog.TimeFieldFormat = zerolog.TimeFormatUnix
-	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
```
bbengfort, line 20: This was some stray code that gets overridden in lines 21 and 26 so it needed to be removed.

pkg/gds/service.go
```diff
@@ -40,7 +37,7 @@ func New(conf config.Config) (s *Service, err error) {
 	}
 
 	// Set the global level
-	zerolog.SetGlobalLevel(zerolog.Level(conf.LogLevel))
+	zerolog.SetGlobalLevel(conf.GetLogLevel())
```
bbengfort, line 40: We had this method in the configuration, figured it would be better to use it.

pkg/trtl/config/config_test.go
```diff
@@ -5,18 +5,24 @@ import (
...
+	"TRTL_LOG_LEVEL":                "debug",
+	"TRTL_CONSOLE_LOG":              "true",
+	"TRTL_DATABASE_URL":             "leveldb:///fixtures/db",
+	"TRTL_DATABASE_REINDEX_ON_BOOT": "true",
+	"TRTL_REPLICA_ENABLED":          "true",
+	"TRTL_REPLICA_PID":              "8",
+	"TRTL_REPLICA_NAME":             "mitchell",
+	"TRTL_REPLICA_REGION":           "us-east-1c",
+	"TRTL_REPLICA_GOSSIP_INTERVAL":  "30m",
+	"TRTL_REPLICA_GOSSIP_SIGMA":     "3m",
```
bbengfort, line 25: Generally speaking this is the definitive list of environment variables used for configuration.

pkg/trtl/config/config_test.go
```diff
@@ -33,27 +39,28 @@ func TestConfig(t *testing.T) {
...
 }
 
 func TestRequiredConfig(t *testing.T) {
-	// TODO: Fix this test
-	t.Skip("TestRequiredConfig is erroring: PID required for enabled replica")
-
 	required := []string{
+		"TRTL_DATABASE_URL",
 		"TRTL_REPLICA_PID",
 		"TRTL_REPLICA_REGION",
```
bbengfort, line 65: These are currently the only required configurations that don't have defaults that would be set.

pkg/trtl/config/config_test.go
```diff
@@ -71,6 +78,9 @@ func TestRequiredConfig(t *testing.T) {
 	}
 	t.Cleanup(cleanup)
 
+	// Replica verification is predicated on its being enabled
+	os.Setenv("GDS_REPLICA_ENABLED", "true")
```
bbengfort, line 82: The `ReplicaConfig` is not validated if it is not enabled.

pkg/trtl/config/config_test.go
```diff
@@ -92,10 +102,35 @@ func TestRequiredConfig(t *testing.T) {
...
+	conf.Region = "raleigh"
+	require.EqualError(t, conf.Validate(), "invalid configuration: specify non-zero gossip interval and sigma")
+
+	// Gossip Interval should be required
+	conf.GossipInterval = time.Second * 30
+	require.EqualError(t, conf.Validate(), "invalid configuration: specify non-zero gossip interval and sigma")
 
+	// Gossip Sigma should be required
+	conf.GossipSigma = time.Millisecond * 500
+	require.NoError(t, conf.Validate())
```
bbengfort, line 133: A test not for the environment but replica config specific validation.

pkg/trtl/namespaces.go
```diff
@@ -3,8 +3,8 @@ package trtl
...
 // If necessary this can be moved to configuration in the future.
 var reservedNamespaces = map[string]struct{}{
-	"peers":    struct{}{},
-	"sequence": struct{}{},
-	"index":    struct{}{},
-	"default":  struct{}{}, // if the user does not specify a namespace
+	"peers":    {},
+	"sequence": {},
+	"index":    {},
+	"default":  {}, // if the user does not specify a namespace
```
bbengfort, line 9: My linter suggested this, so I went ahead and accepted the suggestion.

pkg/trtl/trtl.go
```diff
@@ -87,8 +101,14 @@ func (t *Server) Serve() (err error) {
...
-	// Run the Gossip background routine
-	go t.replica.AntiEntropy()
+	// Run management routines only if we're not in maintenance mode
+	if t.conf.Maintenance {
+		log.Warn().Msg("starting trtl in maintenance mode")
+	} else {
+		// These services should not run in maintenance mode
+		// Run the Gossip background routine
+		go t.replica.AntiEntropy()
+	}
```
bbengfort, line 111: This section only looks at maintenance mode now, not if it is enabled or not.

pkg/utils/logger/level_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"testing"
+
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/require"
+	. "github.com/trisacrypto/directory/pkg/utils/logger"
+)
+
+func TestLevelDecoder(t *testing.T) {
```
bbengfort, line 11: Figured we should have test cases since this is in its own package.

pkg/utils/logger/logger_test.go
```diff
@@ -0,0 +1,121 @@
...
+	w.levels[level]++
+	return w.Write(p)
+}
+
+func (w *testWriter) Write(p []byte) (n int, err error) {
+	if err = json.Unmarshal(p, &w.lastLog); err != nil {
+		return 0, err
+	}
+	return len(p), nil
+}
```
bbengfort, line 33: This is a mock so the logger doesn't write to stdout and we can test the JSON output of the logger.

pkg/utils/logger/logger_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+func TestSeverityHook(t *testing.T) {
```
bbengfort, line 35: These tests didn't exist before so I figured I'd try to bump our coverage a bit.

pkg/gds/config/config.go
pdeziel: Good idea!

pkg/trtl/config/config_test.go
```diff
@@ -71,6 +78,9 @@ func TestRequiredConfig(t *testing.T) {
 	}
 	t.Cleanup(cleanup)
 
+	// Replica verification is predicated on its being enabled
```
pdeziel, line 81:
```suggestion
	// Replica verification is predicated on it being enabled
```

pkg/trtl/namespaces.go
pdeziel: Interesting that the GitHub lint didn't catch this since the linter in my VSCode complains about this as well.

pkg/utils/logger/level_test.go
```diff
@@ -0,0 +1,52 @@
...
+		{
+			"iNFo", zerolog.InfoLevel,
+		},
+		{
+			"debug", zerolog.DebugLevel,
+		},
+		{
+			"trace", zerolog.TraceLevel,
+		},
+	}
```
pdeziel, line 37: Good idea with testing the edge cases of the level string parsing!

pkg/utils/logger/logger_test.go
```diff
@@ -0,0 +1,121 @@
...
+	require.Equal(t, uint16(1), tw.levels[zerolog.TraceLevel])
+	require.Contains(t, tw.lastLog, logger.GCPFieldKeySeverity)
+	require.Equal(t, "DEBUG", tw.lastLog[logger.GCPFieldKeySeverity])
+	require.Contains(t, tw.lastLog, logger.GCPFieldKeyMsg)
+	require.Equal(t, "just a trace", tw.lastLog[logger.GCPFieldKeyMsg])
+	require.Contains(t, tw.lastLog, logger.GCPFieldKeyTime)
+	require.NotEmpty(t, tw.lastLog[logger.GCPFieldKeyMsg])
+
+	log.Debug().Msg("is it on?")
+	require.Equal(t, uint16(1), tw.levels[zerolog.TraceLevel])
```
pdeziel, line 58: Should we be checking the `zerolog.DebugLevel` entry here? Same for the Info, Warn, Error, Fatal, etc. below but with the respective log levels.

pdeziel (approved):
Looks good to me, there is one thing in `logger_test.go` which needs to be addressed but after that I think it's good to merge!

pkg/trtl/namespaces.go
bbengfort: Perhaps because it's just a warning?

pkg/utils/logger/logger_test.go
bbengfort: Whoops, you're right - good catch!

pkg/trtl/namespaces.go
rebeccabilbro: my linter also complained, but did not make any suggestions -- @bbengfort what linter extension are you using for vscode?

bbengfort: This I think: https://marketplace.visualstudio.com/items?itemName=golang.Go with the default tools. Could just be a version difference.

PR #207, opened 2021-11-15
sc-1529 add profiles tests

pdeziel:
This adds tests for the profiles management functionality.

### DIFF
--- a/pkg/gds/client/profile_test.go
+++ b/pkg/gds/client/profile_test.go
@@ -1,0 +1,29 @@
+package client_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/trisacrypto/directory/pkg/gds/client"
+)
+
+// Test that OpenLevelDB opens a LevelDB database.
+func TestOpenLevelDB(t *testing.T) {
+	// No DB specified
+	profile := &client.Profile{}
+	_, err := profile.OpenLevelDB()
+	require.Error(t, err)
+
+	// Invalid DB scheme
+	profile = &client.Profile{DatabaseURL: "foo://bar"}
+	_, err = profile.OpenLevelDB()
+	require.Error(t, err)
+
+	// Valid DB DSN
+	profile = &client.Profile{DatabaseURL: "leveldb:///foo"}
+	db, err := profile.OpenLevelDB()
+	require.NoError(t, err)
+	defer db.Close()
+	defer os.RemoveAll("foo")
+}
--- a/pkg/gds/client/profile_test.go
+++ b/pkg/gds/client/profile_test.go
@@ -1,6 +1,7 @@
 package client_test
 
 import (
+	"io/ioutil"
 	"os"
 	"testing"
 
@@ -21,9 +22,12 @@ func TestOpenLevelDB(t *testing.T) {
 	require.Error(t, err)
 
 	// Valid DB DSN
-	profile = &client.Profile{DatabaseURL: "leveldb:///foo"}
+	dbpath, err := ioutil.TempDir("", "gdsdb-*")
+	require.NoError(t, err, "could not create temp directory")
+	defer os.RemoveAll(dbpath)
+
+	profile = &client.Profile{DatabaseURL: "leveldb:///" + dbpath}
 	db, err := profile.OpenLevelDB()
 	require.NoError(t, err)
 	defer db.Close()
-	defer os.RemoveAll("foo")
 }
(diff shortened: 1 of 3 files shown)


pkg/gds/client/profile_test.go
```diff
@@ -0,0 +1,29 @@
...
+	profile = &client.Profile{DatabaseURL: "foo://bar"}
+	_, err = profile.OpenLevelDB()
+	require.Error(t, err)
+
+	// Valid DB DSN
+	profile = &client.Profile{DatabaseURL: "leveldb:///foo"}
+	db, err := profile.OpenLevelDB()
+	require.NoError(t, err)
+	defer db.Close()
+	defer os.RemoveAll("foo")
```
bbengfort, line 28: Since this is actually creating a leveldb could we modify this to use a tmpdir (its likely to have more success on container based tests like those run by GitHub actions):

```go
dbpath, err := os.MkdirTemp("", "testdb")
require.NoError(t, err)
defer os.RemoveAll(dbpath)

profile := &client.Profile{DatabaseURL: fmt.Sprintf("leveldb:///%s", dbpath)}
db, err := profile.OpenLevelDB()
require.NoError(t, err)
defer db.Close()
```

bbengfort: Note that `fmt.Sprintf("leveldb:///%s", dbpath)` will work if `dbpath` is an absolute path, which I'm guessing it is.

pkg/gds/client/profiles_test.go
```diff
@@ -0,0 +1,196 @@
...
+	"strings"
+	"testing"
+
+	"github.com/shibukawa/configdir"
+	"github.com/stretchr/testify/require"
+	"github.com/trisacrypto/directory/pkg/gds/client"
+	"gopkg.in/yaml.v2"
+)
+
+const profileYAML = "profiles.yaml"
```
bbengfort, line 17: Perhaps we should capitalize the constants [`profileYaml` and `profileVersion`](https://github.com/trisacrypto/directory/blob/main/pkg/gds/client/profiles.go#L39-L42) so we can import them and use them in the tests? Or we could create a helper function `func GetProfileMeta() (yaml, version string)` that returns them without being able to modify them if you'd prefer.

pkg/gds/client/profiles_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+// Test that DefaultProfiles returns a Profiles object with some valid profiles.
+func TestDefaultProfiles(t *testing.T) {
+	profiles := client.DefaultProfiles()
+	require.NotNil(t, profiles)
+	require.True(t, strings.HasPrefix(profiles.Version, "v"))
```
bbengfort, line 23: Per the above constant if we capitalize the `profileVersion` constant we can check that it is as expected, which is probably a good test to make sure we don't mess up when changing versions.

pkg/gds/client/profiles_test.go
```diff
@@ -0,0 +1,196 @@
...
+	require.NotEmpty(t, profiles.Profiles)
+	require.Contains(t, profiles.Profiles, profiles.Active)
```
bbengfort, line 25: Good generic tests that will make it easy to maintain the `DefaultProfiles` function!

pkg/gds/client/profiles_test.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// Test that GetProfilesFolder finds the correct folder.
+func TestGetProfilesFolder(t *testing.T) {
+	// Should default to the local path if it contains a profiles config
```
bbengfort, line 30: So if I have a config file in my user dir and I run the tests locally it shouldn't find my profiles, it will just use the test profiles?

pkg/gds/client/profiles_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+	// Profiles should be written to the specified config.
+	tmp, err := ioutil.TempDir("", "config-*")
+	defer os.RemoveAll(tmp)
+	require.NoError(t, err)
+	folder := &configdir.Config{
+		Path: tmp,
+		Type: configdir.Local,
+	}
+	err = profiles.Save(folder)
```
bbengfort, line 171: It was a good call to have the folder passed to save so that this test would only write to the tempdir.

bbengfort (commented):
This is a tricky package to test since a lot of the code is dependent on `shibukawa/configdir` package which doesn't give a lot of methods for testing. Thank you for tackling it so thoughtfully! 

The tests look very complete to me and well done; I'd like to checkout the local modification of configuration files to see if that affects things and there are some minor comments, but after that we should be good to go.

bbengfort (approved):
:shipit:

PR #208, opened 2021-11-16
GDS Containers Update

bbengfort:
This PR adds the TRTL Dockerfile and updates the GDS Dockerfile to be a
bit lighter-weight and share cache with the TRTL Dockerfile in order to
reduce build times. The GDS container has decreased from 175MB to 132MB
as a result of these changes.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+.env
+.git
+.github
+.vscode
+fixtures
+.DS_Store
+LICENSE
+README.md
--- a/containers/build.sh
+++ b/containers/build.sh
@@ -76,6 +76,7 @@ REPO=$(realpath "$DIR/..")
 
 # Build the primary backend images
 docker build -t trisa/gds:$TAG -f $DIR/gds/Dockerfile $REPO
+docker build -t trisa/trtl:$TAG -f $DIR/trtl/Dockerfile $REPO
 docker build -t trisa/grpc-proxy:$TAG -f $DIR/grpc-proxy/Dockerfile $REPO
 
 # Build the UI images for trisatest.net and vaspdirectory.net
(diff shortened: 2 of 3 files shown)


.dockerignore
```diff
@@ -0,0 +1,8 @@
+.env
+.git
```
rebeccabilbro, line 2: I wonder if this would impact a future `rumi` integration?

containers/gds/Dockerfile
```diff
@@ -1,22 +1,51 @@
...
-WORKDIR /srv/build
+# Build stage
+FROM ${BUILDER_IMAGE} as builder
 
-COPY . .
-RUN go test ./... && go build -v ./cmd/gds
-RUN go build -v ./cmd/sectigo
-RUN go build -v ./cmd/gdsutil
+# Ensure ca-certificates are up to date
+RUN update-ca-certificates
```
rebeccabilbro, line 9: Was this not something we were doing before? Could you clarify how this differs from the steps on lines 41-44?

containers/gds/Dockerfile
```diff
@@ -1,22 +1,51 @@
...
 
-FROM ubuntu:bionic
+WORKDIR $GOPATH/src/github.com/trisacrypto/directory
+
+# Use modules for dependencies
+COPY go.mod .
+COPY go.sum .
+
+ENV CGO_ENABLED=0
+ENV GO111MODULE=on
```
rebeccabilbro, line 18: Would you please clarify the usage of these new lines?

rebeccabilbro (approved):
Thank you for making the docker containers more compact in this PR @bbengfort -- I have noted a few questions that I'd appreciate for understanding purposes, but I don't want to block #209

.dockerignore
bbengfort: I see what you mean, if the `.git` directory is omitted, rumi will not have access to the history. However, I very much doubt that rumi will be run in the context of these docker containers which are intended as a production build for the deployment environment. Rumi will likely run in the context of a local directory (e.g. not in the container) or in a GitHub action, which will clone the entire repo including the .git history.

containers/gds/Dockerfile
bbengfort: We were not doing this before; though it's unlikely that it would cause a problem, it could future-proof dependency downloads in the future. I figured it wouldn't hurt to help us maintain these Dockerfiles in the long run, particularly if we move to distroless or static (the blog post that I posted in the #engineering channel a month or two ago). 

This command is being run in the build stage, and ensures that the certs are up to date for any downloads happening in this stage such as dependency downloads.  Lines 41-44 are in the final stage (which is a different container) - those lines are more about ensuring that TLS certs (such as mTLS certs) can be correctly verified.

containers/gds/Dockerfile
bbengfort: `CGO_ENABLED=0` turns CGO compilation off, which means that Go will generate a statically compiled binary with no external dependencies (runtime linking to C libraries). This is a slightly slower compilation process and requires that we have only pure Go packages and dependencies that don't use C libraries; however, the binary can then be run completely independently in a very lightweight environment without having to install any other libraries (and reduces the overall size of the container). 

`GO111MODULE=on` forces Go to use `go module` even if the source code is in the `$GOPATH` (which it is, see line 11). This is another future proofing thing to make it easier to upgrade to Go 1.18 which will require go modules.

PR #210, opened 2021-11-16
Extends fakerize script to build synthetic certificate requests

rebeccabilbro:
This PR addresses sc-1639, extending the `fakerize.py` script to generate synthetic certificate requests as well as vasps. I have combined the resultant data with the previous data asset and updated `fakes.tgz` inside `pkg/gds/testdata`, and the corresponding loader test in `pkg/gds/gds_test.go`.


fixtures/datagen/fakerize.py
```diff
@@ -419,28 +521,182 @@ def augment_vasps(fake_names=FAKE_VASPS):
...
+    Return a list of dictionaries representing a synthetic but plausible audit log
+    """
+    logs = []
+
+    states = [state]
+    current_state = state
+    prior_state = None
+    while current_state in CERT_STATE_CHANGES:
+        prior_state = CERT_STATE_CHANGES[current_state]["previous_state"]
+        if prior_state == "INITIALIZED":
```
pdeziel, line 655: Maybe I'm not following the logic correctly but doesn't this stop before the `INITIALIZED` state? Since the previous state for `READY_TO_SUBMIT` is also `INITIALIZED`.

pkg/gds/gds_test.go
```diff
@@ -51,14 +53,19 @@ func loadFixtures(s *gdsTestSuite) {
...
 			return nil
 		}
-
+		if strings.HasPrefix(info.Name(), "certs::") {
+			cert := &models.CertificateRequest{}
+			err = protojson.Unmarshal(data, cert)
+			require.NoError(err)
+			dbCerts[info.Name()] = cert
+			return nil
+		}
```
pdeziel, line 68: This was pretty much how I was originally planning on adding the certs, although following discussion in #203 I think I'm going to end up making the dbVASPs/dbCerts into a generic map anyway, so you don't have to include these changes here if you don't want to (although it does help verify for the time being that the CertReq fixtures are valid).

pdeziel (approved):
Thanks for figuring out some of the details with the CertReq fixtures, I think this looks good pending the comments below.

rebeccabilbro:
Reminder to self to force commit the new template

fixtures/datagen/fakerize.py
rebeccabilbro: Good catch; fixed!

PR #211, opened 2021-11-17
Update TRTL protocol buffers

bbengfort:
And regenerate to build versions of protoc-gen-go and protoc-gen-go-grpc

Note that the package is still `github.com/trisacrypto/directory/pkg/trtl/pb/v1` -- this only affects the protocol buffers. The package structure will allow us to generate code into `github.com/trisacrypto/directory/pkg/trtl/pb/v2` etc while keeping the primary `trtl` package organized.


go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/gin-contrib/cors v1.3.1
 	github.com/gin-gonic/gin v1.7.4
 	github.com/golang-jwt/jwt/v4 v4.0.0
+	github.com/golang/protobuf v1.5.2
```
rebeccabilbro, line 12: not sure how we managed not to have this already?

rebeccabilbro (approved):
lgtm

go.mod
bbengfort: me either; that is a bit weird; before our next deploy we may want to run `go generate ./…` to make sure everything is up to date.

PR #212, opened 2021-11-17
TRTL: Implement Iter and Cursor RPCs

bbengfort:
Implementation of the TRTL:

- `Iter` RPC for paginated materialized list requests
-  `Cursor` RPC for server-side streaming list requests.

This is a bit bigger of a project than I expected, so I'm trying to break it down into multiple commits that can be reviewed independently. I've also commented inline in several places to help the review process. 

### Pagination

The `Iter` RPC requires pagination since the result set can be arbitrarily long and this might result in too large a response. I've implemented the "list pagination" pattern from [Google's gRPC common design patterns](https://cloud.google.com/apis/design/design_patterns#list_pagination) document. This requires adding a `page_token` and `page_size` to the request along with a `next_page_token` to the response. The tokens are supposed to be URL-safe base64 encoded protocol buffers, which I implemented as `PageCursor` in `protos/trtl/internal/pagination.go` (see internal notes below). 

When implementing a page token, the token must be verified to prevent leakage between requests. I've implemented two checks:

1. Page size must match the request page size
2. The prefix must not change between requests

The algorithm for pagination is fairly complex and requires Seeking - see inline notes about this. 

### Internal

Internal is supposed to indicate that the protobufs in that package are not to be used by anything outside of `trtl` and can be modified without versioning since they'll be versioned with the package. However, the Go linter apparently respects `internal` packages, when I originally tried to generate the code in `pkg/trtl/pb/internal` and use the code from `pkg/trtl` I got a complaint:

> use of internal package github.com/trisacrypto/directory/pkg/trtl/pb/internal not allowed (compile)

According to this [stackoverflow post](https://stackoverflow.com/questions/59342373/use-of-internal-package-not-allowed) internal package use is allowed in the same package or subpackages, but not any higher level packages, therefore the fix was to generate the code in `pkg/trtl/internal` and now the compiler is happy.  

### Synchronization Issues with `Iter`

`Iter` pagination unfortunately introduces a synchronization problem: Let's say there is are paginated requests that arrive at the server in the following order:

```
Iter() ⟶ page a, next=b
Iter(page_token='b') ⟶ page b, next = c
Iter(page_token='c') ⟶ page c, next = d
Iter(page_token='d') ⟶ page d,
```

If there are other requests, between Iter requests, this could cause problems because there is no snapshot:

```
Iter() ⟶ page a, next=b
Put(c)
Iter(page_token='b') ⟶ page b, next = c
Delete(c)
Iter(page_token='c') ⟶ page d
```

From the perspective of the user iterating, the Put(c) caused the iteration to increase, and could lead to the situation where iteration never stops, if a put keeps increasing the number of pages. 

The delete between Iter b and Iter c can also cause iteration to fail if done incorrectly, or artificially shorten the requests, e.g. "I said there was a next page, but actually there is not". 

The solution is the `Cursor` RPC but it's also something we'll have to consider in Honu. 

### Tests

I moved the Honu tests from `trtl_test.go` to `honu_test.go` just for organizational purposes. I also implemented `grpc bufconn` so that we weren't testing across a live network.


go.mod
```diff
@@ -9,13 +9,12 @@ require (
 	github.com/gin-contrib/cors v1.3.1
 	github.com/gin-gonic/gin v1.7.4
 	github.com/golang-jwt/jwt/v4 v4.0.0
-	github.com/golang/protobuf v1.5.2
```
bbengfort, line 12: @rebeccabilbro and now it's gone … I'll make sure I've rebased correctly, but I'm pretty sure I have.

go.mod
```diff
@@ -9,13 +9,12 @@ require (
...
 	github.com/google/go-querystring v1.1.0
 	github.com/google/uuid v1.2.0
 	github.com/googleapis/gax-go v1.0.3
 	github.com/joho/godotenv v1.3.0
 	github.com/kelseyhightower/envconfig v1.4.0
-	github.com/rotationalio/honu v0.2.0
+	github.com/rotationalio/honu v0.2.1
```
bbengfort, line 17: I needed the latest and greatest Honu so I tagged and updated.

pkg/trtl/honu.go
```diff
@@ -108,8 +112,179 @@ func (h *HonuService) Delete(ctx context.Context, in *pb.DeleteRequest) (out *pb
...
+		opts = in.Options
+	} else {
+		// Create empty options
+		opts = &pb.Options{}
+	}
+
+	// Load defaults into options
+	if opts.PageSize == 0 {
+		opts.PageSize = 100
+	}
```
bbengfort, line 151: We should probably create a helper function that instantiates default options then updates the user options, I think this would be useful in all of our RPC functions. Let me know what you guys think!

pkg/trtl/honu.go
```diff
@@ -108,8 +112,179 @@ func (h *HonuService) Delete(ctx context.Context, in *pb.DeleteRequest) (out *pb
...
+
+	// Compute the actual starting prefix as the namespace plus the key
+	var prefix []byte
+	if in.Namespace != "" {
+		prefix = prepend(in.Namespace, in.Prefix)
+	} else {
+		prefix = prepend("default", in.Prefix)
+	}
```
bbengfort, line 159: This is also repeated code; perhaps another helper function? Possibly we should also make `"default"` a constant?

pkg/trtl/honu.go
```diff
@@ -108,8 +112,179 @@ func (h *HonuService) Delete(ctx context.Context, in *pb.DeleteRequest) (out *pb
...
+			// to continue seeking. Note that we cannot use equality here because the
+			// next key may have been deleted between requests, which means we'd seek to
+			// the end of the iteration without returning the page. Unfortunately, the
+			// lexicographic ordering that we're computing is heavily dependent on the
+			// underlying representation, so I'm just guessing with leveldb for now.
+			// TODO: this needs to be replaced with honu Seek!
+			if comparer.DefaultComparer.Compare(key, cursor.NextKey) < 0 {
+				continue
+			}
+		}
```
bbengfort, line 217: Note the TODOs and the comments; probably this is a story that will require a modification to both honu and trtl. We should also look at the synchronization and configuration issues I mentioned exist in Honu as well. Alternatively we could simply remove Honu as a dependency and shift some of that code to trtl if we wanted to leave Honu in a state to evolve by itself. Some discussion required.

pkg/trtl/honu.go
```diff
@@ -108,8 +112,179 @@ func (h *HonuService) Delete(ctx context.Context, in *pb.DeleteRequest) (out *pb
...
+			break
+		}
+
+		// Otherwise append the current key value pair to the page.
+		// Fetch the metadata since it will need to be loaded for the response anyway.
+		var object *object.Object
+		if object, err = iter.Object(); err != nil {
+			log.Error().Err(err).Str("key", base64.RawURLEncoding.EncodeToString(key)).Msg("could not fetch object metadata")
+			return nil, status.Error(codes.FailedPrecondition, "database is in invalid state")
+		}
```
bbengfort, line 232: This is the critical sequence of the iteration.

pkg/trtl/honu.go
```diff
@@ -108,8 +112,179 @@ func (h *HonuService) Delete(ctx context.Context, in *pb.DeleteRequest) (out *pb
...
+					Pid:     object.Version.Pid,
+					Version: object.Version.Version,
+					Region:  object.Version.Region,
+				},
+				Parent: &pb.Version{
+					Pid:     object.Version.Parent.Pid,
+					Version: object.Version.Parent.Version,
+					Region:  object.Version.Parent.Region,
+				},
+			}
```
bbengfort, line 262: This is also going to be very common code so also a good candidate for a helper function.

proto/trtl/v1/trtl.proto
```diff
@@ -127,8 +128,11 @@ message SyncReply {
 // Options conditions all accesses to trtl, e.g. there are not different structs for
 // Get vs Put options. The semantics of each option depends on the type of request.
 message Options {
-    bool return_meta = 1;   // generally, return the version information for the object in the response
-    bool iter_no_keys = 2;  // do not include keys in an Iter or Batch response, to reduce data transfer load
+    bool return_meta = 1;     // generally, return the version information for the object in the response
+    bool iter_no_keys = 2;    // do not include keys in an Iter or Batch response, to reduce data transfer load
+    bool iter_no_values = 3;  // do not include values in an Iter or Batch response, to reduce data transfer load
+    string page_token = 4;    // specify the page token to fetch the next page of results
+    int32 page_size = 5;      // specify the number of results per page, cannot change between page requests
```
bbengfort, line 135: I'm not 100% sure about if these should be options or if they should be part of the request. The `return_meta` option is pretty common, but the `iter` options are only useful for `Iter`, `Batch` and `Cursor` (I guess maybe `Sync` too). I'd be happy to discuss this.

pkg/trtl/honu_test.go
```diff
@@ -0,0 +1,166 @@
...
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/trisacrypto/directory/pkg/trtl/pb/v1"
+	"google.golang.org/protobuf/proto"
+)
+
+// Test that we can call the Get RPC and get the correct response.
+func (s *trtlTestSuite) TestGet() {
```
bbengfort, line 12: Moved for organizational purposes.

pkg/trtl/honu_test.go
```diff
@@ -0,0 +1,166 @@
...
+	require := s.Require()
+	alice := dbFixtures["alice"]
+	object := dbFixtures["object"]
+	ctx := context.Background()
+
+	// Start the gRPC client.
+	conn, err := s.connect()
+	require.NoError(err)
+	defer conn.Close()
+	client := pb.NewTrtlClient(conn)
```
bbengfort, line 24: No longer relies on a live gRPC server.

pkg/trtl/honu_test.go
```diff
@@ -0,0 +1,166 @@
...
+	})
+	require.NoError(err)
+	require.NotNil(reply.Meta)
+	require.Equal([]byte(alice.Key), reply.Meta.Key)
+	require.Equal(alice.Namespace, reply.Meta.Namespace)
+	require.True(proto.Equal(expectedMeta, reply.Meta))
+}
+
+// Test that we can call the Batch RPC and get the correct response.
+func (s *trtlTestSuite) TestBatch() {
```
bbengfort, line 118: Moved for organizational purposes.

pkg/trtl/honu_test.go
```diff
@@ -0,0 +1,166 @@
...
+	require := s.Require()
+	ctx := context.Background()
+
+	// Start the gRPC client.
+	conn, err := s.connect()
+	require.NoError(err)
+	defer conn.Close()
+	client := pb.NewTrtlClient(conn)
```
bbengfort, line 126: No longer relies on a live gRPC server.

pkg/trtl/trtl_test.go
```diff
@@ -102,6 +109,15 @@ func generateDB(s *trtlTestSuite) {
...
 }
 
+func (s *trtlTestSuite) connect() (*grpc.ClientConn, error) {
+	ctx := context.Background()
+	return grpc.DialContext(ctx, "bufnet", grpc.WithContextDialer(s.dialer), grpc.WithInsecure())
+}
+
+func (s *trtlTestSuite) dialer(context.Context, string) (net.Conn, error) {
+	return s.conn.Dial()
+}
```
bbengfort, line 119: These functions allow tests to create a client that connects via the bufconn. The general pattern for tests is now implemented in `TestGet` and `TestBatch`.

pkg/trtl/trtl_test.go
```diff
@@ -132,6 +148,17 @@ func (s *trtlTestSuite) SetupSuite() {
...
+	// Create the trtl server
+	s.trtl, err = trtl.New(s.conf)
+	require.NoError(err)
+
+	// Create a bufcon listener so that there are no actual network requests
+	s.conn = bufconn.Listen(bufSize)
+
+	// Run the test server without signals, background routines or maintenance mode checks
+	// TODO: do we need to check if there was an error when starting run?
+	go s.trtl.Run(s.conn)
```
bbengfort, line 161: This refactoring allows us to run the trtl server without dialing a real network address and taking a port during tests. The buffconn is also faster to clean up and allows in-memory tests.

pkg/trtl/trtl.go
```diff
@@ -134,6 +129,16 @@ func (t *Server) Serve() (err error) {
...
 
+// Run the gRPC server. This method is extracted from the Serve function so that it can
+// be run in its own go routine and to allow tests to Run a bufconn server without
+// starting a live server with all of the various go routines and channels running.
+func (t *Server) Run(sock net.Listener) {
+	defer sock.Close()
+	if err := t.srv.Serve(sock); err != nil {
+		t.echan <- err
+	}
+}
```
bbengfort, line 140: This change made to ensure that tests can run using bufconn.

pkg/trtl/honu_test.go
```diff
@@ -0,0 +1,330 @@
...
+	require.NoError(err, "could not fetch complete iteration")
+	require.Empty(rep.NextPageToken, "a next page token was returned for a one page response")
+	require.Len(rep.Values, 5, "incorrect responses returned did the fixtures change?")
+
+	expectedOrder := []string{
+		"alice", "bob", "charlie", "darelene", "erica",
+		"franklin", "gregor", "helen", "ivan", "juliet",
+	}
+
+	for i := 0; i < 5; i++ {
```
bbengfort, line 220: Expected order is failing, which is also causing pagination to fail … not sure what's going on; could be something in Honu or the way the fixtures are inserted. More investigation required.

proto/trtl/v1/trtl.proto
bbengfort: Note, iter options does not include `Batch` only `Iter` and `Cursor`.

rebeccabilbro (approved):
Thanks for the walkthrough @bbengfort! Summarizing feedback we already discussed below:

- we'll remove `next` from `CursorReply`; requires some refactoring of this PR but will simplify things considerably
- we want to revisit `Options`, but we don't know exactly how yet; something to come back to
- at some point we'll revisit the idea of memory management, so that we can have more control over long-running iteration
- CLI functionality not needed for now, as we can test with the test suite; we'll revisit when we actually have user stories around the CLI specifically
- We'll merge this in ahead of #213, and I'll rebase and resolve the conflicts in my branch, push up the changes to my PR so that you can review. This PR should also block sc-1620 (Delete) given the refactoring of the test suite, so I should wait to start on that work until this is merged.

PR #213, opened 2021-11-17
Implements trtl Put

rebeccabilbro:
This PR implements trtl's `Put` method to close sc-1619, also adding preliminary test coverage and an initial CLI function. 

Note that the CLI function is not tested (I think my env isn't set up to test the CLI, so I also can't use `db:get`).  I'm not really sure how we'd plan to use the CLI here (put a single value to a single key? Are keys and values just strings or are we expecting more complicated input?) No specific requirements were described in the story, so I've just made my best guesses.


cmd/trtl/main.go
```diff
@@ -118,6 +118,47 @@ func main() {
...
+					Name:    "replica-endpoint",
+					Aliases: []string{"u"},
+					Usage:   "the url to connect to the trtl replication service",
+					EnvVars: []string{"TRISA_DIRECTORY_REPLICA_URL"},
+				},
+				&cli.StringFlag{
+					Name:    "key",
+					Aliases: []string{"k"},
+					Usage:   "specify the key as a string",
+				},
```
rebeccabilbro, line 139: I went back and forth on what to call this, since `key` seemed more intuitive, but it would mean a different interface to `Get`

pkg/trtl/honu.go
```diff
@@ -101,7 +101,69 @@ func (h *HonuService) Get(ctx context.Context, in *pb.GetRequest) (*pb.GetReply,
...
+	var success bool
+	if err := h.db.Put(key, in.Value); err != nil {
+		success = false
+	} else {
+		success = true
+	}
+
+	out = &pb.PutReply{Success: success}
+
+	// if Options include a request for metadata, we need to do a Get
```
rebeccabilbro, line 139: Not really sure if this is what we imagined when we said we would enable Options for `Put`. Let me know if I guessed incorrectly!

rebeccabilbro, line 139: Not really sure if this is what we imagined when we said we would enable Options for `Put`. Let me know if I guessed incorrectly!

pkg/trtl/trtl_test.go
```diff
@@ -306,9 +406,6 @@ func (s *trtlTestSuite) TestBatch() {
 	require.Equal(int64(len(requests)), reply.Failed)
 	require.Equal(int64(0), reply.Successful)
 	require.Len(reply.Errors, len(requests))
-	for _, e := range reply.Errors {
-		require.Contains(requests, e.Id)
-		require.Equal(requests[e.Id].Id, e.Id)
-		require.Contains(e.Error, "not implemented")
-	}
+	require.Contains(requests, reply.Errors[1].Id)
+	require.Equal(requests[reply.Errors[1].Id].Id, reply.Errors[1].Id)
```
rebeccabilbro, line 410: FYI - had to update this test now that Put is implemented, but I can do a better update when I implement Delete in my next PR

pkg/trtl/honu_test.go
```diff
@@ -160,11 +255,8 @@ func (s *trtlTestSuite) TestBatch() {
 	require.Equal(int64(len(requests)), reply.Failed)
 	require.Equal(int64(0), reply.Successful)
 	require.Len(reply.Errors, len(requests))
-	for _, e := range reply.Errors {
-		require.Contains(requests, e.Id)
-		require.Equal(requests[e.Id].Id, e.Id)
-		require.Contains(e.Error, "not implemented")
-	}
+	require.Contains(requests, reply.Errors[1].Id)
+	require.Equal(requests[reply.Errors[1].Id].Id, reply.Errors[1].Id)
```
rebeccabilbro, line 259: FYI - had to update this test now that Put is implemented, but I can do a better update when I implement Delete in my next PR

pkg/trtl/honu_test.go
```diff
@@ -116,6 +118,99 @@ func (s *trtlTestSuite) TestGet() {
...
+		},
+	})
+	require.NoError(err)
+	require.True(reply.Success)
+	require.Empty((reply.Meta))
+
+	// Put a value with return_meta=true.
+	expectedPID, err := strconv.Atoi((os.Getenv("TRTL_REPLICA_PID")))
+	require.NoError(err)
+	expectedRegion := os.Getenv("TRTL_REPLICA_REGION")
```
bbengfort, line 181: Good solution to get the values that weren't hard coded, but I think this is a problem that I created -- sorry about that! I changed these lines: https://github.com/trisacrypto/directory/blob/main/pkg/trtl/trtl_test.go#L159-L162 not realizing that they were being used for the test. I will create a bug story to fix this by adding PID and region to the [test variables](https://github.com/trisacrypto/directory/blob/main/pkg/trtl/trtl_test.go#L28-L31) and making sure they're used in the [fixture](https://github.com/trisacrypto/directory/blob/main/pkg/trtl/trtl_test.go#L43-L53) and in the configuration. My bad!

bbengfort (approved):
Looks great! I had a couple of suggestions that hopefully can be added directly from the web UI - but other than that everything is discussion! I think once we merge, there are a couple of stories we need to open. Thank you so much for taking on the hard work of dealing with the merge conflicts from the `Iter` and `Cursor` PR!

pkg/trtl/honu_test.go
```diff
@@ -116,6 +118,99 @@ func (s *trtlTestSuite) TestGet() {
...
+		Parent: &pb.Version{
+			Pid:     uint64(expectedPID),
+			Version: 3,
+			Region:  expectedRegion,
+		},
+	}
+	reply, err = client.Put(ctx, &pb.PutRequest{
+		Namespace: alice.Namespace,
+		Key:       []byte(alice.Key),
+		Value:     []byte("cheshire cat"),
```
bbengfort, line 201: I just had one random thought - this request should be changing the state of the underlying database from the JSON in the fixture to this value here, which means a subsequent test that tests for the value might error. The tests are fine now, but I wanted to raise the point in case we encounter it in future, to help us find the problem more quickly.

PR #216, opened 2021-11-18
sc-2059 allow multiple status filters

pdeziel:
This allows multiple status filters to be passed into the `ListVASPs` endpoint in the GET params. The GET requests end up looking like this:

`/v2/vasps?status_filters=pending_review&status_filters=verified`

I also included the pagination bug fix here, since I was in the area anyway.

### DIFF
--- a/cmd/gds/main.go
+++ b/cmd/gds/main.go
@@ -364,8 +364,8 @@ func main() {
 						Aliases: []string{"s"},
 						Usage:   "specify the number of items per page",
 					},
-					&cli.StringFlag{
-						Name:    "status",
+					&cli.StringSliceFlag{
+						Name:    "status-filters",
 						Aliases: []string{"S"},
 						Usage:   "filter by verification status",
 					},
--- a/cmd/gds/main.go
+++ b/cmd/gds/main.go
@@ -874,9 +874,9 @@ func adminListVASPs(c *cli.Context) (err error) {
 	defer cancel()
 
 	params := &admin.ListVASPsParams{
-		Page:     c.Int("page"),
-		PageSize: c.Int("page-size"),
-		Status:   c.String("status"),
+		Page:          c.Int("page"),
+		PageSize:      c.Int("page-size"),
+		StatusFilters: c.StringSlice("status-filters"),
 	}
 
 	var rep *admin.ListVASPsReply
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -570,16 +570,18 @@ func (s *Admin) ListVASPs(c *gin.Context) {
 	}
 
 	// Determine status filter
-	var status pb.VerificationState
-	if in.Status != "" {
-		in.Status = strings.ToUpper(strings.ReplaceAll(in.Status, " ", "_"))
-		sn, ok := pb.VerificationState_value[in.Status]
-		if !ok {
-			log.Warn().Str("status", in.Status).Msg("unknown verification status")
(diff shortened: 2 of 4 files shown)


cmd/gds/main.go
```diff
@@ -874,9 +874,9 @@ func adminListVASPs(c *cli.Context) (err error) {
 	defer cancel()
 
 	params := &admin.ListVASPsParams{
-		Page:     c.Int("page"),
-		PageSize: c.Int("page-size"),
-		Status:   c.String("status"),
+		Page:          c.Int("page"),
+		PageSize:      c.Int("page-size"),
+		StatusFilters: c.StringSlice("status-filters"),
```
bbengfort, line 879: Nice! Don't forget you also have to change the flag definition on [line 367](https://github.com/trisacrypto/directory/blob/main/cmd/gds/main.go#L367)

pkg/gds/admin.go
```diff
@@ -604,8 +606,7 @@ func (s *Admin) ListVASPs(c *gin.Context) {
 	// Query the list of VASPs from the data store
 	iter := s.db.ListVASPs()
 	defer iter.Release()
-	for iter.Next() {
-		out.Count++
+	for out.Count = 0; iter.Next(); out.Count++ {
```
bbengfort, line 609: :-)

pkg/gds/admin.go
```diff
@@ -615,8 +616,8 @@ func (s *Admin) ListVASPs(c *gin.Context) {
 				continue
 			}
 
-			// Check the status before continuing
-			if status != pb.VerificationState_NO_VERIFICATION && vasp.VerificationStatus != status {
+			// Check against the status filters before continuing
+			if _, ok := filters[vasp.VerificationStatus]; len(filters) > 0 && !ok {
```
bbengfort, line 620: Nice and compact test!

bbengfort (approved):
@pdeziel thank you for putting this together so fast; only one quick change that needs to be made.

@elysee15 once we get this merged it should unblock you.

pkg/gds/admin/v2/api.go
```diff
@@ -87,9 +87,9 @@ type AutocompleteReply struct {
 // ListVASPsParams is a request-like struct that passes query params to the ListVASPs
 // GET request. All query params are optional and modify how and what data is retrieved.
 type ListVASPsParams struct {
-	Status   string `url:"status,omitempty" form:"status"`
-	Page     int    `url:"page,omitempty" form:"page" default:"1"`             // defaults to page 1 if not included
-	PageSize int    `url:"page_size,omitempty" form:"page_size" default:"100"` // defaults to 100 if not included
+	StatusFilters []string `url:"status_filters,omitempty" form:"status_filters"`
```
bbengfort, line 90: I'm going to rename the url back to `status` because this makes more sense as a query param, e.g. the query is:

`/v2/vasps?status=foo&status=bar`

pkg/gds/admin.go
```diff
@@ -604,19 +606,20 @@ func (s *Admin) ListVASPs(c *gin.Context) {
...
 				// VASP could not be parsed; error logged in VASP() method continue iteration
+				out.Count--
 				continue
 			}
 
-			// Check the status before continuing
-			if status != pb.VerificationState_NO_VERIFICATION && vasp.VerificationStatus != status {
+			// Check against the status filters before continuing
+			if _, ok := filters[vasp.VerificationStatus]; len(filters) > 0 && !ok {
+				out.Count--
```
bbengfort, line 622: Count is definitely still having trouble … decrementing here if we're filtering so the count reflects the number we're expecting back. Will need to write tests for this.

PR #217, opened 2021-11-18
sc-2045 create aliases for imports

elysee15:


bbengfort:
@elysee15 it looks like this branch needs to be rebased with main to update it with respect to the new contact handling.

bbengfort (approved):
Very handy, and very cool - thank you for making this change!

PR #219, opened 2021-11-19
fix: Re-render VASP detail is issuing many API requests

elysee15:
Fixes SC-2080


bbengfort (approved):
This makes sense - good fix!

PR #220, opened 2021-11-19
fix: Filter VASP list via query param and UI controls

elysee15:
Fixes SC-1296
Screen Shot 2021-11-19 at 22 10 58


tl6kk (approved):
Verified:

- select multiple statuses
- export to .csv
- link modified with status selections
- status names

PR #221, opened 2021-11-20
Implements trtl delete

rebeccabilbro:
This PR implements trtl's Delete method, with some preliminary tests and an (untested) prototype CLI command. 

I've opened several follow on issues to deal with the honu issue we uncovered whilst working on this PR (doesn't currently handle namespaces) and the test fixtures (bypassing honu loading results in unexpected metadata).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/gin-contrib/cors v1.3.1
 	github.com/gin-gonic/gin v1.7.4
 	github.com/golang-jwt/jwt/v4 v4.0.0
+	github.com/google/go-cmp v0.5.5 // indirect
 	github.com/google/go-querystring v1.1.0
 	github.com/google/uuid v1.2.0
 	github.com/googleapis/gax-go v1.0.3
--- a/pkg/trtl/honu_test.go
+++ b/pkg/trtl/honu_test.go
@@ -213,6 +213,95 @@ func (s *trtlTestSuite) TestPut() {
 	require.True(proto.Equal(expectedMeta, reply.Meta))
 }
 
+// Test that we can call the Delete RPC and get the correct response.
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/gin-contrib/cors v1.3.1
 	github.com/gin-gonic/gin v1.7.4
 	github.com/golang-jwt/jwt/v4 v4.0.0
+	github.com/google/go-cmp v0.5.5 // indirect
```
rebeccabilbro, line 12: using to compare protocol buffers

pkg/trtl/honu_test.go
```diff
@@ -213,6 +213,95 @@ func (s *trtlTestSuite) TestPut() {
...
+	// 		Version: 1,
+	// 		Region:  expectedRegion,
+	// 	},
+	// }
+	require.NoError(err)
+	require.True(withMeta.Success)
+	require.NotNil(withMeta.Meta)
+	// TODO uncomment when sc-2102 and sc-2103 are fixed
+	// require.Equal(tempKey, withMeta.Meta.Key)
+	// require.Equal(tempNS, withMeta.Meta.Namespace)
```
rebeccabilbro, line 301: Sorry for the ugly block of commented out code. I've opened up a few follow on tasks to address the problems that prevent these tests from passing as expected; hopefully we will be able to repair them soon

bbengfort: Thank you so much for catching the issue with the prototype Honu code!

bbengfort (approved):
Everything looks great - thank you so much for finding the issue with Honu Put and the fixture  - I know it was supposed to be a small PR but it turned out to be a very important one!

PR #223, opened 2021-11-22
GitHub Actions for Container Build and Push

bbengfort:
This PR adds container builds to GitHub Actions. There is a single job per container build meaning that we've got 7 additional jobs. All of the jobs are fairly similar with slight configuration differences but it is nice to have all the builds run in parallel on GitHub actions - it speeds things up a lot! 

On pull request, the containers will be built but not pushed to GCR and DockerHub. When tagging or merging with main the images will be pushed to both DockerHub and gcr.io. 

Hold until #222 is merged.


.github/workflows/containers.yaml
```diff
@@ -0,0 +1,416 @@
...
+    tags:
+      - 'v*'
+  pull_request:
+    branches:
+      - main
+
+# Each container image is built in its own job. To add an image, simply copy and paste
+# one of the jobs and configure the build to point to the container's Dockerfile and
+# tags. If the build is main or for a tag then the images are pushed to dockerhub and
+# gcr; images are not pushed on pull requests.
```
rebeccabilbro, line 15: Thank you for adding this annotation to help us with maintenance!

containers/gds/Dockerfile
```diff
@@ -25,7 +25,7 @@ COPY pkg ./pkg
 COPY proto ./proto
 
 # Run tests to avoid building a broken container
-RUN go test -timeout 30s -v ./...
+RUN go test -timeout 180s ./...
```
rebeccabilbro, line 28: Assuming we bumped this up because we were observing premature timeouts?

rebeccabilbro (approved):
This looks really good @bbengfort - I'm excited to see all the components of the GDS build so nicely coordinated here! Thank you especially for adding in notes to the `containers.yaml` so that we have a template for adding new containers/components, and for maintaining those we've already integrated. 

Per your note, we'll hold on merging this in until #222 is resolved.

containers/gds/Dockerfile
bbengfort: I only saw it happen once, and I don't think it was related to our tests, it was related to the Actions environment; however our tests will only become more comprehensive so I figured it wouldn't hurt to bump the timeout.

bbengfort:
@rebeccabilbro @pdeziel  ok - I'm going to merge; cross your fingers that the image push works when we merge to main!

PR #224, opened 2021-11-22
Update GDS UI README with plural guidelines 

tl6kk:


rebeccabilbro (approved):
Thank you so much @tl6kk for adding in this bit of documentation around pluralization for Lingui; it definitely clarifies the mystery of the error messages we were getting earlier when we tried to build the multilingual version of the GDS UI, and it will help us as we progress in our i18n and l10n efforts!

PR #225, opened 2021-11-22
sc-2186 GDS mock

pdeziel:
This adds a mock for GDS to support testing. In order to have a framework for testing the RPCs, I've added a bufconn listener which the tests can invoke to mock live server behavior. This ended up being very similar to what we're doing with the trtl testing, so I abstracted the bufconn support code into its own utils package.


pkg/gds/gds.go
```diff
@@ -71,23 +71,28 @@ func (s *GDS) Serve() (err error) {
...
+	log.Info().
+		Str("listen", s.conf.BindAddr).
+		Str("version", pkg.Version()).
+		Msg("trisa directory server started")
+
+	if err := s.srv.Serve(sock); err != nil {
+		s.svc.echan <- err
+	}
+}
+
```
pdeziel, line 95: This makes me nervous but hopefully it should be obvious if it breaks anything?

pkg/gds/gds.go
```diff
@@ -71,23 +71,28 @@ func (s *GDS) Serve() (err error) {
...
+	defer sock.Close()
+	log.Info().
+		Str("listen", s.conf.BindAddr).
+		Str("version", pkg.Version()).
+		Msg("trisa directory server started")
+
+	if err := s.srv.Serve(sock); err != nil {
+		s.svc.echan <- err
+	}
+}
```
bbengfort, line 94:
```suggestion
	defer sock.Close()
	if err := s.srv.Serve(sock); err != nil {
		s.svc.echan <- err
	}
}
```

pkg/gds/gds.go
```diff
@@ -71,23 +71,28 @@ func (s *GDS) Serve() (err error) {
...
-		log.Info().
-			Str("listen", s.conf.BindAddr).
-			Str("version", pkg.Version()).
-			Msg("trisa directory server started")
-
-		if err := s.srv.Serve(sock); err != nil {
-			s.svc.echan <- err
-		}
-	}()
+	go s.Run(sock)
```
bbengfort, line 74:
```suggestion
	go s.Run(sock)
	log.Info().Str("listen", s.conf.BindAddr).Str("version", pkg.Version()).Msg("trisa directory server started")
```

pkg/gds/gds.go
bbengfort: It looks fine to me! Would you mind moving the log statement back to the `Serve` method? I think it makes a bit more sense there and I believe I did the same thing in trtl. I've added a suggestion to make it easy.

pkg/gds/gds_test.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"context"
+	"time"
+
+	api "github.com/trisacrypto/trisa/pkg/trisa/gds/api/v1beta1"
+)
+
+// TestStatus tests that the Status RPC returns the correct status response.
+func (s *gdsTestSuite) TestStatus() {
+	s.LoadEmptyFixtures()
```
bbengfort, line 12:
```suggestion
```

Since this test doesn't require the database at all, it's better to not specify any fixtures - e.g. the test is completely agnostic to the fixture. That will minimize the switching the test package does.

pkg/gds/gds_test.go
```diff
@@ -0,0 +1,35 @@
...
+	require.NoError(err)
+	require.Equal(api.ServiceState_HEALTHY, status.Status)
+
+	// Timestamps should be close to expected.
+	notBefore, err := time.Parse(time.RFC3339, status.NotBefore)
+	require.NoError(err)
+	require.True(notBefore.Sub(expectedNotBefore) < time.Minute)
+	notAfer, err := time.Parse(time.RFC3339, status.NotAfter)
+	require.NoError(err)
+	require.True(notAfer.Sub(expectedNotAfter) < time.Minute)
```
bbengfort, line 34: Good test!

pkg/utils/listener/listener.go
```diff
@@ -0,0 +1,45 @@
...
+// connection under which to send client calls, Close() to close the connection, and
+// Release() to close the listener.
+type GRPCListener struct {
+	Listener *bufconn.Listener
+	Conn     *grpc.ClientConn
+}
+
+func New(bufSize int) *GRPCListener {
+	return &GRPCListener{
+		Listener: bufconn.Listen(1024 * 1024),
```
bbengfort, line 23:
```suggestion
		Listener: bufconn.Listen(bufSize),
```

pkg/utils/listener/listener.go
```diff
@@ -0,0 +1,45 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/test/bufconn"
+)
+
+// GRPCListener handles gRPC connections using a bufconn listener. This is useful for
+// testing when it's unnecessary to have a live gRPC server running. The normal
+// workflow is to call New() to start the listener, Connect() to start a gRPC
+// connection under which to send client calls, Close() to close the connection, and
+// Release() to close the listener.
+type GRPCListener struct {
```
bbengfort, line 16: Nice - I like how you've compiled all this functionality into a single place to share between trtl and gds! It simplifies things which is nice! 

Would you mind if we put this in `pkg/utils/bufconn/listener.go`? - listener is a bigger thing and I think we'll remember bufconn for grpc tests a bit easier.

pkg/utils/listener/listener.go
```diff
@@ -0,0 +1,45 @@
+package listener
```
bbengfort, line 1:
```suggestion
package bufconn
```

bbengfort (approved):
@pdeziel looks great, thank you for consolidating the bufconn interactions into a single place and unifying the functionality between trtl and gds! The mock looked pretty straight forward - I had a couple of comments, suggestions, and one suggested change, but please feel free to merge when you're ready.

bbengfort (approved):
@pdeziel looks great, thank you for consolidating the bufconn interactions into a single place and unifying the functionality between trtl and gds! The mock looked pretty straight forward - I had a couple of comments, suggestions, and one suggested change, but please feel free to merge when you're ready.

pkg/utils/listener/listener.go
pdeziel: Not a problem with me, I was trying to avoid having conflicting names with other packages to keep things simple, but I don't have a better suggestion either.

PR #226, opened 2021-11-22
sc-2167 authentication middleware test

pdeziel:
This adds a table test to make sure that the admin API routes do not accept unauthenticated requests.

### DIFF
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -130,11 +130,6 @@ func (s *Admin) Shutdown() (err error) {
 	return nil
 }
 
-// Routes returns the Admin API router for testing purposes.
-func (s *Admin) Routes() http.Handler {
-	return s.router
-}
-
 func (s *Admin) setupRoutes() (err error) {
 	// Application Middleware
 	s.router.Use(ginzerolog.Logger("gin"))
@@ -1490,4 +1485,9 @@ func (s *Admin) Available() gin.HandlerFunc {
 // GetTokenManager returns the underlying token manager for testing.
 func (s *Admin) GetTokenManager() *tokens.TokenManager {
 	return s.tokens
+}
+
+// Routes returns the Admin API router for testing purposes.
+func (s *Admin) Routes() http.Handler {
+	return s.router
 }
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -1487,7 +1487,7 @@ func (s *Admin) GetTokenManager() *tokens.TokenManager {
 	return s.tokens
 }
 
-// Routes returns the Admin API router for testing purposes.
-func (s *Admin) Routes() http.Handler {
+// GetRouter returns the Admin API router for testing purposes.
+func (s *Admin) GetRouter() http.Handler {
 	return s.router
 }
--- a/pkg/gds/admin_test.go
+++ b/pkg/gds/admin_test.go
@@ -8,11 +8,13 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"sort"
+	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


pkg/gds/admin.go
```diff
@@ -1491,3 +1486,8 @@ func (s *Admin) Available() gin.HandlerFunc {
 func (s *Admin) GetTokenManager() *tokens.TokenManager {
 	return s.tokens
 }
+
+// Routes returns the Admin API router for testing purposes.
+func (s *Admin) Routes() http.Handler {
```
bbengfort, line 1491: If you'd like to name it `GetRoutes` to match the other accessors, that's fine with me; I think this is a holdover from Whisper and the tests I wrote for it.

pkg/gds/mock.go
```diff
@@ -46,6 +46,11 @@ func NewMock(conf config.Config) (s *Service, err error) {
 	if admin.tokens, err = tokens.MockTokenManager(); err != nil {
 		return nil, err
 	}
+	gin.SetMode(admin.conf.Mode)
+	admin.router = gin.New()
+	if err = admin.setupRoutes(); err != nil {
+		return nil, err
+	}
```
bbengfort, line 53: If we get too much closer to the `NewAdmin` function, we might want to consider having that setup the mock instead of doing it in this function. Let's keep an eye on it.

pkg/gds/admin_test.go
```diff
@@ -86,6 +88,41 @@ func (s *gdsTestSuite) doRequest(handle gin.HandlerFunc, c *gin.Context, w *http
 	return res
 }
 
+// Test that the middleware returns the corect error when making unauthenticated
+// requests to protected endpoints.
+func (s *gdsTestSuite) TestAuthenticationMiddleware() {
+	s.svc.GetAdmin().SetHealth(true)
```
bbengfort, line 94: Thought it might be useful to add a comment here.

```suggestion
	// We're not directly running the admin server so we need to manually set it as healthy       
	s.svc.GetAdmin().SetHealth(true)
```

pkg/gds/admin_test.go
```diff
@@ -86,6 +88,41 @@ func (s *gdsTestSuite) doRequest(handle gin.HandlerFunc, c *gin.Context, w *http
...
+		{"review", http.MethodPost, "/v2/vasps/42/review", http.StatusUnauthorized},
+		{"resend", http.MethodPost, "/v2/vasps/42/resend", http.StatusUnauthorized},
+		{"listReviewNotes", http.MethodGet, "/v2/vasps/42/notes", http.StatusUnauthorized},
+		{"createReviewNote", http.MethodPost, "/v2/vasps/42/notes", http.StatusUnauthorized},
+		{"updateReviewNote", http.MethodPut, "/v2/vasps/42/notes/1", http.StatusUnauthorized},
+		{"deleteReviewNote", http.MethodDelete, "/v2/vasps/42/notes/1", http.StatusUnauthorized},
+	} {
+		s.T().Run(endpoint.name, func(t *testing.T) {
+			serv := httptest.NewServer(s.svc.GetAdmin().Routes())
+			defer serv.Close()
```
bbengfort, line 116: Any reason you're running the server inside of the test closure? Servers are designed to handle multiple requests in isolation, I think it might be better to extract this to the outer function scope just to make the tests a bit more efficient.

bbengfort (approved):
Excellent! I'm glad the table-driven tests made this straightforward!

bbengfort:
@pdeziel not sure if you're waiting on me or not - but I took a look at your changes - thank you! Feel free to merge when you're ready.

PR #228, opened 2021-11-23
Map VASP to CertificateRequests

bbengfort:
Adds `certificate_requests` to the `GDSExtraData` field to store Certificate Request IDs on the VASP so we don't have to loop over all certificate requests to find the ones associated with a VASP. 

- Adds `GetCertReqIDs` and `AppendCertReqIDs` functions to the `models.go` extra helpers
- Saves the Certificate Request ID on the VASP on Register
- Uses `GetCertReqIDs` on approve/reject instead of looping over all certreqs in the database 

NOTE: this PR will require a database migration before we deploy GDS again!


pkg/gds/gds.go
```diff
@@ -202,23 +202,36 @@ func (s *GDS) Register(ctx context.Context, in *api.RegisterRequest) (out *api.R
...
+		return nil, status.Error(codes.Internal, "internal error with registration, please contact admins")
+	}
+
+	// Add the CertificateRequest to the VASP
+	if err = models.AppendCertReqID(vasp, certRequest.Id); err != nil {
+		log.Error().Err(err).Str("vasp", vasp.Id).Msg("could not add cert request to VASP")
+		return nil, status.Error(codes.Internal, "internal error with registration, please contact admins")
+	}
+
+	// Store VASP with updated certificate requests
```
bbengfort, line 231: @pdeziel can't wait to have GDS service tests to check stuff like this!

pkg/gds/admin.go
```diff
@@ -1102,28 +1102,39 @@ func (s *Admin) acceptRegistration(vasp *pb.VASP, claims *tokens.Claims) (msg st
...
+			continue
+		}
+
+		if careq.Status == models.CertificateRequestState_INITIALIZED {
+			if err = models.UpdateCertificateRequestStatus(careq, models.CertificateRequestState_READY_TO_SUBMIT, "registration request received", claims.Email); err != nil {
 				return "", err
 			}
-			if err = s.db.UpdateCertReq(req); err != nil {
+			if err = s.db.UpdateCertReq(careq); err != nil {
 				return "", err
```
pdeziel, line 1132: This check is much easier to parse split out like this, thanks.

pkg/gds/models/v1/models.go
```diff
@@ -124,6 +124,56 @@ func AppendAuditLog(vasp *pb.VASP, entry *AuditLogEntry) (err error) {
...
+			return fmt.Errorf("could not deserialize previous extra: %s", err)
+		}
+	} else {
+		extra.CertificateRequests = make([]string, 0, 1)
+	}
+
+	// Do not allow duplicate certificate requests to be appended
+	for _, containsID := range extra.CertificateRequests {
+		if certreqID == containsID {
+			// Do not return an error
```
pdeziel, line 162: Do we expect this to happen during normal operation or should we be logging a warning?

pkg/gds/models/v1/models_test.go
```diff
@@ -255,6 +271,40 @@ func TestReviewNotes(t *testing.T) {
...
+
+	// Append an ID with data already inside
+	err = AppendCertReqID(vasp, "9676bf6a-ffdb-4185-8fa5-87cdae6f6eef")
+	require.NoError(t, err)
+	ids, err = GetCertReqIDs(vasp)
+	require.NoError(t, err)
+	require.Len(t, ids, 2)
+
+	// Do not allow duplicate IDs
+	err = AppendCertReqID(vasp, "9676bf6a-ffdb-4185-8fa5-87cdae6f6eef")
```
pdeziel, line 301: Thanks for including this test case!

pdeziel (approved):
I had one comment in `models.go` but other than that it looks good to go!

pkg/gds/models/v1/models.go
bbengfort: It's mostly just a convenience to make sure we don't do anything wrong in the migration - Append is only being called once in the code right now so I'm not too worried about it. Great question though, thank you!

PR #230, opened 2021-11-23
GDS Admin API: Update VASP record Endpoint

bbengfort:
UpdateVASP is a single entry point to a variety of different patches that can be made
to the VASP object. In particular, the user may update the business details (website,
categories, and established on), update the IVMS 101 Legal Person entity, change
their responses to the TRIXO form, update the common name or endpoint, or manage
contact details. Although technically, this endpoint would allow all those changes to
be made simultaneously, the idea is that the PATCH only happens inside of those
collections or groups of fields. Individual update methods define the logic for how
each of those groups is updated together.

There are a lot of TODOs that I couldn't get to from this story including:

- updating contacts (and possibly resending verification emails)
- writing tests for all the update cases 

Hopefully when we review we can spot the TODOs in code and create stories from them. This PR will unblock @elysee15 though.


rebeccabilbro (approved):
Thank you for adding this new feature to support updating VASPs more systematically in GDS! And thanks also for the thorough live code review last week. I've attempted to capture all the notes I had during our discussion, and opened up some new Shortcut stories to capture future work. Let me know if you have any questions!

PR #232, opened 2021-11-23
sc-2180 add sectigo client to service mock

pdeziel:
This adds the sectigo client to the service mock for the GDS test suite. This should enable GDS tests to use the mocked sectigo server which is enabled behind the `Testing` flag in `SectigoConfig`, as well as future integration tests.


pkg/gds/mock.go
```diff
@@ -101,6 +114,7 @@ func MockConfig() config.Config {
 			Username: "foo",
 			Password: "supersecretsquirrel",
 			Profile:  "CipherTrace EE",
+			Testing:  true,
```
pdeziel, line 117: Not sure if the mocked server should be running by default, or if the tests that require it should specify it via a new config.

pkg/gds/service.go
```diff
@@ -114,6 +120,7 @@ type Service struct {
 	admin  *Admin
 	conf   config.Config
 	certs  *sectigo.Sectigo
+	mock   *mock.Server
```
pdeziel, line 123: The idea here is to have a way to start the mocked Sectigo server when running locally or via integration tests.

pkg/sectigo/profiles/profiles.go
```diff
@@ -1,4 +1,4 @@
-package sectigo
+package profiles
```
pdeziel, line 1: This is a temporary workaround so that `sectigo.go` and `config.go` don't reference each other.

bbengfort: This seems like a bit much for just a few constants. I'd like to refactor as follows:

1. Move `config.SectigoConfig` to `sectigo.Config` in `pkg/sectigo/config.go` (including validation)
2. Replace the profiles.go
3. Make the `config.Sectigo` property a `sectigo.Config` 

I think this should solve the problem of the recursive import and I think that `sectigo.Config` is much neater and decoupling anyway.

cmd/sectigo/main.go
```diff
@@ -217,7 +219,12 @@ func main() {
 }
 
 func initAPI(c *cli.Context) (err error) {
-	if api, err = sectigo.New(c.String("username"), c.String("password"), c.String("profile")); err != nil {
+	conf := config.SectigoConfig{
+		Username: c.String("username"),
+		Password: c.String("password"),
+		Profile:  c.String("profile"),
+	}
+	if api, err = sectigo.New(conf); err != nil {
```
bbengfort, line 227: I definitely like using the config in new - it simplifies things a lot.

cmd/sectigo/main.go
```diff
@@ -291,7 +298,7 @@ func createSingle(c *cli.Context) (err error) {
 		fmt.Printf("pkcs12 password: %s\n", params["pkcs12Password"])
 	}
 
-	if api.Profile() == sectigo.ProfileCipherTraceEndEntityCertificate || api.Profile() == sectigo.ProfileIDCipherTraceEndEntityCertificate {
+	if api.Profile() == sectigoprofiles.ProfileCipherTraceEndEntityCertificate || api.Profile() == sectigoprofiles.ProfileIDCipherTraceEndEntityCertificate {
```
bbengfort, line 301: See below, I'd like to revert this back to the `sectigo` package.

pkg/gds/certs.go
```diff
@@ -128,7 +129,7 @@ func (s *Service) submitCertificateRequest(r *models.CertificateRequest) (err er
 	params["pkcs12Password"] = string(pkcs12Password)
 
 	profile := s.certs.Profile()
-	if profile == sectigo.ProfileCipherTraceEndEntityCertificate || profile == sectigo.ProfileIDCipherTraceEndEntityCertificate {
+	if profile == profiles.ProfileCipherTraceEndEntityCertificate || profile == profiles.ProfileIDCipherTraceEndEntityCertificate {
```
bbengfort, line 132: See below, I'd like to revert this back to the `sectigo` package.

pkg/gds/mock.go
bbengfort: This is a good question - it's kind of a chicken and an egg problem. I suppose I'd rather have Sectigo itself start the mock server if testing is true so that the tests and everything outside don't have to worry about it. We could add a module level `sectigo.Cleanup()` to close and reset the test server. More on that later.

pkg/gds/mock.go
```diff
@@ -63,6 +65,17 @@ func NewMock(conf config.Config) (s *Service, err error) {
...
+
+	if svc.certs, err = sectigo.New(conf.Sectigo); err != nil {
+		return nil, err
+	}
+
+	if conf.Sectigo.Testing {
+		if svc.mock, err = mock.New(); err != nil {
+			return nil, err
+		}
+	}
```
bbengfort, line 77: This will probably not be necessary, see below comments.

pkg/gds/service.go
```diff
@@ -70,10 +71,15 @@ func New(conf config.Config) (s *Service, err error) {
...
+	if s.certs, err = sectigo.New(conf.Sectigo); err != nil {
+		return nil, err
+	}
+
+	// Start mocked Sectigo server if testing is enabled
+	if conf.Sectigo.Testing {
+		if s.mock, err = mock.New(); err != nil {
+			return nil, err
+		}
 	}
```
bbengfort, line 83: This will not be necessary, see below comments.

pkg/gds/service.go
bbengfort: I understand what the motivation is, but I'd rather not have the `Service` know anything about the mocking, it should just know about the Sectigo client since that's all it uses and we want to make sure there is a distinct separation of concerns. If Testing=True in the conf, Sectigo should go into mocked mode. I'll comment on that shortly, but likely we'll want to remove this.

pkg/gds/service.go
```diff
@@ -172,6 +179,11 @@ func (s *Service) Shutdown() (err error) {
 		log.Error().Err(err).Msg("could not shutdown DirectoryAdministration service")
 	}
 
+	// Shutdown the Sectigo mock server
+	if s.mock != nil {
+		s.mock.Close()
+	}
```
bbengfort, line 185: Will likely also be removed per other comments.

pkg/sectigo/sectigo.go
```diff
@@ -40,16 +42,16 @@ type Sectigo struct {
 // $SECTIGO_USERNAME and $SECTIGO_PASSWORD respectively; alternatively if not given and
 // not stored in the environment, as long as valid access credentials are cached the
 // credentials will be loaded.
-func New(username, password, profile string) (client *Sectigo, err error) {
+func New(conf config.SectigoConfig) (client *Sectigo, err error) {
```
bbengfort, line 45: I think that we'll have to do checking if we're in testing mode at this point and startup the Mock server if we are. There may be some issues with recursive imports though, so we'll have to figure that out. 

My suggestion is to create a module level variable `mockSectigoServer *mock.Server` and on `New` if `conf.Testing` then we create the new `mockSectigoServer`. We should create module level functions `GetMock`, `ResetMock`, and `CloseMock` that allow test code to manipulate the module level variable, checking if it is `nil` and handling it appropriately. 

I'm hoping that the only thing we'll have to change to prevent recursive imports, is to have the sectigo code set it's baseurl rather than having the mock server do it. 

Let me know if that makes sense.

bbengfort (commented):
@pdeziel a very good start - I definitely like the sectigo configuration! We're getting a little tangled up in coupling, so I've provided some refactoring suggestions that I'm hoping will make the components more loosely coupled, particularly the dependencies between the GDS Service, its configuration, the Sectigo client and the mock server. Happy to discuss if you have any questions!

pkg/sectigo/sectigo.go
pdeziel: Makes sense but unfortunately resolving the recursive import looks like it's a bit more complicated. In the mock package we need access to the endpoints and the Request/Reply structs, which currently exist in the sectigo package.

bbengfort: Yeh, things are super gnarly aren't they. My first instinct is to move the mock back into the `sectigo` package and have no external packages call it, but we need them to be able to modify the handlers on demand for specific tests. 

Perhaps we can chat about this tomorrow morning and see if we can figure out a solution? We might just have to move the mock back into the test code.

pkg/gds/mock.go
```diff
@@ -67,6 +69,17 @@ func NewMock(conf config.Config) (s *Service, err error) {
 		return nil, err
 	}
 	svc.admin = admin
+
+	if conf.Sectigo.Testing {
+		if err = mock.Start(); err != nil {
```
bbengfort, line 74: Not for this PR but we may need to change this to `sectigo_mock` or something so that it's more understandable in the future.

pkg/gds/service.go
```diff
@@ -70,10 +70,8 @@ func New(conf config.Config) (s *Service, err error) {
 	}
 
 	// Create the Sectigo API client
-	if s.certs, err = sectigo.New(conf.Sectigo.Username, conf.Sectigo.Password, conf.Sectigo.Profile); err != nil {
-		log.Error().Err(err).Msg("could not initialize sectigo client")
-		// HACK: to get local testing going we're just ignoring this error - we should instead mock the Sectigo client.
-		// return nil, err
+	if s.certs, err = sectigo.New(conf.Sectigo); err != nil {
+		return nil, err
```
bbengfort, line 74: Thank you for cleaning up that technical debt!

pkg/sectigo/mock/mock.go
```diff
@@ -34,29 +36,36 @@ type Server struct {
...
 }
 
-// Close the test server to complete the tests and cleanup.
-func (s *Server) Close() {
-	s.server.Close()
+// Stop the test server and reset the Sectigo server URL to complete the tests and cleanup.
+func Stop() {
+	if mockServer != nil {
+		mockServer.server.Close()
+	}
```
bbengfort, line 68: Good panic guard.

pkg/sectigo/sectigo_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/gin-gonic/gin"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
-	. "github.com/trisacrypto/directory/pkg/sectigo"
+	"github.com/trisacrypto/directory/pkg/sectigo"
```
bbengfort, line 14: Didn't like importing everything from `sectigo` into the current namespace? Was it causing an issue?

bbengfort (approved):
@pdeziel well done! You neatly skirted the issue with the recursive imports and I think this code is very clean, (more) loosely coupled, and easily understandable!

pkg/sectigo/sectigo_test.go
pdeziel: I don't remember why I made that change but I just reverted it since it makes the code a bit more readable and doesn't appear to break anything.

PR #234, opened 2021-11-24
sc-2163 test case for successful Admin API authentication

pdeziel:
This adds the missing test cases for successful authentication on the Admin API `Authenticate` endpoint. The Admin API handler uses Google's `idtoken.Validate()` and is expecting a Google token. I got around this by adding a `validate` function to the token manager which must be configured when a new `TokenManager` is instantiated. When we're running the tests, the mocked `TokenManager` will be used which uses a mocked validation function instead of the Google one. The mocked function simply parses the token and injects the claims into the returned payload, bypassing all the checks that the Google handler is doing. This allows us to test that we are handling authentication properly assuming that we were passed a valid credential.

### DIFF
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -247,7 +247,7 @@ func (s *Admin) Authenticate(c *gin.Context) {
 	}
 
 	// Validate the credential with Google
-	if claims, err = idtoken.Validate(c.Request.Context(), in.Credential, s.conf.Oauth.GoogleAudience); err != nil {
+	if claims, err = s.tokens.Validate(c.Request.Context(), in.Credential, s.conf.Oauth.GoogleAudience); err != nil {
 		log.Warn().Err(err).Msg("invalid credentials used for authentication")
 		c.JSON(http.StatusUnauthorized, admin.ErrorResponse("invalid credentials"))
 		return
--- a/pkg/gds/mock.go
+++ b/pkg/gds/mock.go
@@ -88,7 +88,7 @@ func MockConfig() config.Config {
 			CookieDomain: "admin.gds.dev",
 			Audience:     "http://api.admin.gds.dev",
 			Oauth: config.OauthConfig{
-				GoogleAudience:         "4284607864536-notarealgoogleaudience.apps.googleusercontent.com",
+				GoogleAudience:         "http://localhost",
 				AuthorizedEmailDomains: []string{"gds.dev"},
 			},
 			TokenKeys: nil,
--- a/pkg/gds/tokens/mock.go
+++ b/pkg/gds/tokens/mock.go
@@ -1,10 +1,14 @@
 package tokens
 
 import (
+	"context"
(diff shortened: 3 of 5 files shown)


pkg/gds/admin.go
```diff
@@ -247,7 +247,7 @@ func (s *Admin) Authenticate(c *gin.Context) {
 	}
 
 	// Validate the credential with Google
-	if claims, err = idtoken.Validate(c.Request.Context(), in.Credential, s.conf.Oauth.GoogleAudience); err != nil {
+	if claims, err = s.tokens.Validate(c.Request.Context(), in.Credential, s.conf.Oauth.GoogleAudience); err != nil {
```
rebeccabilbro, line 250: Guessing this was just a stray bug you discovered while testing?

pkg/gds/mock.go
```diff
@@ -88,7 +88,7 @@ func MockConfig() config.Config {
 			CookieDomain: "admin.gds.dev",
 			Audience:     "http://api.admin.gds.dev",
 			Oauth: config.OauthConfig{
-				GoogleAudience:         "4284607864536-notarealgoogleaudience.apps.googleusercontent.com",
+				GoogleAudience:         "http://localhost",
```
rebeccabilbro, line 91: Changed to make the integration testing easier?

pkg/gds/tokens/mock.go
```diff
@@ -25,3 +30,40 @@ func MockTokenManager() (tm *TokenManager, err error) {
 	tm.keys[tm.currentKeyID] = &key.PublicKey
 	return tm, nil
 }
+
+// mockValidate is a mock for Google's token validation which enables authentication
+// testing by internally parsing the given token and injecting the claims into the
+// returned payload. This will only validate badly constructed tokens and tokens where
+// the supplied audience does not match the audience used to sign the token.
```
rebeccabilbro, line 37: You mention that this mock will only allow us to catch bad tokens and mismatched audiences; are there other edge cases we should be thinking about moving forward?

rebeccabilbro (approved):
This looks good to me @pdeziel! Abstracting out the validation mechanism to make it easier to test seems to have worked nicely -- I left you a few questions/comments (not as familiar with the Admin API so just trying to get up to speed), but I think this should be good to merge 👍

pkg/gds/mock.go
pdeziel: The other option was to change the secrets mock to use the fake google audience above, for testing purposes they just have to match so I went for the simpler option.

pkg/gds/tokens/mock.go
pdeziel: Good question, we could also verify that the expiration time and issued time make sense which isn't that complicated of a check. Currently we're passing basically all of the token validation off to Google so from a testing perspective there's only really one validation error path, so I didn't find it absolutely necessary to include those checks in the mock for the scope of what this PR is testing.

PR #235, opened 2021-11-24
Updates trtl with the new honu release

rebeccabilbro:
This PR updates the trtl RPC methods slightly given the latest version of honu, which now performs namespace handling. I've also streamlined the methods to make them more consistent (sc-2063), including:

- [x] checking that log messages are consistent
- [x] checking that docstrings are consistent
- [x] creating helper functions for shared code


pkg/trtl/honu.go
```diff
@@ -30,72 +30,68 @@ func NewHonuService(s *Server) (*HonuService, error) {
 }
 
 // Get is a unary request to retrieve a value for a key.
+// If metadata is requested in the GetRequest, the request will use honu.Object() to
+// retrieve the entire object, including the metadata. If no metadata is requested, the
+// request will use honu.Get() to get only the value.
+// If a namespace is provided, the namespace is passed to the internal honu Options,
+// to look in that namespace only.
```
pdeziel, line 37: Thanks for improving the description!

pkg/trtl/honu.go
```diff
@@ -30,72 +30,68 @@ func NewHonuService(s *Server) (*HonuService, error) {
...
+
+	// But we do have to check if we have a namespace
 	var value []byte
-	if value, err = h.db.Get(key); err != nil {
-		// TODO: Check for the honu not found error instead.
+	if in.Namespace != "" {
+		value, err = h.db.Get(in.Key, options.WithNamespace(in.Namespace))
+	} else {
+		value, err = h.db.Get(in.Key)
+	}
```
pdeziel, line 86: We'll probably want to abstract this out as well at some point to maintain our own sanity.

pkg/trtl/honu.go
```diff
@@ -590,23 +527,7 @@ func (h *HonuService) Cursor(in *pb.CursorRequest, stream pb.Trtl_CursorServer)
...
-					Version: object.Version.Version,
-					Region:  object.Version.Region,
-				},
-				Parent: &pb.Version{
-					Pid:     object.Version.Parent.Pid,
-					Version: object.Version.Parent.Version,
-					Region:  object.Version.Parent.Region,
-				},
-			}
+			msg.Meta = returnMeta(*object)
```
pdeziel, line 530: Any particular reason why we need to dereference the Honu object here?

pkg/trtl/honu.go
```diff
@@ -636,12 +557,22 @@ func (h *HonuService) Sync(stream pb.Trtl_SyncServer) (err error) {
...
+			Pid:     object.Version.Pid,
+			Version: object.Version.Version,
+			Region:  object.Version.Region,
+		},
+		Parent: &pb.Version{
+			Pid:     object.Version.Parent.Pid,
+			Version: object.Version.Parent.Version,
+			Region:  object.Version.Parent.Region,
+		},
+	}
```
pdeziel, line 577: Love that this is its own function now.

pkg/trtl/honu.go
```diff
@@ -30,72 +30,68 @@ func NewHonuService(s *Server) (*HonuService, error) {
...
 		var object *object.Object
-		if object, err = h.db.Object(key); err != nil {
+		if in.Namespace != "" {
+			object, err = h.db.Object(in.Key, options.WithNamespace(in.Namespace))
+		} else {
+			object, err = h.db.Object(in.Key)
+		}
+
+		if err != nil {
 			// TODO: Check for the honu not found error instead.
```
pdeziel, line 63: I think this should be unblocked now because the newest version of Honu should expose the NotFound error?

pdeziel (approved):
Looks good, I had a few comments but nothing blocking the merge.

pkg/trtl/honu.go
rebeccabilbro: Not sure exactly, I tried it without and got this as a suggested change...

pkg/trtl/honu.go
rebeccabilbro: Great idea! Any suggestions?

PR #236, opened 2021-11-25
sc-2187 GDS register test

pdeziel:
This adds a test for the GDS Register RPC. The test itself is pretty straightforward, but I did have to add the mocked secret manager to the mocked service to get the password code to work, and change up how we're starting/stopping the bufconn listener due to a bug I encountered.

### DIFF
--- a/pkg/gds/gds_test.go
+++ b/pkg/gds/gds_test.go
@@ -4,8 +4,106 @@ import (
 	"context"
 	"time"
 
+	"github.com/trisacrypto/directory/pkg/gds/models/v1"
 	api "github.com/trisacrypto/trisa/pkg/trisa/gds/api/v1beta1"
+	pb "github.com/trisacrypto/trisa/pkg/trisa/gds/models/v1beta1"
 )
+
+// TestRegister tests that the Register RPC correcty registers a new VASP with GDS.
+func (s *gdsTestSuite) TestRegister() {
+	s.LoadEmptyFixtures()
+	defer s.ResetEmptyFixtures()
+	require := s.Require()
+	ctx := context.Background()
+	refVASP := s.fixtures[vasps]["d9da630e-41aa-11ec-9d29-acde48001122"].(*pb.VASP)
+
+	// Start the gRPC client
+	require.NoError(s.grpc.Connect())
+	defer s.grpc.Close()
+	client := api.NewTRISADirectoryClient(s.grpc.Conn)
+
+	// Emails need to be filled in for a valid VASP registration. Need to make copies
+	// of the contacts here to avoid modifying the fixtures for other tests.
+	contacts := *refVASP.Contacts
+	admin := *refVASP.Contacts.Administrative
+	contacts.Administrative = &admin
+	contacts.Administrative.Email = "admin@example.com"
+	billing := *refVASP.Contacts.Billing
+	contacts.Billing = &billing
+	contacts.Billing.Email = "billing@example.com"
+	legal := *refVASP.Contacts.Legal
+	contacts.Legal = &legal
+	contacts.Legal.Email = "legal@example.com"
+	technical := *refVASP.Contacts.Technical
+	contacts.Technical = &technical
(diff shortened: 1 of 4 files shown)


pkg/gds/gds_test.go
```diff
@@ -4,9 +4,107 @@ import (
...
+	ctx := context.Background()
+	refVASP := s.fixtures[vasps]["d9da630e-41aa-11ec-9d29-acde48001122"].(*pb.VASP)
+
+	// Start the gRPC client
+	require.NoError(s.grpc.Connect())
+	defer s.grpc.Close()
+	client := api.NewTRISADirectoryClient(s.grpc.Conn)
+
+	// Emails need to be filled in for a valid VASP registration. Need to make copies
+	// of the contacts here to avoid modifying the fixtures for other tests.
```
pdeziel, line 26: It might be easier to modify one of the fixtures so that all the contacts contain an email.

pkg/gds/service_test.go
```diff
@@ -345,11 +344,26 @@ func (s *gdsTestSuite) loadFixtures(ftype fixtureType, fpath string) {
...
 	// Create the new service with a database to the specified path
 	conf := gds.MockConfig()
 	conf.Database.URL = "leveldb:///" + s.dbPaths[ftype]
 	s.svc, err = gds.NewMock(conf)
 	require.NoError(err, "could not create mock GDS service")
+
+	// Start the new GDS server using a bufconn listener to avoid network requests
+	s.grpc = bufconn.New(bufSize)
+	go s.svc.GetGDS().Run(s.grpc.Listener)
+
```
pdeziel, line 366: I was completed baffled for about 2 hours as to why the GDS database appeared to be closed when calling the `Register` RPC. Then I realized that we were replacing `s.svc` here with an entirely new service, which includes a brand new GDS, but the gRPC server was still pointing to the old GDS with the old database! :facepalm: So I've moved the code which starts the new gRPC server here instead of in `SetupTestSuite`.

rebeccabilbro: Ooft, that was definitely a tricky one. Great catch!

rebeccabilbro (approved):
Nice work @pdeziel -- looks like you caught at least a few bugs in the process of working on this one test, which is pretty impressive. Nice to feel like we can have confidence in our tests!

pkg/gds/secrets/mock.go
```diff
@@ -68,20 +68,25 @@ func (c *mockSecretManagerClient) CreateSecret(ctx context.Context, req *smpb.Cr
 	}
 
 	// Handle the expiration
-	switch expires := req.Secret.Expiration.(type) {
-	case *smpb.Secret_ExpireTime:
-		secret.Expires = expires.ExpireTime.AsTime()
-		if secret.Expires.IsZero() {
-			return nil, status.Error(codes.InvalidArgument, "invalid expiration time")
+	if req.Secret.Expiration != nil {
```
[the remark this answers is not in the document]
pdeziel: I was actually getting an invalid expiration type error, I figured it would be okay to set a default here since this is a mock.

pkg/gds/secrets/mock.go
```diff
@@ -68,20 +68,25 @@ func (c *mockSecretManagerClient) CreateSecret(ctx context.Context, req *smpb.Cr
...
-		ttl := expires.Ttl.AsDuration()
-		if ttl < 1 {
-			return nil, status.Error(codes.InvalidArgument, "invalid time to live")
-		}
-		secret.Expires = secret.Created.Add(ttl)
-	default:
-		return nil, status.Error(codes.InvalidArgument, "unknown expiration type")
+	} else {
+		// Default to 1 hour expiration
+		secret.Expires = secret.Created.Add(time.Hour)
```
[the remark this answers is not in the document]
pdeziel: I didn't put much thought into it since I was just trying to unblock the tests, something like 10 minutes is probably more suitable for a registration password.

PR #240, opened 2021-11-26
Remove peers from GDS (sc-1754)

rebeccabilbro:
We have migrated Peers from GDS to trtl, and GDS should no longer have direct access to replicas/peers. This PR removes the implementation of Peers from GDS and migrates the `wire` package to `utils` so that it can be used across the codebase.


bbengfort (approved):
Looks good - thank you so much for taking care of this technical debt! :shipit:

PR #242, opened 2021-11-29
sc-2189 verification RPC test

pdeziel:
This adds a test for the GDS Verification RPC.

### DIFF
--- a/pkg/gds/gds.go
+++ b/pkg/gds/gds.go
@@ -347,8 +347,8 @@ func (s *GDS) Search(ctx context.Context, in *api.SearchRequest) (out *api.Searc
 	return out, nil
 }
 
-// Status returns the status of a VASP including its verification and service status if
-// the directory service is performing health check monitoring.
+// Verification returns the status of a VASP including its verification and service
+// status if the directory service is performing health check monitoring.
 func (s *GDS) Verification(ctx context.Context, in *api.VerificationRequest) (out *api.VerificationReply, err error) {
 	var vasp *pb.VASP
 	switch {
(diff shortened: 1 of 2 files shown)


pkg/gds/gds_test.go
```diff
@@ -105,6 +106,59 @@ func (s *gdsTestSuite) TestRegister() {
...
+	s.LoadFullFixtures()
+	require := s.Require()
+	ctx := context.Background()
+
+	id := "d9da630e-41aa-11ec-9d29-acde48001122"
+
+	// The reference fixture doesn't contain the updated timestamp, so we retrieve the
+	// real VASP object here for comparison purposes.
+	vasp, err := s.svc.GetStore().RetrieveVASP(id)
+	require.NoError(err)
```
pdeziel, line 121: The problem here is that when `generateDB` creates the VASPs in the store the `lastUpdated` timestamp gets updated, meaning the field on the VASP is different than the field in the reference fixture.

pkg/gds/gds.go
```diff
@@ -347,8 +347,8 @@ func (s *GDS) Search(ctx context.Context, in *api.SearchRequest) (out *api.Searc
 	return out, nil
 }
 
-// Status returns the status of a VASP including its verification and service status if
-// the directory service is performing health check monitoring.
+// Verification returns the status of a VASP including its verification and service
```
rebeccabilbro, line 350: Thank you for catching this docstring error!

pkg/gds/gds.go
```diff
@@ -372,7 +372,7 @@ func (s *GDS) Verification(ctx context.Context, in *api.VerificationRequest) (ou
 
 		vasp = vasps[0]
 	default:
-		log.Warn().Str("rpc", "lookup").Msg("no arguments supplied")
+		log.Warn().Str("rpc", "verification").Msg("no arguments supplied")
```
rebeccabilbro, line 375: Very glad that you're doing these tests 🙂  good copypasta 🍝  catch!

rebeccabilbro (approved):
Very good PR @pdeziel, nice catches with those copy and paste errors!

PR #243, opened 2021-11-29
sc-2188 lookup and search tests

pdeziel:
This adds tests for the GDS lookup and search RPCs.


pkg/gds/gds_test.go
```diff
@@ -105,6 +106,160 @@ func (s *gdsTestSuite) TestRegister() {
...
+		VerifiedOn:          vasp.VerifiedOn,
+		Name:                "CharlieBank",
+	}
+
+	// VASP exists in the database
+	request.Id = id
+	reply, err := client.Lookup(ctx, request)
+	require.NoError(err)
+	require.True(proto.Equal(expected, reply))
+	// TODO: Check that a signing certificate is returned
```
pdeziel, line 146: I modified the fakerize script to produce signing certificates so I could test this case, but realized that the script is producing new VASP and CertReq IDs every time it's run. This breaks the logic of the tests because we are assuming that the "source of truth" fixture IDs are constant. We might have to modify the tests to not care about specific fixture IDs and use more meaningful names (e.g., "CharlieBank") or modify the fakerize script to produce consistent UUIDs, but that is definitely out of the scope of this PR.

rebeccabilbro: Good point! I think either of these suggestions would work -- either fixing the uuids in `fakerize.py` or changing the tests to compare the names rather than uuids. I see that you've opened sc-2282 for us to revisit this, so I think we're good to go here 👍

pkg/gds/gds_test.go
```diff
@@ -105,6 +106,160 @@ func (s *gdsTestSuite) TestRegister() {
...
+	}
+	reply, err = client.Search(ctx, request)
+	require.NoError(err)
+	require.Empty(reply.Error)
+	require.Len(reply.Results, 0)
+
+	// Filter by VASP category
+	request = &api.SearchRequest{
+		Name:         []string{"CharlieBank"},
+		VaspCategory: []string{"Miner"},
```
rebeccabilbro, line 245: These VASP Category fields (e.g. "Miner", "Project", etc)  are also randomly selected by `fakerize.py`, so we might need to address these in sc-2282 too.

rebeccabilbro (approved):
Looks great @pdeziel, and thank you for opening up sc-2282 for us to address the issues with our fakerized test fixtures!

PR #244, opened 2021-11-30
feat: implement search bar in header

elysee15:
Fixes SC-548

### DIFF
--- a/web/gds-admin-ui/src/components/TopbarSearch.js
+++ b/web/gds-admin-ui/src/components/TopbarSearch.js
@@ -1,70 +1,13 @@
-// @flow
 import React from 'react';
 import { Link } from 'react-router-dom';
 import Select, { components } from 'react-select';
+import { useDispatch, useSelector } from 'react-redux';
+import useSafeDispatch from 'hooks/useSafeDispatch';
+import { fetchAutocomplete } from 'redux/actions';
+import { fetchAllAutocomplete } from 'redux/selectors';
+import { useHistory } from 'react-router-dom';
 import classNames from 'classnames';
 
-/*
- * get options
- */
-const optionGetter = (option) => {
-    switch (option.type) {
-        case 'report':
-            return (
-                <Link to="/" className={classNames('dropdown-item', 'notify-item', 'p-0')}>
-                    <i className={classNames(option.icon, 'font-16', 'me-1')}></i>
-                    <span>{option.label}</span>
-                </Link>
-            );
-        case 'help':
-            return (
-                <Link to="/" className={classNames('dropdown-item', 'notify-item', 'p-0')}>
-                    <i className={classNames(option.icon, 'font-16', 'me-1')}></i>
-                    <span>{option.label}</span>
-                </Link>
-            );
-        case 'settings':
-            return (
-                <Link to="/" className={classNames('dropdown-item', 'notify-item', 'p-0')}>
-                    <i className={classNames(option.icon, 'font-16', 'me-1')}></i>
-                    <span>{option.label}</span>
-                </Link>
-            );
-        case 'title':
-            return (
-                <div className="noti-title">
-                    <h6 className="text-overflow mb-2 text-uppercase">Users</h6>
-                </div>
-            );
-        case 'users':
-            return (
-                <>
-                    <Link to="/" className="dropdown-item notify-item p-0">
-                        <div className="d-flex">
-                            <img
-                                src={option.userDetails.avatar}
-                                alt=""
-                                className="d-flex me-2 rounded-circle"
-                                height="32"
-                            />
-                            <div className="w-100">
-                                <h5 className="m-0 font-14">
-                                    {option.userDetails.firstname} {option.userDetails.lastname}
-                                </h5>
-                                <span className="font-12 mb-0">{option.userDetails.position}</span>
-                            </div>
-                        </div>
-                    </Link>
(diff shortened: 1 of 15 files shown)


web/gds-admin-ui/src/redux/autocomplete/saga.js
```diff
@@ -0,0 +1,20 @@
+const { call, put, takeEvery } = require("redux-saga/effects")
+const { default: getAllAutocomplete } = require("services/autocomplete")
+const { fetchAutocompleteApiResponseSuccess, FetchAutocompleteActionTypes, fetchAutocompleteApiResponseError } = require(".")
+
+function* fetchAutocompletes() {
```
bbengfort, line 5: I haven't seen the `function*` keyword before, what does this do? Is it an ES2017 thing?

web/gds-admin-ui/src/components/TopbarSearch.js
```diff
@@ -1,67 +1,69 @@
...
+//                             </div>
+//                         </div>
+//                     </Link>
+//                 </>
+//             );
+
+//         default:
+//             return;
+//     }
+// };
```
bbengfort, line 66: I think we can probably get rid of this once we're ready to move forward - it's unlikely that we'll have multiple types of options in the near future, maybe just some icons and we can reference the original code right? Let me know if you're uncomfortable removing this code for some other reason.

web/gds-admin-ui/src/components/TopbarSearch.js
```diff
@@ -108,38 +96,77 @@ const MenuList = (props) => {
 
 /* fomates the option label */
 const handleFormatOptionLabel = (option) => {
-    const formattedOption = optionGetter(option);
-    return <div>{formattedOption}</div>;
+    return <Link className='text-muted d-block' to={`/vasps/${option.value}`}>{option.label}</Link>;
```
bbengfort, line 99: Any reason you're not using `className={classNames('dropdown-item', 'notify-item', 'p-0')}` similar to the items in the example? The class might need an inner span e.g. `<span>{option.label}</span>` but I think that might look better? Or is the `d-block` for overflow? There is also a `text-overflow` class.

web/gds-admin-ui/src/components/TopbarSearch.js
```diff
@@ -108,38 +96,77 @@ const MenuList = (props) => {
...
+
+    const loadOptions = input => {
+        return new Promise((resolve, reject) => {
+            if (input.length < 2) {
+                return resolve([])
+            }
+
+            setTimeout(() => {
+                return resolve(filteredOptions(input))
+            }, 500)
```
bbengfort, line 135: Is this throttling to prevent a huge number of API requests as the user types?

web/gds-admin-ui/src/components/TopbarSearch.js
```diff
@@ -108,38 +96,77 @@ const MenuList = (props) => {
...
+    const handleInputChange = async (input = '') => {
+        setIsLoading(true)
+        setInputValue(input)
+        if (input.length < 2) {
+            setMenuIsOpen(false)
+            setOptions([])
+            setIsLoading(false)
+        } else {
+            setMenuIsOpen(true)
+            const options = await loadOptions(input)
```
bbengfort, line 147: We're always loading the auto complete options from the API right? We're not cacheing them in redux and cache busting after a couple of minutes?

bbengfort (commented):
@elysee15 great start, I think it looks really good and this is going to be a very useful tool for navigating in the dashboard! I've added a couple of comments and I have a video in the Shortcut story with some more questions. Thank you!

web/gds-admin-ui/src/redux/autocomplete/saga.js
elysee15: It's an es6 thing, we call them `generators`

web/gds-admin-ui/src/components/TopbarSearch.js
elysee15: I just forgot to remove it

web/gds-admin-ui/src/components/TopbarSearch.js
elysee15: No, it’s to make it a little asynchronous since I receive all the data in one request

elysee15: No, it’s to make it a little asynchronous since I receive all the data in one request

web/gds-admin-ui/src/components/TopbarSearch.js
elysee15: The process is that I 
- request `GET /autocomplete` api endpoint and get all autocompletes 
- add it in the store
- retrieve it and filter

web/gds-admin-ui/src/components/TopbarSearch.js
bbengfort: Ok, I would refer to that as throttling - but I'm glad it's in the code!

web/gds-admin-ui/src/components/TopbarSearch.js
elysee15: Yes

bbengfort (approved):
@elysee15 thank you so much for making those changes - the user experience is much nicer now. I also think that the formatting style changes are much more readable.

PR #245, opened 2021-11-30
sc-2190 VerifyContact test

pdeziel:
This adds a test for the GDS VerifyContact RPC.


pkg/gds/gds_test.go
```diff
@@ -105,6 +107,57 @@ func (s *gdsTestSuite) TestRegister() {
...
+	request.Token = "invalid"
+	_, err = client.VerifyContact(ctx, request)
+	require.Error(err)
+
+	// TODO: Test previously verified contact - requires modifying the fixtures to
+	// include a non-empty verification token
+
+	// Successful verification
+	request.Token = ""
+	reply, err := client.VerifyContact(ctx, request)
```
pdeziel, line 143: This only works right now because we currently don't check if the input token is empty, which we probably should. We need to modify the test fixtures to include the verification token on some of the contacts so that we can properly test this.

rebeccabilbro: Should we try to address this in sc-2282 or is this a different fixture?

pkg/gds/admin_test.go
```diff
@@ -920,6 +921,7 @@ func (s *gdsTestSuite) TestReview() {
 func (s *gdsTestSuite) TestResend() {
 	s.LoadFullFixtures()
 	defer s.ResetFullFixtures()
+	defer emails.PurgeMockEmails()
```
rebeccabilbro, line 924: Guessing this just deletes temp files?

rebeccabilbro (approved):
Just a few minor questions for you @pdeziel but otherwise looks good!

pkg/gds/gds_test.go
pdeziel: Yeah we should try to address it in sc-2282.

pkg/gds/admin_test.go
pdeziel: We are storing the "mocked" emails as raw bytes in a [][]byte slice. Since it's global, any test that writes to it needs to clear it so we can maintain independence between tests. I had forgotten to do that when I originally wrote this test.

PR #247, opened 2021-12-06
Make trtl region and pid test vars

rebeccabilbro:
This PR resolves sc-2098, doing a bit of cleanup on the environment variables for the trtl & honu tests.

### DIFF
--- a/pkg/trtl/honu_test.go
+++ b/pkg/trtl/honu_test.go
@@ -4,8 +4,8 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"fmt"
 	"io"
+	"strconv"
 
 	"github.com/trisacrypto/directory/pkg/trtl/pb/v1"
 	codes "google.golang.org/grpc/codes"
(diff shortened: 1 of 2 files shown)


pkg/trtl/honu_test.go
```diff
@@ -269,26 +264,21 @@ func (s *trtlTestSuite) TestDelete() {
 		},
 	})
 	require.NoError(err)
-	// TODO update os.Getenv to use test fixtures in sc-2098
-	pid := os.Getenv("TRTL_REPLICA_PID")
-	expectedPID, err := strconv.Atoi(pid)
-	require.NoError(err)
-	expectedRegion := os.Getenv("TRTL_REPLICA_REGION")
-	owner := bytes.Join([][]byte{[]byte(pid), []byte(expectedRegion)}, []byte(":"))
+	owner := bytes.Join([][]byte{[]byte(fmt.Sprint(metaPID)), []byte(metaRegion)}, []byte(":"))
```
pdeziel, line 267: The convention I try to follow is to use `strconv.Itoa` when converting integers to strings, and `fmt.Sprint` when something needs to be formatted for printing output, although I'm not sure what the underlying difference actually is.

pdeziel (approved):
Looks good to me, I had a small comment but it probably shouldn't block the merge.

rebeccabilbro:
Look like this is failing due to a docker 500 error, likely related to the ongoing us-east-1 outage. Will try again later

PR #248, opened 2021-12-07
Hotfix: Handle Sectigo COLLECTED batch detail status

bbengfort:
This PR implements the batch status Sectigo API endpoint. If the batch
detail for a certificate request comes back with an unknown status then
the certificate manager attempts to lookup the batch status directly.
Logging for this error is improved.

Fixes SC-2353


rebeccabilbro (approved):
Looks good @bbengfort, hopefully this makes the next Sectigo API gotcha a little less hectic for us!

PR #249, opened 2021-12-08
refact: review JWT authentication

elysee15:
Fixes SC-1631

This refactor aims to modify the way I managed the refresh of the token which was not efficient since at each change of route there was enough verification so I preferred to do the verification when I intercept an error 401.
[axios-auth-refresh](https://www.npmjs.com/package/axios-auth-refresh) has some limits when it comes to rerun more than one failed requests that's the reason of this logic in the 401 condition

Screen Shot 2021-12-13 at 12 57 59


web/gds-admin-ui/src/components/ProfileDropdown.js
```diff
@@ -1,30 +1,20 @@
...
 
     const toggleDropdown = () => {
         setDropdownOpen(!dropdownOpen);
     };
 
-    const handleLogoutClick = () => {
-        api.deleteUserSession()
+    const handleLogoutClick = async () => {
+        await logout()
     }
```
bbengfort, line 18: Much nicer!

web/gds-admin-ui/src/helpers/api/apiCore.js
```diff
@@ -1,40 +1,87 @@
 import axios from 'axios';
 import { defaultEndpointPrefix, getCookie } from 'utils';
-import jwtDecode from 'jwt-decode'
-import toast from 'react-hot-toast';
 
-axios.defaults.headers.post['Content-Type'] = 'application/json';
-axios.defaults.baseURL = defaultEndpointPrefix();
+const instance = axios.create({
```
bbengfort, line 4: Creating the instance and using the interceptors like this seems a lot simpler and the code is much reader. Very well written!

bbengfort (approved):
@elysee15 sorry it's taken me so long to review this - I've been putting it off because I wanted to spend the time to review it in detail. Now that I have, I wish I hadn't - because this refactor is _excellent_ and is exactly what I hoped for when we discussed interceptors. 

It is easy to see that the quality of the code has improved a lot because the code is much simpler and more understandable than it was before. The `AuthProvider` was a very nice touch, as were the custom hooks. I think this is neatly done and improves the security of the Admin UI significantly. Very well done!

PR #251, opened 2021-12-08
sc-2354 add contact email to audit log on GDS RPCs

pdeziel:
This makes some changes to the GDS RPCs to add the contact email as the "source" for the audit logs. Also, new audit log entries are created for verified contacts.

### DIFF
--- a/pkg/gds/gds.go
+++ b/pkg/gds/gds.go
@@ -117,10 +117,6 @@ func (s *GDS) Register(ctx context.Context, in *api.RegisterRequest) (out *api.R
 		VerificationStatus:  pb.VerificationState_NO_VERIFICATION,
 		Version:             &pb.Version{Version: 1},
 	}
-	if err := models.UpdateVerificationStatus(vasp, pb.VerificationState_SUBMITTED, "register request recevied", "automated"); err != nil {
-		log.Warn().Err(err).Msg("could not update VASP verification status")
-		return nil, status.Error(codes.Aborted, "could not add new entry to VASP audit log")
-	}
 
 	// Compute the common name from the TRISA endpoint if not specified
 	if vasp.CommonName == "" && vasp.TrisaEndpoint != "" {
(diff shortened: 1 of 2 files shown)


pkg/gds/gds.go
```diff
@@ -140,6 +136,19 @@ func (s *GDS) Register(ctx context.Context, in *api.RegisterRequest) (out *api.R
 		log.Warn().Err(err).Msg("ignoring validation error")
 	}
 
+	// Retrieve email address from one of the supplied contacts.
+	var email string
+	if email = getContactEmail(vasp); email == "" {
```
pdeziel, line 141: Not sure what email we should be using here if multiple contacts have emails, right now it's using the same order defined in other parts of the code.

pkg/gds/gds.go
```diff
@@ -140,6 +136,19 @@ func (s *GDS) Register(ctx context.Context, in *api.RegisterRequest) (out *api.R
...
+		log.Warn().Msg("no contact email address found")
+		return nil, status.Error(codes.InvalidArgument, "no email address in supplied VASP contacts")
```
bbengfort, line 143: In the [`vasp.Validate()`](https://github.com/trisacrypto/trisa/blob/main/pkg/trisa/gds/models/v1beta1/models.go#L107-L132) step we do check to make sure there is at least one contact and that all contacts supplied have email addresses. So these lines of code should never fire. Because of that could we change this to:

```suggestion
		log.Error().Err(errors.New("no contact email address found")).Msg("incorrect access on validated VASP")
		return nil, status.Error(codes.InvalidArgument, "no email address in supplied VASP contacts")
```

E.g. I want to make sure we get a critical alert if this case fires for some reason. 

As for the return - I'm happy to return an invalid argument, but only if we haven't saved anything to the database prior to this step. Would you mind checking to make sure there are no database modifications before we get here? (Otherwise we should return aborted and raise the stakes even higher for the log alert).

pkg/gds/gds.go
bbengfort: I think that makes sense - we don't have any idea who is submitting the registration, so our best guess is expressed in the order that you specified. Thank you!

bbengfort (approved):
Looks great, thank you for adding this - it will make understanding the audit logs so much easier! I just had one minor comment, then feel free to merge!

PR #253, opened 2021-12-09
Translate header texts

tl6kk:
The headers seem to have been handled in [previous PR](https://github.com/trisacrypto/directory/pull/195/files). We haven't sent these additional messages (35 missing) for translation. 

I added in Chinese translations for the headers, and this is what it currently looks like:
Screen Shot 2021-12-09 at 3 26 48 PM

One issue I found is the `Switch to TestNet` in the `TopNav`. It seems to be loaded only once, and won't be changed when the locale is switched. Only when I modify the code to cause the UI to refresh on the `zh` page, it will render the translated `跳转至TestNet`. This might be related to this implementation:

```
const getDirectoryURL = () => {
  if (isTestNet()) {
      return [
          "https://vaspdirectory.net",
          t`Switch to Production`,
          t`You're currently on the TestNet Directory`,
      ]    
  }

  return [
          "https://trisatest.net",
          t`Switch to TestNet`,
          t`You're currently on the Production Directory`,
      ]
};
```

Not sure if we should change the implementation to a `MainNav` and a `TestNav`, like what we did with the header?


bbengfort:
@elysee15 would you mind reviewing this PR please?

elysee15 (approved):

PR #256, opened 2021-12-09
sc-1861 backup manager tests

pdeziel:
This adds some tests for the backup manger.


pkg/gds/backup.go
```diff
@@ -78,6 +83,14 @@ backups:
...
 			}
 		}
+
+		if stop != nil {
+			// Stop if someone wrote to the stop channel
+			for range stop {
+				log.Info().Msg("backup manager received stop signal")
+				return
+			}
+		}
```
pdeziel, line 93: The backup manager keeps running once it's started, so I added some synchronization here so we can start/stop it within the context of a single test.

pkg/gds/backup_test.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	s.SetConfig(conf)
+	defer s.ResetConfig()
+	s.LoadEmptyFixtures()
+	require := s.Require()
+
+	// Start the backup manager
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go s.svc.BackupManager(nil, &wg)
```
bbengfort, line 30: Above I mentioned that I didn't think that the wait group was necessary to pass to the `BackupManager` because I believe this is simpler in the test case:

```suggestion
	go func() {
		defer wg.Done()
		s.svc.BackupManager(nil)
	}()
```

Now the backup manager doesn't have to worry about the wait group which is only used for tests since the BackupManager should return immediately since it's disabled. 

Note, however, I also mentioned in an early comment that it might be nice to add a stop signal to the `s.Shutdown()` command to gracefully shutdown both the CertManager and the BackupManager -- in that case these would both need a wait group that both routines would be a part of. That wait group would have to be [contextual](https://stackoverflow.com/a/32843750/488917). So it's up to you what you'd like to do - if you think we should add a `Shutdown` step, then we can simply comment that the wait groups are used for tests but will eventually also be used for shutdown. Otherwise, we can go the route suggested above if the wait groups are just for tests.

pkg/gds/backup_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// Wait for the backup interval to elapse
+	time.Sleep(s.svc.GetConf().Backup.Interval * 2)
+
+	// Backup should not be created
+	backupDir := s.svc.GetConf().Backup.Storage
+	require.NoDirExists(backupDir)
+
+	// Make sure the backup manager is stopped before we exit
+	wg.Wait()
```
bbengfort, line 40: Because our expectation is that the backup routine is not running, this could replace the Sleep on Line 33. I do think we should have the Wait before we check the backupDir if you want to keep the sleep, just to prevent a race condition.

pkg/gds/backup_test.go
```diff
@@ -0,0 +1,83 @@
...
+	for f := range files {
+		if files[f].IsDir() {
+			numBackups++
+		}
+	}
+	require.Equal(1, numBackups, "wrong number of backups created")
+
+	// Make sure that the backup manager is stopped before we exit
+	stop <- true
+	wg.Wait()
```
bbengfort, line 82: Should we do this before we check the backups were created, e.g. right after the sleep?

pkg/gds/backup_test.go
```diff
@@ -0,0 +1,83 @@
...
+	defer s.ResetConfig()
+	s.LoadEmptyFixtures()
+	defer os.RemoveAll(s.svc.GetConf().Backup.Storage)
+	require := s.Require()
+
+	// Start the backup manager
+	stop := make(chan bool)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go s.svc.BackupManager(stop, &wg)
```
bbengfort, line 62: Same suggestion as above:
```suggestion
	go func() {
		defer wg.Done()
		s.svc.BackupManager(stop)
	}()
```

pkg/gds/service.go
```diff
@@ -138,7 +138,7 @@ func (s *Service) Serve() (err error) {
 		go s.CertManager()
 
 		// Start the backup manager go routine process
-		go s.BackupManager()
+		go s.BackupManager(nil, nil)
```
bbengfort, line 141:
```suggestion
		go s.BackupManager(nil)
```

bbengfort: I don't know if it's possible to accept my suggestions and have a working PR - but I've added this one in for completeness just in case - feel free to ignore.

bbengfort (commented):
@pdeziel very well done and thoughtfully composed! I really like your idea of adding synchronization to the backup routine and I think this will also apply to the cert manager routine as well (not sure if we have tests for that yet or not). I have a couple of Go language comments that should make things easier, and then a philosophical discussion about go routines and shutdown that I'd like to talk about.

pdeziel:
@bbengfort Thanks for your suggestions and helping me simplify the synchronization code, this will also be helpful for the CertManager tests once I get around to them. The purpose of the waitgroup was to just enable testing, but we could revisit the `Shutdown` idea in a future story if we want to. There is one lint error that I've been wrestling with; it's the reason why I had the for loop in there originally and I don't exactly know what syntax it wants here:

`pkg/gds/backup.go:82:3: should use a simple channel send/receive instead of select with a single case (S1000)`

pdeziel:
@bbengfort I've updated the select statement as we discussed in our call, should be ready for another look now.

bbengfort (approved):
Looks good, thanks for making those changes!

PR #257, opened 2021-12-09
Iter and Cursor Seek

bbengfort:
This PR refactors the `Iter` RPC to use Honu seek that was added in v0.2.4 - however in the course of making this change (which I thought was going to be simple) I discovered some namespace handling bugs. Those bugs are resolved in [github.com/rotationalio/honu #12](https://github.com/rotationalio/honu/pull/12) so this PR is blocked until that PR is merged. 

The better namespace handling also allowed me to simplify some of the namespace logic in the code, and to guard against changes, I implemented another test for a standard series of access (Get, Put, Get, Delete, Get, Put, Get). 

The last thing to do in this PR is to consider adding `Seek` to the cursor implementation.


go.mod
```diff
@@ -4,17 +4,27 @@ go 1.16
 
 require (
 	cloud.google.com/go v0.82.0
+	github.com/DataDog/zstd v1.5.0 // indirect
+	github.com/cespare/xxhash/v2 v2.1.2 // indirect
+	github.com/cockroachdb/errors v1.8.6 // indirect
+	github.com/cockroachdb/logtags v0.0.0-20211118104740-dabe8e521a4f // indirect
+	github.com/cockroachdb/pebble v0.0.0-20211208200752-b10ca3a5c8f1 // indirect
+	github.com/cockroachdb/redact v1.1.3 // indirect
```
bbengfort, line 12: I believe most of these dependencies are added by the Honu pebble dependency (and the ones below).

go.mod
```diff
@@ -4,17 +4,27 @@ go 1.16
...
 	github.com/google/go-querystring v1.1.0
 	github.com/google/uuid v1.2.0
 	github.com/googleapis/gax-go v1.0.3
 	github.com/joho/godotenv v1.3.0
 	github.com/kelseyhightower/envconfig v1.4.0
-	github.com/rotationalio/honu v0.2.3
+	github.com/klauspost/compress v1.13.6 // indirect
+	github.com/kr/pretty v0.3.0 // indirect
+	github.com/rogpeppe/go-internal v1.8.0 // indirect
+	github.com/rotationalio/honu v0.2.5-0.20211209153613-4aa259a30af1
```
bbengfort, line 27: TODO: to get the tests to pass in this PR, I'm directly working off of the PR banch in Honu; however before we merge this PR we need to accurately set the Honu dependency to a tagged dependency. I'm thinking of resetting the `v0.2.4` dependency, but we can also bump it to `v0.2.5` if you think that's best.

proto/trtl/v1/trtl.proto
```diff
@@ -94,8 +94,9 @@ message BatchReply {
 
 message CursorRequest {
     bytes prefix = 1;       // the prefix to range over, if nil all objects are returned
-    string namespace = 2;
-    Options options = 3;
+    bytes seek_key = 2;     // a key to seekto within the prefix (optional)
+    string namespace = 3;
+    Options options = 4;
```
bbengfort, line 99: The idea behind `seek_key` is that the user wants to move forward inside of a common prefix range. For example, let's say the keys are:

- aa1
- aa2
- aa3
- aa4
- aa5
- b1
- b1
- b3

If the user only cares about keys that start with 'a', but wants to start iterating from a3, they would make the request as follows:

```go
&CursorRequest{Prefix: []byte('a'), SeekKey: []byte('a3')}
```

pkg/trtl/server.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	// Create the server and prepare to serve
+	s = &Server{
+		conf:  conf,
+		echan: make(chan error, 1),
+		srv:   grpc.NewServer(grpc.UnaryInterceptor(s.interceptor)),
+	}
+
+	// TODO: check for maintenance mode
```
rebeccabilbro, line 69: Just a heads up that you'll need to handle a merge conflict issue here; we merged in the maintenance mode PR before you did the file renaming. I believe that merge will also impact the file now called `trtl/server_test.go`

pkg/trtl/trtl.go
```diff
@@ -1,156 +1,566 @@
...
 	}
 
-	// Listen for TCP requests
-	var sock net.Listener
-	if sock, err = net.Listen("tcp", t.conf.BindAddr); err != nil {
-		log.Error().Err(err).Str("bindaddr", t.conf.BindAddr).Msg("could not listen on given bindaddr")
-		return err
+	// Load defaults into options
+	if opts.PageSize == 0 {
+		opts.PageSize = 100
```
rebeccabilbro, line 204: I wonder if we should pull this out and make it a global variable at the top of the file so that it's easier to find in the future if we need to do any troubleshooting around pagination?

pkg/trtl/trtl.go
```diff
@@ -1,156 +1,566 @@
...
+			return nil, status.Error(codes.InvalidArgument, "invalid page token")
+		}
+
+		// Validate the request has not changed
+		if cursor.PageSize != opts.PageSize {
+			log.Debug().Int32("cursor", cursor.PageSize).Int32("opts", opts.PageSize).Msg("invalid iter request: mismatched page size")
+			return nil, status.Error(codes.InvalidArgument, "page size cannot change between requests")
+		}
+
+		// Note - prefix check happens on next key, but namespace check must match Honu iterator
```
rebeccabilbro, line 234: Thank you for adding these comments to help future us remember!

pkg/trtl/trtl.go
```diff
@@ -1,156 +1,566 @@
...
+
+	// Check if there is a next page cursor
+	if len(cursor.NextKey) != 0 {
+		if out.NextPageToken, err = cursor.Dump(); err != nil {
+			log.Error().Err(err).Str("namespace", in.Namespace).Msg("could not serialize next page token")
+			return nil, status.Error(codes.FailedPrecondition, "could not serialize next page token")
+		}
+	}
+
+	// Request complete
```
rebeccabilbro, line 332: Just want to check that in our Prometheus integration this is where we'd want to increment the Iter counter and not anywhere else before this?

pkg/trtl/trtl_test.go
```diff
@@ -1,177 +1,689 @@
...
+	// Test invalid page token
+	token := "CAISLHBlb3BsZTo6NDZlNzg5MTctOGQyMC00N2MwLWIwZDEtZTUyMDQxNDlhOTM2"
+	_, err = client.Iter(ctx, &pb.IterRequest{Options: &pb.Options{PageToken: "foo"}})
+	s.StatusError(err, codes.InvalidArgument, "invalid page token")
+	_, err = client.Iter(ctx, &pb.IterRequest{Namespace: "people", Options: &pb.Options{PageToken: token, PageSize: 27}})
+	s.StatusError(err, codes.InvalidArgument, "page size cannot change between requests")
+	_, err = client.Iter(ctx, &pb.IterRequest{Namespace: "things", Options: &pb.Options{PageToken: token, PageSize: 2}})
+	s.StatusError(err, codes.InvalidArgument, "namespace cannot change between requests")
+	_, err = client.Iter(ctx, &pb.IterRequest{Prefix: []byte("zed"), Namespace: "people", Options: &pb.Options{PageToken: token, PageSize: 2}})
+	s.StatusError(err, codes.InvalidArgument, "prefix cannot change between requests")
```
rebeccabilbro, line 435: Thank you for reintegrating these checks!

pkg/trtl/trtl.go
```diff
@@ -1,156 +1,566 @@
...
+		log.Error().Err(err).Str("namespace", in.Namespace).Msg("could not create honu iterator")
+		return nil, status.Errorf(codes.FailedPrecondition, "could not create iterator: %s", err)
+	}
+	defer iter.Release()
+
+	// Perform namespace check to ensure that the page cursor matches the iterator namespace
+	// NOTE: this section must come after the iterator is created, though it would be preferable
+	// if it was in the section where the page cursor is created. This is because we want to
+	// check the namespace that the iterator is operating on rather than the one specified by
+	// the user in the first request (e.g. because a default namespace may be used).
```
rebeccabilbro, line 263: good comment!

proto/trtl/internal/pagination.proto
```diff
@@ -10,6 +10,7 @@ option go_package = "github.com/trisacrypto/directory/pkg/trtl/internal";
 // size in the cursor matches the page size in the request.
 // See https://cloud.google.com/apis/design/design_patterns#list_pagination for more.
 message PageCursor {
-    int32 page_size = 1; // the number of results returned on each iteration.
-    bytes next_key = 2;  // the key to start the iteration from
+    int32 page_size = 1;  // the number of results returned on each iteration.
+    bytes next_key = 2;   // the key to start the iteration from
+    string namespace = 3; // the namespace the cursor is iterating on
```
rebeccabilbro, line 15: nice addition that allows us to more straightforwardly compare `cursor.Namespace` and `iter.Namespace`!

rebeccabilbro (approved):
This looks good to me @bbengfort -- thank you for your thorough annotation of the new Iter and Seek functionalities, which will help us remember some of the intricacies of these functions after we've moved on to other development. Thank you also for the file renaming to help as we cross-develop `trtl` and `honu`!  I've left you a handful of minor comments for your consideration and one or two questions for my understanding. 

No good deed goes unpunished 👼 😈  so just a heads up about some exciting collisions -- the merge of this PR will require rebasing (from the trtl maintenance mode PR), will require #260 to get rebased, and is currently blocking the Prometheus integration.

pkg/trtl/trtl.go
bbengfort: Yes, you're right - I've done that to make it more obvious. 

However, I think that when we finally hone in on what we think will be the stable API for options, we should create a `LoadOptions` method or similar to load options with the defaults so all the defaults across all methods are in one place. Probably it should also do the work of merging the options specified by the user as well to replace all of lines 193-205 and the duplicated code in the other RPCs :-)  -- I don't want to do this though until we're feeling confident that `Options` is how we're going to move forward and not flattening these arguments into the request RPCs.

pkg/trtl/trtl.go
bbengfort: I believe so, yes. I think of the `Iter` as a single read even though it's of multiple keys since the read is across a snapshot; therefore when the scan is complete, that's the time to instrument it (as opposed to each individual key fetch). If we do server-side statefulness for pagination then we might want to create an additional Prometheus metric for the number of open page cursors.

PR #258, opened 2021-12-09
Trtl maintenance mode sc-2047

rebeccabilbro:
This PR introduces a maintenance mode for trtl that is similar to the maintenance mode we have for GDS; it allows the underlying trtl database to be accessible outside the context of normal operations (e.g. migrations, repairs, etc). 

One observation is that we may want to create another story for introducing a `Status` endpoint for trtl, since that is not currently defined in the protos; this is probably related to the work we'll do in sc-952.

### DIFF
--- a/pkg/trtl/config/config.go
+++ b/pkg/trtl/config/config.go
@@ -10,6 +10,9 @@ import (
 	"github.com/trisacrypto/directory/pkg/utils/logger"
 )
 
+// Config defines the struct that is expected to initialize the trtl server
+// Note: because we need to validate the configuration, `config.New()`
+// must be called to ensure that the `processed` is correctly set
 type Config struct {
 	Maintenance bool                `split_words:"true" default:"false"`
 	BindAddr    string              `split_words:"true" default:":4435"`
--- a/pkg/trtl/trtl_test.go
+++ b/pkg/trtl/trtl_test.go
@@ -150,6 +150,8 @@ func (s *trtlTestSuite) SetupSuite() {
 	os.Setenv("TRTL_DATABASE_URL", "leveldb:///"+s.db)
 	os.Setenv("TRTL_REPLICA_PID", fmt.Sprint(metaPID))
 	os.Setenv("TRTL_REPLICA_REGION", metaRegion)
+	os.Setenv("TRTL_MAINTENANCE", "false")
+
 	s.conf, err = config.New()
 	require.NoError(err)
 
--- a/pkg/trtl/trtl_test.go
+++ b/pkg/trtl/trtl_test.go
@@ -169,6 +169,7 @@ func (s *trtlTestSuite) SetupSuite() {
 
 func (s *trtlTestSuite) TearDownSuite() {
 	require := s.Require()
+	require.NoError(s.trtl.GetDB().Close())
 	err := os.RemoveAll(s.db)
 	require.NoError(err)
 	s.grpc.Release()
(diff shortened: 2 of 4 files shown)


pkg/trtl/trtl_test.go
```diff
@@ -175,3 +177,18 @@ func (s *trtlTestSuite) TearDownSuite() {
...
+	conf.Maintenance = true
+	conf.Database.URL = "test"
+	conf.Replica.Region = "tauceti"
+
+	var server *trtl.Server
+	server, err := trtl.New(conf)
+	require.Nil(server)
+	require.Error(err)
+	require.Equal(err.Error(), "honu error: resource temporarily unavailable")
+}
```
bbengfort, line 194: I believe the `"honu error: resource temporarily unavailable"` is occurring because `trtl.New` is attempting to [open a leveldb database](https://github.com/syndtr/goleveldb/issues/76) that is already opened by the TestSuite on [Line 159](https://github.com/trisacrypto/directory/pull/258/files#diff-bda168dd057e1a36e57dbf0cfc5f8fd1e080d956722a53f7e59fb7e9682fff49R159). When we start the server in maintenance mode, we should expect no errors, e.g. the server is ready to serve and respond to requests as unavailable. 

However, the odd thing is that the `trtl.New` call should not be opening a database at all - the database should be nil. So unless I'm mistaken about why the error is occurring, something a little strange is going on here that we should probably look into. 

I think we want to assert that `server.db` is nil in this test, however `server.db` is private. One of the ways that we've been allowing access to private variables by our tests is to add accessor functions, e.g. the [GDS accessor functions in `pkg/gds/service.go` Lines 185-207](https://github.com/trisacrypto/directory/blob/main/pkg/gds/service.go#L185-L207). I recommend we do the same thing here, at the bottom of `pkg/trtl/trtl.go` add an accessor:

```go
// Accessors - used primarily for testing

// GetDB returns the underlying Honu database used by all sub-services.
func (t *Server) GetDB() *honu.DB {
	return t.db
}
```

We can then do the maintenance test as follows:

```suggestion
func (s *trtlTestSuite) TestMaintenance() {
	require := s.Require()

	var conf config.Config
	conf.Maintenance = true
	conf.Database.URL = "test"
	conf.Replica.Region = "tauceti"

	server, err := trtl.New(conf)
	require.NotEmpty(server, "no maintenance mode server was returned")
	require.NoError(err, "starting the server in maintenance mode caused an error")
	require.Nil(server.GetDB(), "maintenance mode database was not nil") 
}
```

Basically this test assertion is "if we're in maintenance mode, we don't connect to the database" - which is exactly what we want! I haven't addressed the issue with the error in these comments, but hopefully it will help us hone in on the problem. 

Also, would you mind modifying the [`TearDownSuite` method on Lines 168-173](https://github.com/trisacrypto/directory/blob/main/pkg/trtl/trtl_test.go#L168-L173) as follows:

```go
func (s *trtlTestSuite) TearDownSuite() {
	require := s.Require()
	require.NoError(s.trtl.GetDB().Close())
	err := os.RemoveAll(s.db)
	require.NoError(err)
	s.grpc.Release()
}
```

This will allow us to use the accessor to do a better job cleaning up the database when we're done with it, and prevent the resource temporarily unavailable error for other tests outside the trtl test suite if they exist.

bbengfort (commented):
@rebeccabilbro thank you for adding in maintenance mode - I know this was a tricky one even though it seemed small to start off with! I'm a little unsure about what's going on with the test, I had quite a long comment before that didn't address my confusion directly, but I'm hoping that they help us work toward the issue.

pkg/trtl/trtl_test.go
rebeccabilbro: @bbengfort - here are the updated test failures:


```golang
require.NotEmpty(server, "no maintenance mode server was returned")
``` 
Should NOT be empty, but was <nil>


```golang
require.NoError(err, "starting the server in maintenance mode caused an error")
```
honu error: resource temporarily unavailable


```golang
require.Nil(server.GetDB(), "maintenance mode database was not nil") 
```
test panicked: runtime error: invalid memory address or nil pointer dereference

pkg/trtl/config/config.go
```diff
@@ -10,6 +10,9 @@ import (
 	"github.com/trisacrypto/directory/pkg/utils/logger"
 )
 
+// Config defines the struct that is expected to initialize the trtl server
+// Note: because we need to validate the configuration, `config.New()`
+// must be called to ensure that the `processed` is correctly set
```
bbengfort, line 15: Thank you for adding this docstring!

bbengfort (approved):

PR #259, opened 2021-12-09
sc-2370 use consistent fixtures

pdeziel:
This changes `fakerize.py` to use more consistent fixtures by creating instances of `random.Random` with predefined seeds. The idea is that the `make_` functions take in a `random.Random` which by default uses a random seed but can be overwritten by the parent to produce consistent pseudorandom values based on the VASP or CertReq name.

The fixtures are now indexed by the first word of the VASP/CertReq in lowercase, e.g., "charliebank", which makes them easier to reference from the test code.

I also added basic CLI argument support to `fakerize.py` so that we can easily copy the fixture archive to `pkg/gds/testdata`.

A lot of changes were required in the test code, mostly updating the expected values to the newly generated ones, but hopefully we shouldn't have to do this as much in the future since the fixtures are more consistent.


rebeccabilbro (approved):
Looks good @pdeziel - thank you for the live walkthrough!

PR #260, opened 2021-12-10
Load trtl fixtures with honu

rebeccabilbro:
This PR updates trtl's tests to load fixtures with honu instead of vanilla leveldb. 

Loading with ldb allowed us to skip over honu as an initial convenience. However, when we implemented the Delete RPC method in sc-1620 , we realized that this was preventing us from being able to accurately test the version incrementation performed by honu in trtl's RPC methods, sometimes resulting in unexpected metadata.

One thing I noticed is that [trtl](https://github.com/trisacrypto/directory/blob/main/pkg/trtl/config/config.go#L28-L35) and [honu](https://github.com/rotationalio/honu/blob/dd97110bd1649887d6b417af8f0c28bb7928990f/config/config.go#L8-L16) have slightly different definitions for `ReplicaConfig`. I did eventually find a method  called `GetHonuConfig()`, though it only converts between `config.Config` and `honu.ReplicaConfig` (the docstring might cause some confusion).


pdeziel (approved):
Looks good to me once the merge conflicts are resolved. I think the `GetHonuConfig` workaround is the correct way to deal with the current config mismatch between trtl and honu.

PR #261, opened 2021-12-10
sc-1343 verify emails sent in tests

pdeziel:
This adds an assert helper which hopefully makes it easier to verify which emails are sent in the tests, by checking against both the audit log and the email mock.


bbengfort (approved):
@pdeziel nicely done - this is neat, intuitive, and well executed.

PR #262, opened 2021-12-13
sc-2347 add CreateToken test

pdeziel:
This adds a simple test for `CreateToken`.

### DIFF
--- a/pkg/gds/secrets/token_test.go
+++ b/pkg/gds/secrets/token_test.go
@@ -1,0 +1,24 @@
+package secrets_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/trisacrypto/directory/pkg/gds/secrets"
+)
+
+// Test that the CreateToken function creates a valid token of the given length.
+func TestCreateToken(t *testing.T) {
+	// Negative or zero length should return an empty string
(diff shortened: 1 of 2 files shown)


pkg/gds/secrets/token_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+// Test that the CreateToken function creates a valid token of the given length.
+func TestCreateToken(t *testing.T) {
+	// Negative or zero length should return an empty string
+	require.Equal(t, "", secrets.CreateToken(-1))
+	require.Equal(t, "", secrets.CreateToken(0))
+
+	// Returned token should not contain any unexpected characters
+	token := secrets.CreateToken(20)
+	require.True(t, secrets.ValidateToken(token), "token %s contains invalid characters", token)
```
rebeccabilbro, line 18: Should we also add a sad path test where `ValidateToken` evaluates to false?

rebeccabilbro (approved):
Looks good @pdeziel - just had one thought for your consideration.

pkg/gds/secrets/token_test.go
pdeziel: Good call, I've added a test for `ValidateToken` so I think that case should be covered now.

PR #263, opened 2021-12-13
sc-2346 CSV load test

pdeziel:
This adds some tests for `store.Load` which loads VASPs from a CSV file. In order to make the testing easier I changed the `mockdb` struct a bit, although I don't think we were using it before so it shouldn't affect anything.

### DIFF
--- a/pkg/gds/store/mockdb/mockdb.go
+++ b/pkg/gds/store/mockdb/mockdb.go
@@ -1,125 +1,159 @@
 package mockdb
 
 import (
+	"errors"
+	"fmt"
+
 	"github.com/trisacrypto/directory/pkg/gds/models/v1"
 	"github.com/trisacrypto/directory/pkg/gds/store"
 	"github.com/trisacrypto/directory/pkg/gds/store/iterator"
 	pb "github.com/trisacrypto/trisa/pkg/trisa/gds/models/v1beta1"
 )
 
-var _ store.Store = &MockDB{}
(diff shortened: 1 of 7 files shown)


pkg/gds/store/mockdb/mockdb.go
```diff
@@ -1,125 +1,159 @@
...
+	"fmt"
+
 	"github.com/trisacrypto/directory/pkg/gds/models/v1"
 	"github.com/trisacrypto/directory/pkg/gds/store"
 	"github.com/trisacrypto/directory/pkg/gds/store/iterator"
 	pb "github.com/trisacrypto/trisa/pkg/trisa/gds/models/v1beta1"
 )
 
-var _ store.Store = &MockDB{}
+var db store.Store = &MockDB{}
```
bbengfort, line 13: The line of code `var _ store.Store = &MockDB{}` is a compiler based interface check; e.g. this noop will fail at compile time if `MockDB` doesn't implement `store.Store`. 

Keeping a singleton `db` and `state` around is a pattern in our other packages, so I'm fine with the change I just wanted to mention why the previous line was there before!

bbengfort (approved):
Looks good - a very thorough set of tests!

pkg/gds/store/mockdb/mockdb.go
pdeziel: Makes sense, thanks for the clarification.

PR #264, opened 2021-12-13
GDS: Members Service

bbengfort:
This PR implements a members-only experimental service that can be used to facilitate TRISA network growth through prototyping without requiring specification changes. The number one request for the members service is a List endpoint that returns all currently verified VASPs. This RPC is defined in `members.proto` and is accessed using mTLS authentication on the members-only endpoint.


pkg/gds/models/v1/pagination.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"encoding/base64"
+	"fmt"
+
+	"google.golang.org/protobuf/proto"
+)
+
+// Load a PageCursor from a page_token string.
+// This is very similar code to trtl.internal but decoupled so Trtl and GDS aren't
```
rebeccabilbro, line 11: I agree with you on decoupling GDS and TRTL, though I think future us may find it convenient to have a unified concept of pagination throughout the project.

pkg/gds/members.go
```diff
@@ -9,13 +9,19 @@ import (
...
 	"github.com/trisacrypto/directory/pkg/gds/store"
+	pb "github.com/trisacrypto/trisa/pkg/trisa/gds/models/v1beta1"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
 
+const (
+	defaultPageSize = 100
+)
```
rebeccabilbro, line 23: 🙏

pkg/gds/members.go
```diff
@@ -97,6 +103,109 @@ func (s *Members) Shutdown() (err error) {
...
+		}
+
+		if vasp.Entity != nil {
+			info.Country = vasp.Entity.CountryOfRegistration
+		}
+
+		out.Vasps = append(out.Vasps, info)
+	}
+
+	if err = iter.Error(); err != nil {
```
rebeccabilbro, line 195: Does the call to `iter.Error` here just poll the iterator for any errors it collected during the `iter.Next` loop?

rebeccabilbro (approved):
Looks good @bbengfort - I had a few thoughts about pagination more generally (see inline comments), but nothing blocking, particularly as we want to do an alpha release of this new functionality as quickly as possible to get feedback from users! We can always circle back to some of these questions when we firm things up in a future release.

pkg/gds/members.go
bbengfort: Yes, if there is an error during processing `iter.Next()` will return false breaking the loop.

proto/gds/models/v1/models.proto
```diff
@@ -148,5 +148,5 @@ message EmailLogEntry {
 // See https://cloud.google.com/apis/design/design_patterns#list_pagination for more.
 message PageCursor {
     int32 page_size = 1;  // the number of results returned on each iteration.
-    bytes next_key = 2;   // the key to start the iteration from
+    string next_vasp = 2; // the VASP id to start the iteration from
```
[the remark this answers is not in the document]
bbengfort: It's possible that in the future we could generalize things, but the store is set up right now with `DirectoryIterator` for VASPs and a `CertificateIterator` for CertificateRequests that are tied to type. I went in hoping it would be more generic, hence `PageCursor` instead of `DirectoryPageCursor` - but we'll have to do some refactoring to get there I think.

pkg/gds/models/v1/pagination.go
bbengfort: I'm certainly open to exploring it.

PR #265, opened 2021-12-14
GDS: Fuzzy Search Fix

bbengfort:
Modifies the VASP search to allow for prefix searches if the query > 3 characters. For example the query `"alice"` will match "AliceCoin" bu the query `"al"` will not. The search remains case-insensitive. Note that searches such as `"trisa-"` or `"trisa."` now return far more results since the common name field is also included.

### DIFF
--- a/pkg/gds/gds.go
+++ b/pkg/gds/gds.go
@@ -274,7 +274,11 @@ func (s *GDS) Lookup(ctx context.Context, in *api.LookupRequest) (out *api.Looku
 		}
 
 		if len(vasps) != 1 {
-			log.Warn().Str("common_name", in.CommonName).Int("nresults", len(vasps)).Msg("wrong number of VASPs returned from search")
+			// Don't warn when common name is not found, just when multiple results are returned
+			if len(vasps) > 1 {
+				log.Warn().Str("common_name", in.CommonName).Int("nresults", len(vasps)).Msg("multiple VASPs returned from common name search in lookup")
+			}
+			log.Debug().Msg("could not lookup VASP by common name")
 			return nil, status.Error(codes.NotFound, "could not find VASP by common name")
 		}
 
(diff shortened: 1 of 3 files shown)


pkg/gds/gds.go
```diff
@@ -274,7 +274,11 @@ func (s *GDS) Lookup(ctx context.Context, in *api.LookupRequest) (out *api.Looku
 		}
 
 		if len(vasps) != 1 {
-			log.Warn().Str("common_name", in.CommonName).Int("nresults", len(vasps)).Msg("wrong number of VASPs returned from search")
+			// Don't warn when common name is not found, just when multiple results are returned
+			if len(vasps) > 1 {
+				log.Warn().Str("common_name", in.CommonName).Int("nresults", len(vasps)).Msg("multiple VASPs returned from common name search in lookup")
```
bbengfort, line 279: This will reduce some of the warning alerts we get and also make the problem more apparent.

rebeccabilbro: 🙌

pkg/gds/gds.go
```diff
@@ -375,7 +378,11 @@ func (s *GDS) Verification(ctx context.Context, in *api.VerificationRequest) (ou
 		}
 
 		if len(vasps) != 1 {
-			log.Warn().Str("common_name", in.CommonName).Int("nresults", len(vasps)).Msg("wrong number of VASPs returned from search")
+			if len(vasps) > 1 {
+				// Don't warn when common name is not found, just when multiple results are returned
+				log.Warn().Str("common_name", in.CommonName).Int("nresults", len(vasps)).Msg("multiple VASPs returned from common name search in verification")
```
rebeccabilbro, line 383: Thank you for adding "in verification" to the string here so that we can differentiate it from the similar `Lookup` case!

pkg/gds/store/leveldb/leveldb.go
```diff
@@ -282,7 +284,15 @@ func (s *Store) SearchVASPs(query map[string]interface{}) (vasps []*pb.VASP, err
 		log.Debug().Strs("name", names).Msg("search name query")
 		for _, name := range names {
 			if id := s.names[name]; id != "" {
+				// exact match
 				records[id] = struct{}{}
+			} else if len(name) > 2 {
```
rebeccabilbro, line 289: I think 3 is a totally reasonable number to go with and I'm glad that it's still possible to have 2-letter names (I remember hearing that this is far more common in Asia), just not < 3 and also fuzzy. If we think we might like to play around with length limits in the future, we could make this a global variable.

rebeccabilbro (commented):
Looks good @bbengfort - just one tiny thought re: name length, but this should be ready to merge for tomorrow's deployment!

rebeccabilbro (approved):

pkg/gds/store/leveldb/leveldb.go
bbengfort: I'm hoping that with 1 and 2 letter names (particularly if they are Chinese characters) prefix matching becomes less of an issue and the exact match case triggers more often. If we find ourselves in this situation we'll probably want to rethink the whole search methodology anyway - maybe Trtl/Honu can have search indices?

PR #268, opened 2021-12-14
sc-2439 increase VerifyContact test coverage

pdeziel:
This adds a few more test cases to the `VerifyContact` test:

1. Verification token in the request is empty.
2. One of the contacts was already verified.
3. Duplicate `VerifyContact` request.

In order to do this, I needed to modify the generated fixtures to include verification tokens on the contacts and set `verified` to false on the contacts if the VASP state is `SUBMITTED`.


pkg/gds/admin_test.go
```diff
@@ -401,7 +401,7 @@ func (s *gdsTestSuite) TestSummary() {
 		VASPsCount:           14,
 		PendingRegistrations: 5,
 		ContactsCount:        39,
-		VerifiedContacts:     28,
+		VerifiedContacts:     24,
```
pdeziel, line 404: Should we change some of these checks to be less specific?

fixtures/datagen/fakerize.py
```diff
@@ -455,6 +455,9 @@ def make_unverified(
     with open(template, "r") as f:
         record = json.load(f)
 
+    # A VASP that is not at least EMAIL_VERIFIED cannot have verified contacts
+    email_verified = (state != "SUBMITTED")
```
rebeccabilbro, line 459: Nice :)

pkg/gds/admin_test.go
rebeccabilbro: We could; alternatively we could add error messages to the tests to the effect of this to help us more quickly figure out what's going on if the tests fail
```golang
require.Len(t, actual.ContactsCount, 39, "contact count is incorrect; have the fixtures changed?")
```

pkg/gds/gds_test.go
```diff
@@ -322,55 +322,86 @@ func (s *gdsTestSuite) TestVerifyContact() {
...
+	require.Error(err)
 
-	// Audit log should contain new entries for contact verifications, EMAIL_VERIFIED,
-	// PENDING_REVIEW, along with the intitial SUBMITTED.
+	// Check audit log entries
 	log, err := models.GetAuditLog(vasp)
 	require.NoError(err)
-	// Currently verifies all contacts because the fixtures all have the empty token.
-	require.Len(log, 7)
+	require.Len(log, 5)
```
rebeccabilbro, line 391: Just confirming that this dropped from 7 to 5 because only 5 have verified contacts?

pdeziel: With the previous fixtures we were basically abusing an edge case in `VerifyContact` where an empty token string was treated as valid, meaning all four of the contacts were verified. With this PR, only two of the contacts are verified. The audit log has 5 entries in total because of the initial `SUBMITTED` state in the fixture, 2 for contact verifications (although the VASP state is unchanged), and 2 for changing the VASP state to `EMAIL_VERIFIED` and `PENDING` during the contact verification process.

PR #269, opened 2021-12-14
Prometheus metrics for Trtl

rebeccabilbro:
This PR begins to integrate Prometheus into Trtl, to allow us to track some metrics such as counts of Gets, Puts, Deletes, and Iters. Some metrics may have to wait until updates are made to honu to return more information than is currently being provided.


pkg/trtl/metrics.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/rs/zerolog/log"
+)
+
+var (
+	pmPuts  *prometheus.CounterVec // count of trtl Puts per namespace
+	pmGets  *prometheus.CounterVec // count of trtl Gets per namespace
+	pmDels  *prometheus.CounterVec // count of trtl Deletes per namespace
+	pmIters *prometheus.CounterVec // count of trtl Iters per namespace
+	// pmObjects    *prometheus.CounterVec   // count of objects being managed by trtl, by namespace
+	// pmTombstones *prometheus.CounterVec   // count of tombstones per namespace; increases on delete, decrease on overwrite of tombstone
```
rebeccabilbro, line 19: These might not be possible yet - see my notes inline; though perhaps I missed some other opportunity that you'll see!

pkg/trtl/metrics.go
```diff
@@ -0,0 +1,141 @@
...
+
+// Gracefully shutdown the Prometheus metrics service
+func (m *MetricsService) Shutdown() error {
+	// Might want to share context from Trtl more globally?
+	ctx := context.Background()
+	if err := m.srv.Shutdown(ctx); err != nil {
+		log.Error().Err(err).Msg("unable to gracefully shutdown prometheus metrics")
+	}
+	return nil
+}
```
rebeccabilbro, line 50: Wasn't sure this was quite what you had in mind for `Serve` and `Shutdown` @bbengfort so I guessed; happy to redo this, just let me know what you were thinking.

pkg/trtl/trtl.go
```diff
@@ -42,6 +43,16 @@ const (
 // to look in that namespace only.
 func (h *TrtlService) Get(ctx context.Context, in *pb.GetRequest) (*pb.GetReply, error) {
 	var err error
+	var ns string
+
+	if in.Namespace == "" {
+		ns = "default"
+	} else {
+		ns = in.Namespace
+	}
```
rebeccabilbro, line 52: Didn't love having to add this to each of the RPC calls but I couldn't think of another way to get you the call data reliably disaggregated by namespaces the way we have the `default` behavior set up now.

pkg/trtl/trtl.go
```diff
@@ -131,13 +169,36 @@ func (h *TrtlService) Put(ctx context.Context, in *pb.PutRequest) (out *pb.PutRe
...
+	pmLatency.WithLabelValues("Put").Observe(latency)
+
+	// Increment prometheus Put counter
+	pmPuts.WithLabelValues(ns).Inc()
+
+	// TODO: prometheus
+	// If in.Options.ReturnMeta is true, we will get metadata from honu
+	// Unfortunately, we currently we don't get tombstone information, but if we did,
+	// we could use it to decrement our `pmTombstone` counter here
+
```
rebeccabilbro, line 183: Put in this TODO but wanted to call it out here -- can you think of any way we can get the tombstones without fiddling with `honu`?

pkg/trtl/trtl.go
```diff
@@ -525,6 +615,11 @@ func (h *TrtlService) Cursor(in *pb.CursorRequest, stream pb.Trtl_CursorServer)
 
 		// Count the number of messages successfully sent
 		nMessages++
+
+		// TODO: Prometheus
+		// If in.Prefix is nil, nMessages will be all the objects in in.Namespace so we
+		// could use this opportunity to update our Prometheus counter if we can find
+		// out how to call something like update on the counter rather than increment
```
rebeccabilbro, line 622: Here's a place where I think we have an opportunity to count objects in the namespace, though prometheus likes to increment, not count. Maybe we can figure out a workaround?

pkg/trtl/server.go
```diff
@@ -117,6 +123,14 @@ func (t *Server) Serve() (err error) {
 		go t.replica.AntiEntropy()
 	}
 
+	// If metrics are enabled, start Prometheus metrics server as separate go routine
+	if t.conf.MetricsEnabled {
+		log.Info().Str("listen", t.conf.MetricsAddr).Msg("trtl metrics server started")
+		go t.metrics.Serve(t.conf.MetricsAddr)
+	} else {
+		log.Info().Str("listen", t.conf.MetricsAddr).Msg("trtl metrics disabled")
```
bbengfort, line 131: Could we make this a `log.Warn()` please?

pkg/trtl/server.go
```diff
@@ -117,6 +123,14 @@ func (t *Server) Serve() (err error) {
 		go t.replica.AntiEntropy()
 	}
 
+	// If metrics are enabled, start Prometheus metrics server as separate go routine
+	if t.conf.MetricsEnabled {
+		log.Info().Str("listen", t.conf.MetricsAddr).Msg("trtl metrics server started")
```
bbengfort, line 128: The `serveMetrics` function also has a `log.Info()` - is this log message redundant? If not, should we make one of them a debug message?

pkg/trtl/server.go
```diff
@@ -162,6 +179,14 @@ func (t *Server) Shutdown() (err error) {
 	return nil
 }
 
+//===========================================================================
+// Initialize Prometheus metrics
+//===========================================================================
+
+func init() {
+	initMetrics()
```
bbengfort, line 187: Should we init the metrics in `NewMetricsService`? I don't think we make any calls to the prometheus metrics until the server is started and running so we could probably wait until the Server is created. If you think it should remain in `init()` - could we move this to the top of this file; Because the `init` runs on import always, I like to have that at top of mind in the code so we don't forget what is being run in `init`.

pkg/trtl/trtl.go
bbengfort: I know what you mean, if we change the name of the "default" namespace, in Honu, this will cause a minor problem that we'll have to deal with. 

One option is to get the namespace from Honu after we perform the db operation. Both `Put` and `Delete` return object meta which has the Namespace on it - so that should be fairly straight forward. In `Get` we're only doing `h.db.Object` if `in.Options.ReturnMeta` is true - but for this option we could always do `h.db.Object` so that we can get access to the meta data and therefore the namespace. We'll incur some small parsing cost from Honu, but it's really pretty insignificant with our recent changes. It will also simplify `Get` which might make it easier to maintain - e.g. you wouldn't need two separate metrics calls like right now.

pkg/trtl/trtl.go
bbengfort: I see what you mean, unfortunately, the only thing I can think of is to do an `h.db.Object` before the `Put` to see if the current version is a Tombstone - however, because we don't have transactions this introduces the possibility of an inconsistency if two concurrent puts see a tombstone and decrement the tombstone counter twice instead of once.

The right way would be to change Honu to have it indicate if the parent version was a Tombstone or not; but that does require a change to Honu. I think this `TODO` is the best we can do for now.

pkg/trtl/trtl.go
bbengfort: Can a prometheus `Gauge` count?

pkg/trtl/metrics.go
bbengfort: Yes, I see exactly what you mean - excellent notes and comments below! The only thing I can think of is a go routine that kicks off every 10 minutes or so and counts all the objects and tombstones in the database. Otherwise there are some changes to Honu we could make to make the metrics happen correctly. Either way, I think this is probably a follow on task we could create a story for?

bbengfort (approved):
Well done - and very well written! I'm definitely glad to have metrics in the GDS system so we can get visibility into what's happening beyond log messages - this is excellent! Thank you also so much for your comments, which helped me review and understand your PR in detail. I've made a couple of minor requests for changes to serve and init - but once those are done I think this is ready to merge at your convenience!

pkg/trtl/trtl.go
```diff
@@ -301,6 +347,7 @@ func (h *TrtlService) Iter(ctx context.Context, in *pb.IterRequest) (out *pb.Ite
 			log.Error().Err(err).Str("key", base64.RawURLEncoding.EncodeToString(iter.Key())).Msg("could not fetch object metadata")
 			return nil, status.Error(codes.FailedPrecondition, "database is in invalid state")
 		}
+		ns = object.Namespace
```
rebeccabilbro, line 350: @bbengfort is there a better place to retrieve Namespace from in Iter?

bbengfort: We added `iter.Namespace()` to access the namespace that is being iterated on and you can use it outside of the `for` loop if that's easier.

PR #271, opened 2021-12-16
Remove sqlite from GDS (sc-186)

rebeccabilbro:
The GDS implementation contained artifacts of an initial plan to provide sqlite support for GDS that is no longer needed. This PR removes those remnants.


pdeziel (approved):
Seems pretty straightforward!

PR #273, opened 2021-12-17
feat: add placeholder for data loading

elysee15:
Fixes SC-2501
I added in this task loader to let the user know that there an ongoing request

https://user-images.githubusercontent.com/54010836/146603493-ae97560b-e989-4b6a-9798-1b52f348f496.mov

### DIFF
--- a/web/gds-admin-ui/package.json
+++ b/web/gds-admin-ui/package.json
@@ -18,6 +18,7 @@
         "jsvectormap": "^1.3.2",
         "jwt-decode": "^3.1.2",
         "node-sass": "^6.0.1",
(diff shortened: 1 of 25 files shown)


bbengfort (approved):
@elysee15 thank you for thoughtfully adding this loader; it's certainly a better user experience to have a loader rather than nothing. While testing, I wasn't able to really see the loader since the responses are so fast; but I'm glad it's there rather than not there. 

If in the future we wanted to switch to a different loader UI, e.g. instead of an oval, a bar across the top of the page, would we still be able to use NProgress?

PR #274, opened 2021-12-20
sc-2345 add CertManager tests

pdeziel:
This adds tests for the CertManager using a method similar to the backup manager tests. This required changing the fixture generator script again, to add new VASP and CertReq states and implement the relation between them so that the CertManager can do its thing. I also added a certificate fixture zip file so that we can actually test the Sectigo certificate download functionality using our previously built mocks.


bbengfort (approved):
Looks great! Thank you so much for catching and working through all of those race conditions, I'm also super happy to have this code tested - it's one of the most complicated parts of the GDS code base and the most prone to failure when Sectigo changes! We can iterate on this code with confidence now!

PR #275, opened 2021-12-21
Quick fixes: testdata and docker compose

bbengfort:
* removes double colons from wire package testdata filenames
* adds GDS members service environment variables to docker compose

TODO: @rebeccabilbro I know you mentioned we should update the `fakerize.py` script and I can see where this is saving files with the `::` prefix; however we're then gzipping those files into a `db.gz` file for the testdata. It appears that this is a Go problem (e.g. `go get` and `go doc`) not a general Linux filesystem problem (e.g. when Go can't see the `::` files in `db.gz` it doesn't have a problem, and the tests have been running just fine extracting those files, reading them and inserting them into the database). Even though it's not a problem do you still want me to change `fakerize.py` and regenerate the test fixtures?

@elysee15 this should fix the issue you were having running GDS. I'm testing Docker compose right now.


rebeccabilbro:
@bbengfort - I was interpreting this change in the test data to be a change not only to some test filenames, but to our fixture naming conventions more globally (namely, don't use `::` in filenames, since they don't appear to be compatible with Go), in which case I think it might be confusing to future contributors to see `fakerize.py` storing files in this same project with `::`. Does that makes sense? We can make this a follow-on task if you prefer.

bbengfort:
It does make sense - I'll make those changes now.

containers/.env.template
```diff
@@ -7,6 +7,7 @@
 SECTIGO_USERNAME=""
 SECTIGO_PASSWORD=""
 SECTIGO_PROFILE="CipherTrace EE"
+SECTIGO_TESTING=true
```
bbengfort, line 10: All of us need to add this to our `containers/.env` file to ensure that when we're using docker compose we're not hitting Sectigo.

containers/docker-compose.yaml
```diff
@@ -31,14 +31,18 @@ services:
       - GDS_ADMIN_ALLOW_ORIGINS=http://localhost:3001,http://localhost:3000
       - GDS_ADMIN_COOKIE_DOMAIN=localhost
       - GDS_ADMIN_AUDIENCE=http://localhost:4433
+      - GDS_MEMBERS_ENABLED=true
+      - GDS_MEMBERS_BIND_ADDR=:4435
+      - GDS_MEMBERS_INSECURE=true
       - GDS_REPLICA_ENABLED=false
-      - GDS_REPLICA_BIND_ADDR=:4435
+      - GDS_REPLICA_BIND_ADDR=:4436
```
bbengfort, line 38: Ports are assigned as follows:

- `:4433`: GDS gRPC Server
- `:4434`: GDS Admin Server
- `:4435`: Members gRPC Server
- `:4436`: Trtl Server

fixtures/datagen/fakerize.py
```diff
@@ -379,11 +379,12 @@ def store(fakes, kind="vasps", directory=OUTPUT_DIRECTORY):
...
     Return the path
     """
+    directory = os.path.join(directory, kind)
     if not os.path.exists(directory):
         os.makedirs(directory)
 
     for idx, fake in fakes.items():
-        fname = os.path.join(directory, kind + "::" + shorten(idx) + ".json")
+        fname = os.path.join(directory, shorten(idx) + ".json")
         with open(fname, "w") as outfile:
```
bbengfort, line 388: Instead of storing the files in `vasps:1234.json` and `certreqs::4568.json` we're using a directory to split the different records up, e.g. `vasps/1234.json` and `certreqs/4568.json`.

rebeccabilbro (approved):
Thank you @bbengfort for making these changes to the fixtures to address the issues we've been encountering with `go doc` and `go get` related to `::` in filenames, as well as refreshing the docker compose to improve the developer experience for our frontend team!

PR #277, opened 2021-12-21
Preliminary test coverage for Members endpoint

rebeccabilbro:
This PR is intended to provide initial coverage for the new experimental Members endpoint. 

In the course of working on this, I've done some light refactoring of the way we were previously setting up the GDS tests -- before we were using the `loadFixtures` method to stop any previously running server, and start running a new server at the beginning of each test function that leveraged the fixtures. Now the startup/shutdown methods are separate from the fixture loading, meaning that they both have to be run to properly set up each test (load fixtures first, then setup the GDS server or Members server as appropriate).


bbengfort (approved):
I wish I understood why shutting down the servers was causing an issue, but I'm glad there was no conflict between `SetupMembers` and `SetupGDS` with the `grpc` buffconn. The Members list test is well written and covers the most important parts of the code and the refactor to the `Setup` server methods is very well written! Thank you for making these changes!

PR #278, opened 2021-12-22
sc-2170 review endpoint accept and reject tests

pdeziel:
This adds tests for the accept and reject paths of the Review endpoint in the Admin API. This required editing the fixtures a bit to add the verification token and a new VASP-cert relationship.

I also encountered a bug where we were not persisting the email logs to the database after sending the rejection emails, so I added an update step after the accept/reject logic in the Review handler.


fixtures/datagen/templates/no_cert.json
```diff
@@ -134,7 +134,7 @@
     },
     "extra": {
         "@type": "type.googleapis.com/gds.models.v1.GDSExtraData",
-        "admin_verification_token": "",
+        "admin_verification_token": "supersecrettoken",
```
bbengfort, line 137: my favorite token :-)

pkg/gds/admin.go
```diff
@@ -1568,6 +1568,12 @@ func (s *Admin) Review(c *gin.Context) {
 		}
 	}
 
+	// Persist the VASP record to the database
```
bbengfort, line 1571: Good catch!

bbengfort (approved):
@pdeziel good tests - thanks for adding the new fixtures and breaking down the test into three separate test methods!

PR #279, opened 2021-12-22
Trtl: Peers-Honu interactions and Refactored Trtl tests

bbengfort:
This PR implements the Peers management server interactions with the Honu database and implements tests for the service. The Trtl tests were also refactored a bit based on our experience with GDS testing. 

Note: this PR is not sc-1623 but it was a necessary prerequisite that we didn't manage to create a story for. I'll continue the branch with sc-1623 after this PR is reviewed.


pkg/trtl/config/config.go
```diff
@@ -61,6 +61,15 @@ func (c Config) IsZero() bool {
...
 }
 
+// Mark a manually constructed as processed as long as it is validated.
+func (c Config) Mark() (Config, error) {
+	if err := c.Validate(); err != nil {
+		return c, err
+	}
+	c.processed = true
+	return c, nil
+}
```
bbengfort, line 71: This code is in `pkg/gds/config` so that tests can be manually created without loading the values from the environment; this is useful for `Trtl` too, so I've added it to `pkg/trtl/config`.

pkg/trtl/peers.go
```diff
@@ -3,43 +3,30 @@ package trtl
...
-
-func (s *notImplementedStore) CreatePeer(peer *peers.Peer) (string, error) {
-	return "", errors.New("not implemented")
-}
-func (s *notImplementedStore) DeletePeer(id string) error   { return errors.New("not implemented") }
-func (s *notImplementedStore) ListPeers() iterator.Iterator { return nil }
-func (s *notImplementedStore) AllPeers() ([]*peers.Peer, error) {
-	return nil, errors.New("not implemented")
-}
-
```
bbengfort, line 31: Removed in favor of actual Honu DB interactions.

pkg/trtl/peers.go
```diff
@@ -53,12 +40,47 @@ func (p *PeerService) GetPeers(ctx context.Context, in *peers.PeersFilter) (out
...
-	// CreatePeer handles possibility of an already-existing or previously deleted peer
-	if _, err := p.store.CreatePeer(in); err != nil {
-		log.Error().Err(err).Msg("unable to add peer")
-		return nil, status.Error(codes.InvalidArgument, "invalid peer; could not be added")
+	// Compute the key from the Peer method to ensure consistent key access.
+	// If the key is empty, that means the peer ID == 0; return an error.
+	key := in.Key()
+	if key == "" {
+		return nil, status.Error(codes.InvalidArgument, "a valid peer ID is required")
+	}
```
bbengfort, line 50: Because the Key is based on the PID, it is impossible to create a Peer with a duplicate PID.

pkg/trtl/peers.go
```diff
@@ -53,12 +40,47 @@ func (p *PeerService) GetPeers(ctx context.Context, in *peers.PeersFilter) (out
...
+
+	// Check if the peer is in the database; if it is unmarshal it and update it
```
bbengfort, line 52: Alternatively we could error here, requiring "editing" to be delete then add again -- this might be a good idea to help us accidentally overwrite a Peer with a duplicate PID.

pkg/trtl/peers.go
```diff
@@ -96,41 +117,57 @@ func (p *PeerService) RmPeers(ctx context.Context, in *peers.Peer) (out *peers.P
...
+		// TODO: why is Honu returning tombstones in iter?
+		var obj *object.Object
+		if obj, err = ps.Object(); err != nil {
+			log.Error().Err(err).Str("key", string(ps.Key())).Msg("could not retrieve object from db")
+			continue
+		}
+
+		if obj.Tombstone() {
+			continue
+		}
```
bbengfort, line 145: We probably need to deal with this in Honu.

pkg/trtl/testdata/db.json
```diff
@@ -94,15 +94,5 @@
...
-    },
-    "peer": {
-        "namespace": "peers",
-        "key": "82abc34d-f8f9-4f0e-b8e3-f8f9f8f8f8f8",
-        "value": {
-            "Id": 1,
-            "Addr": "127.0.0.1:1234",
-            "Name": "peer",
-            "Region": "us-east-1"
-        }
```
bbengfort, line 106: The peer needs to be in protocol buffer wire format, so we can't add it as JSON to the fixtures unfortunately.

pkg/trtl/peers/v1/generate.go
```diff
@@ -1,3 +1,3 @@
 package peers
 
-//go:generate protoc -I=../../../../proto --go_out=. --go_opt=module=github.com/trisacrypto/directory/pkg/trtl/peers/v1 --go-grpc_out=. --go-grpc_opt=module=github.com/trisacrypto/directory/pkg/trtl/peers/v1 gds/peers/v1/peers.proto
+//go:generate protoc -I=../../../../proto --go_out=. --go_opt=module=github.com/trisacrypto/directory/pkg/trtl/peers/v1 --go-grpc_out=. --go-grpc_opt=module=github.com/trisacrypto/directory/pkg/trtl/peers/v1 trtl/peers/v1/peers.proto
```
bbengfort, line 3: I moved the peers from `proto/gds/peers` to `proto/trtl/peers`.

pkg/trtl/peers.go
```diff
@@ -53,12 +40,47 @@ func (p *PeerService) GetPeers(ctx context.Context, in *peers.PeersFilter) (out
...
+
+	}
+
+	// TODO: validate other Peer fields
+	in.Modified = time.Now().Format(time.RFC3339)
+
+	// Insert the peer into the database
+	var value []byte
+	if value, err = proto.Marshal(in); err != nil {
+		log.Error().Err(err).Msg("could not marshal peers protocol buffers")
```
pdeziel, line 77:
```suggestion
		log.Error().Err(err).Msg("could not marshal peer protocol buffers")
```

pkg/trtl/peers.go
```diff
@@ -96,41 +117,57 @@ func (p *PeerService) RmPeers(ctx context.Context, in *peers.Peer) (out *peers.P
...
+		out.Status.NetworkSize++
+		out.Status.Regions[peer.Region]++
+
+		// If it's not a status only, get the details for each Peer
+		if !in.StatusOnly {
+			// If we've been asked to filter by region
+			if len(in.Region) > 0 {
+				for _, region := range in.Region {
+					if peer.Region == region {
+						out.Peers = append(out.Peers, peer)
```
pdeziel, line 162: Do we need to break out of the loop here? If the request contains duplicate regions then we could end up returning the same peer more than once.

pkg/trtl/peers.go
pdeziel: Good point, I'm not sure which behavior would be preferred here. I guess it depends on if we are expecting the caller to know whether or not the peer already exists.

pkg/trtl/peers_test.go
```diff
@@ -0,0 +1,131 @@
...
+			Id:     3,
+			Addr:   "ei.trtl.dev:443",
+			Name:   "charlie",
+			Region: "epsiloniridini",
+		},
+		{
+			Id:     4,
+			Addr:   "wolf.trtl.dev:443",
+			Name:   "delta",
+			Region: "wolf579",
```
pdeziel, line 53: Not going with the star trek [reference](https://memory-alpha.fandom.com/wiki/Battle_of_Wolf_359)?

pkg/trtl/server_test.go
```diff
@@ -124,73 +122,242 @@ func (s *trtlTestSuite) StatusError(err error, code codes.Code, theError string)
 	require.Equal(theError, serr.Message(), "status error message does not match")
 }
 
-func (s *trtlTestSuite) SetupSuite() {
-	var err error
+// EqualMeta is a helper assertion function that checks if the actual metadata matches
+// expectations about what the version should be. This helper function relies on some
+// test fixture information to minimize what test users must supply.
+func (s *trtlTestSuite) EqualMeta(expectedKey []byte, expectedNamespace string, expectedVersion, expectedParent *pb.Version, actualMeta *pb.Meta) {
```
pdeziel, line 128: Thanks for adding this, it will be so much easier to debug test failures!

pdeziel (approved):
I had a few questions and suggested changes but looks just about ready to me, thanks for figuring out the specifics of the Trtl test suite fixtures handling!

pkg/trtl/namespaces.go
```diff
@@ -1,10 +1,17 @@
...
 var reservedNamespaces = map[string]struct{}{
-	"peers":    {},
-	"sequence": {},
-	"index":    {},
-	"default":  {}, // if the user does not specify a namespace
+	NamespacePeers:    {},
+	NamespaceSequence: {},
+	NamespaceIndex:    {},
+	NamespaceDefault:  {}, // if the user does not specify a namespace
 }
```
rebeccabilbro, line 17: Was this change related to the tests?

pkg/trtl/server_test.go
```diff
@@ -58,60 +54,62 @@ type dbEntry struct {
...
-	grpc *bufconn.GRPCListener
+	trtl  *trtl.Server
+	conf  *config.Config
+	grpc  *bufconn.GRPCListener
+	tmpdb string
 }
 
-// loads the test fixtures from a JSON file and stores them in the dbFixtures map
-func (s *trtlTestSuite) loadFixtures() {
+func (s *trtlTestSuite) SetupSuite() {
```
rebeccabilbro, line 69: Thank you for making this pattern more analogous to what we're doing in the GDS to help us manage cognitive load across subpackages!

rebeccabilbro (approved):
Whew, there is definitely a lot going on in this PR, but everything looks good @bbengfort. I like the refactoring of the trtl tests and the addition of new tooling to help compare protocol buffers -- might be something we want to break out eventually into an `x` type package.

pkg/trtl/namespaces.go
bbengfort: I thought it would be easier to manage the code if we had these strings as constants.

pkg/trtl/peers_test.go
bbengfort: Whoops -- that's actually what I was going for; I'll update it to the correct wolf!

pkg/trtl/peers.go
bbengfort: Ok, I guess we should continue discussing this in the future?

pkg/trtl/peers.go
bbengfort: Good point, I'll make that change.

PR #280, opened 2021-12-24
Trtl: Anti-Entropy and Gossip

bbengfort:
This PR implements the `ReplicaServer` including the `AntiEntropy` routine and `Gossip` Service.

Blocked by https://github.com/rotationalio/honu/pull/15


go.mod
```diff
@@ -23,9 +23,9 @@ require (
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/klauspost/compress v1.13.6 // indirect
 	github.com/kr/pretty v0.3.0 // indirect
-	github.com/prometheus/client_golang v1.11.0 // indirect
-	github.com/rogpeppe/go-internal v1.8.0 // indirect
-	github.com/rotationalio/honu v0.2.5
+	github.com/prometheus/client_golang v1.11.0
+	github.com/rogpeppe/go-internal v1.8.1 // indirect
+	github.com/rotationalio/honu v0.2.6-0.20211224014939-17859381b1e2
```
bbengfort, line 28: This will have to be updated to the next honu version before merge.

pkg/trtl/replica.go
```diff
@@ -85,13 +411,264 @@ func (r *ReplicaService) SelectPeer() (peer *peers.Peer) {
...
+					sync.Object = local
+					sync.Error = ""
+					if err = stream.Send(sync); err != nil {
+						logctx.Error().Err(err).Msg("could not send gossip message, prematurely quitting phase 1")
+						return
+					}
+					atomic.AddUint64(&updates, 1)
+				case sync.Object.Version.IsLater(local.Version):
+					// Send a check request back to the initiating replica to fetch its version
+					if err = stream.Send(sync); err != nil {
```
bbengfort, line 608: this needs to be modified to local.Version

pkg/trtl/replica.go
```diff
@@ -85,13 +419,264 @@ func (r *ReplicaService) SelectPeer() (peer *peers.Peer) {
...
-// use a partial flag to indicate that it is requesting specific versions. This
-// mechanism implements bilateral anti-entropy: a push and pull gossip.
-func (r *ReplicaService) Gossip(ctx context.Context, in *replica.VersionVectors) (out *replica.Updates, err error) {
-	return nil, status.Error(codes.Unimplemented, "this replica does not yet implement gossip")
+// Gossip implements biltateral anti-entropy: during a Gossip session the initiating
+// replica pushes updates to the remote peer and pulls requested changes. Using
+// bidirectional streaming, the initiating peer sends data-less sync messages with
+// the versions of objects it stores locally. The remote replica then responds with
+// data if its local version is later or sends a sync message back requesting the
+// data from the initating replica if its local version is earlier (no exchange)
```
pdeziel, line 427:
```suggestion
// data from the initiating replica if its local version is earlier. No exchange
```

pkg/trtl/replica.go
```diff
@@ -85,13 +419,264 @@ func (r *ReplicaService) SelectPeer() (peer *peers.Peer) {
...
+// occurs if both replicas have the same version. At the end of a gossip session,
+// both replicas should have synchronized and have identical underlying data stores.
+func (r *ReplicaService) Gossip(stream replica.Replication_GossipServer) (err error) {
```
pdeziel, line 430: Is there some way we can better clarify that the `AntiEntropySync` is the "client" method and the `Gossip` is the "server" method in this interaction? Perhaps they should be in separate files? The way I'm thinking about this is that there is one set of things that the initiator does and another set of things that the remote does and separating those two roles makes sense to me.

pdeziel (approved):
I made another pass through this, thanks for your patience! I did have some small suggested docstring changes and one potential refactoring change that I'd like to get your opinion on.

pkg/trtl/replica.go
```diff
@@ -2,39 +2,365 @@ package trtl
...
+// AntiEntropySync performs bilateral anti-entropy with the specified remote peer using
+// the streaming Gossip RPC. This method initiates the Gossip stream with the remote
+// peer, exiting if it cannot connect to the replica. In the pull phase, this method
+// sends check sync messages for all objects stored locally; the remote replica responds
+// with repairs. Then in the push phase, the method waits until all requested remote
+// repairs are complete before exiting.
+func (r *ReplicaService) AntiEntropySync(peer *peers.Peer, log zerolog.Logger) (err error) {
+	// Create a context with a timeout of the gossip interval, so that anti-entropy
+	// gossip sessions do not span multiple anti-entropy intervals.
+	ctx, cancel := context.WithTimeout(context.Background(), r.conf.GossipInterval)
```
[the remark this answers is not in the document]
bbengfort: That's an excellent question - if we subtracted `r.conf.GossipInterval - r.conf.GossipSigma` we'd capture 68% of delay introduced by jitter; if  we subtracted `r.conf.GossipInterval - 2*r.conf.GossipSigma` we'd be ahead of 95% of the delays chosen introduced by jitter. I think it's worth doing, so I'll put it in - thanks for the suggestion!

pkg/trtl/replica.go
```diff
@@ -2,39 +2,365 @@ package trtl
...
+	// Kick off two go routines - one to send messages on and one to recv them on, this
+	// is necessary because gRPC send and recv are blocking operations that can only be
+	// used by one go routine, e.g. they are not thread safe. These go routines place
+	// channels between the send and recv operations so that multiple go routines can
+	// send and recv concurrently.
+	// TODO: standardize the functionality in a utility package.
+	var stop uint32
+	var wg sync.WaitGroup
+	send := make(chan *replica.Sync, 8)
+	recv := make(chan *replica.Sync, 8)
```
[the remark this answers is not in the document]
bbengfort: On the 9th message the go routine will block until there is space available on the channel; this creates back pressure to ensure that our various go routines across multiple processes can't get too far ahead or behind of each other.

pkg/trtl/replica.go
```diff
@@ -2,39 +2,365 @@ package trtl
...
+		for msg := range msgs {
+			log.Trace().Str("status", msg.Status.String()).Msg("sending sync")
+			if err := stream.Send(msg); err != nil {
+				if err != io.EOF {
+					log.Error().Err(err).Msg("could not send gossip message to remote peer")
+				}
+
+				// Let external go routines know that they can no longer send on this channel.
+				atomic.AddUint32(&stop, 1)
+				return
```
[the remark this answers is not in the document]
bbengfort: here is the fixed code for reference:

```go
		for msg := range msgs {
			log.Trace().Str("status", msg.Status.String()).Msg("sending sync")
			if err := stream.Send(msg); err != nil {
				if err != io.EOF {
					log.Error().Err(err).Msg("could not send gossip message to remote peer")
				}

				// Let external go routines know that they can no longer send on this channel.
				atomic.AddUint32(&stop, 1)
				break
			}
		}

		// Continue to consume messages until the msgs channel is closed to ensure that
		// go routines that do not check if they can send messages don't get blocked if
		// the message channel is filled up. This won't do anything with the messages.
		// NOTE: ranging over a closed channel immediately returns so even if the msgs
		// channel was closed in the first loop, this second loop will not execute nor
		// will it panic.
		for msg := range msgs {
			log.Trace().Str("status", msg.Status.String()).Msg("dropped sync message (not sent)")
		}
```

pkg/trtl/replica.go
bbengfort: I'll think about it in sc-2694 and add your comment there.

PR #287, opened 2022-01-10
Trtl simulator

rebeccabilbro:
This PR solves sc-2367, creating a simulator to allow us to understand Trtl's read/write/delete behavior independently of GDS.  

Note: This is not necessarily intended for merging


cmd/trtlsim/main.go
```diff
@@ -0,0 +1,174 @@
...
+	regions   = 7                      // number of regions simultaneously using the accessor
+	endpoint  = "localhost:831"        // the endpoint of the running trtl server
+	insecure  = true                   // connect without mTLS
+	certs     = ""                     // path to file on disk with certificates if using mTLS
+	keyspace  = 1000                   // the number of keys the simulator operates on
+	chunkSize = 512                    // size of each write
+)
+
+// dummy namespaces for simulation
+var namespaces = []string{"catchfireBarons", "falselightCutters", "fullCrowns", ""}
```
bbengfort, line 31: :-)

cmd/trtlsim/main.go
```diff
@@ -0,0 +1,174 @@
...
+
+// probabilities for reads/writes/deletes
+var probabilities = map[string]uint{
+	"read":   60,
+	"write":  38,
+	"delete": 20,
```
bbengfort, line 37: Should the delete weight be 2?

cmd/trtlsim/main.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+		opts = append(opts, grpc.WithTransportCredentials(credentials.NewTLS(conf)))
+	}
+
+	// Connect the replica client
+	var cc *grpc.ClientConn
+	if cc, err = grpc.Dial(s.Endpoint, opts...); err != nil {
+		return nil, err
+	}
+	return pb.NewTrtlClient(cc), nil
```
bbengfort, line 137: To figure out what problems we might be having in the deployment, could we add a log statement before we return the client, e.g. something like `log.Printf("connected to trtl server at %s\n", s.Endpoint)`. If we've gotten the configuration wrong, this will help us quickly understand what's going on from the k8s output.

cmd/trtlsim/main.go
```diff
@@ -0,0 +1,238 @@
...
+					if serr, ok := status.FromError(err); ok && serr.Code() == codes.NotFound {
+						continue
+					}
+					panic(fmt.Errorf("could not delete from database: %v", err))
+				}
+			}
+		default:
+			panic(errors.New("unknown database operation"))
+		}
+	}
```
bbengfort, line 220: At the end of this function, just for debugging and analysis would it be possible to print the number of gets, puts, and deletes that the accessor just performed using `log.Println`? That will help us see that the sim is running when we deploy it and will give us a sense of what accesses happened in the last round.

bbengfort (approved):
@rebeccabilbro it looks very good! I've verified it with the following actions:

- built and ran the dockerfile 
- run the trtl server locally and ran the simulation in another terminal window
- checked the trtl database - plenty of Camorri gangs in there! 

I have a couple of recommendations below, but these recommendations are just about making devops a bit easier on Friday and managing the simulation while it's running. Thank you so much for putting this together!

PR #295, opened 2022-01-14
BugFix: Common Name Validation

bbengfort:
It appears that the semantics of net.SplitHostPort have changed and now
if a port is not included in an addr, then an error is returned. This
prevents us from using this method for common name validation and the
errors have started showing up in the UI. The new method of validation
uses a regular expression, which isn't great, but it does do the job for
all of the cases provided in the tests.

NOTE: this must be included in the GDS v1.2 release.

### DIFF
--- a/pkg/gds/gds.go
+++ b/pkg/gds/gds.go
@@ -5,7 +5,9 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"regexp"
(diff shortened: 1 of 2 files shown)


rebeccabilbro (approved):
Looks good to me @bbengfort - this happened because of a change in the ‘net’ package?

PR #296, opened 2022-01-14
Test business infos modal

elysee15:
Fixes SC-2806


bbengfort (approved):
Excellent, thank you for adding these tests! The more test coverage in the front end, the better. So anywhere you could create stories for tests I'd appreciate it!

PR #298, opened 2022-01-15
Replication Status Helpers

bbengfort:
The original intent for this story was to add some metrics and add a `Status` endpoint so that we could both track anti-entropy as it was running and prepare for reinforcement learning with Honu. However, as @rebeccabilbro pointed out, we might be able to simply use prometheus metrics for this. 

Therefore, I've changed the intent of this story a little bit - writing a few tools and helpers for us to better observe what is happening while our anti-entropy experiments are running. I've also taken care of a couple of `TODOs` in the comments since I was there. 

This PR adds:

1. A `Status` endpoint (and corresponding CLI command) that will return ok, the version, and the uptime of the server so that we can quickly check if we can communicate with the server. This endpoint will run even if we're in maintenance mode.
2. A `last_synchronized` timestamp added to the `PeersStatus` so we can make sure that anti-entropy is occurring. This timestamp is only incremented if data was exchanged between the replicas.
3. A `db:list` CLI command so that we can get all the keys from a namespace from the server, e.g. to see if a Replica has had objects synchronized from it.


cmd/trtl/main.go
```diff
@@ -730,19 +813,26 @@ func loadProfile(c *cli.Context) (err error) {
...
+			UseEnumNumbers:  false,
+			EmitUnpopulated: true,
+		}
+		if data, err = opts.Marshal(msg); err != nil {
+			return cli.Exit(err, 1)
+		}
+	default:
+		if data, err = json.MarshalIndent(m, "", "  "); err != nil {
+			return cli.Exit(err, 1)
+		}
```
pdeziel, line 835: I like that this is abstracted out to its own function now.

pkg/trtl/interceptor.go
```diff
@@ -32,27 +32,36 @@ func (t *Server) interceptor(ctx context.Context, in interface{}, info *grpc.Una
...
-		log.Error().Err(err).Msg("unable to retrieve remote peer info")
-		// TODO: Uncomment when mTLS is added
-		//return nil, status.Error(codes.Unauthenticated, "unable to retrieve authenticated peer information")
-	}
+	// Fetch peer information from the TLS info if we're not in insecure mode.
+	if !t.conf.MTLS.Insecure {
+		var peer *PeerInfo
+		if peer, err = peerFromTLS(ctx); err != nil {
+			return nil, status.Error(codes.Unauthenticated, "unable to retrieve authenticated peer information")
+		}
```
pdeziel, line 50: Thanks for addressing this, I suppose we should create a test for `peerFromTLS` at some point.

pdeziel (approved):
Looks good to me

PR #299, opened 2022-01-15
Refactor Phase 1/Phase 2 Anti-Entropy and Gossip

bbengfort:
This PR refactors anti-entropy (and hopefully corrects some of the bugs). It is broken into several commits to show the restructuring.

The first commit moves the `trtl.ReplicaService` into its own package, `pkg/trtl/replica` so that the new struct is `replica.Service`. @pdeziel suggested multiple files to break down and organize some of the code, and decoupling replication into its own package allowed us to do this without cluttering the `trtl` package. It also gives us the opportunity to create a replication-specific go doc (SC-2697). I did my best to follow the conventions we're establishing in `trtl` - but there is a little bit of weirdness, such as what has to be passed to `replica.New` in order to allow the decoupling to happen. 

The second commit refactors the `AntiEntropySync` method, breaking out the `initiatorPhase1` and `initiatorPhase2` go routines for better readability. It also creates a data structure for the `streamSender` since both `AntiEntropySync` and `Gossip` will require this functionality. To make things more understandable, I've removed the unnecessary stream recv go routine, since `initiatorPhase2` is the only go routine that receives messages. I've also added mTLS configuration (though this doesn't appear to be working yet) in a `connect` method. This update eliminates many of the bugs we saw with the first implementation; including not sending all messages (the `AntiEntropySync` method blocks until all messages are sent), the `context canceled` error (also fixed by blocking until the sender completes), and the could not select remote peer error - by copying the key `[]byte` into our keys array so that the pointer doesn't change in the `for` loop. I tested this locally and everything is working smoothly, but still more work to be done! 

The third commit pulls in [honu PR #18](https://github.com/rotationalio/honu/pull/18). The refactoring required a change to `honu.Update` so it checks the version before writing it, otherwise we will introduce the possibility of an inconsistency where a concurrent sync or Put changes the local version while anti-entropy is occurring. 

Blocked by #298 (SC-952)


pkg/trtl/config/config.go
```diff
@@ -180,3 +169,48 @@ func (c MTLSConfig) ParseTLSConfig() (_ *tls.Config, err error) {
...
+func (c MTLSConfig) GetCert() (_ tls.Certificate, err error) {
+	if len(c.cert.Certificate) == 0 {
+		if err = c.load(); err != nil {
+			return c.cert, err
+		}
+	}
+	return c.cert, nil
+}
+
+func (c MTLSConfig) load() (err error) {
```
bbengfort, line 191: grrr …. debugging for an hour and it turns out this is a pointer error … the linter even saw it below … 🤦🏽

pkg/trtl/replica/sync.go
```diff
@@ -0,0 +1,491 @@
...
+	go r.initiatorPhase1(ctx, wg, log, sender)
+
+	// Start phase 2: this phase is concurrent with phase 1 since it listens for and
+	// responds to all messages from the remote replica. This is also called the "push"
+	// phase, since we're pushing objects back to the remote replica. This phase ends
+	// when the remote replica sends a COMPLETE message, meaning it is done sending
+	// messages. At that point, we will no longer send any messages so this phase will
+	// close the sender go routine, which will stop when all messages have been sent.
+	wg.Add(1)
+	go r.initiatorPhase2(ctx, wg, log, sender, stream)
```
rebeccabilbro, line 216: Interesting! Possibly I had this backwards before. I was under the impression that phase one was called the push phase because the initiator is pushing (first versions, then data) to the remote. Push is all that happens in the non bilateral AE case, and the remote is the only replica that makes changes in that case. Bilateral AE introduces phase two, the pull phase, where the initiator has the opportunity to receive repairs from the remote. 

If I do have that backwards, please let me know and I'll fix the [anti-entropy docs](https://docs.google.com/presentation/d/1HhP5PwV5vyqU93SQsLx3IwD_5t-vPZFxYF8kTIiMfEc/edit?usp=sharing) I wrote.

pkg/trtl/replica/sync.go
```diff
@@ -0,0 +1,491 @@
...
+		if sync, err = stream.Recv(); err != nil {
+			if err != io.EOF {
+				// If the error is not EOF then something has gone wrong.
+				log.Error().Err(err).Msg("anti-entropy aborted early with recv error")
+			}
+			return
+		}
+
+		switch sync.Status {
+		case replica.Sync_CHECK:
```
rebeccabilbro, line 406: Is this the mirror CHECK mentioned in Gossip?

pkg/trtl/replica/sync.go
```diff
@@ -0,0 +1,491 @@
...
+			}
+
+		case replica.Sync_REPAIR:
+			// Receiving a repaired object, check if it's still later than our local
+			// replica's and if so, save it to disk with an update instead of a put.
+			//
+			// NOTE: we must check if it's later than our local version in case a
+			// concurrent Put (a stomp) or concurrent synchronization updated it.
+			// NOTE: honu.Update performs the version checking in a transaction.
+			if err = r.db.Update(sync.Object, options.WithNamespace(sync.Object.Namespace)); err != nil {
```
rebeccabilbro, line 447: Is this the right place to track a skip or stomp?

pkg/trtl/replica/replica.go
```diff
@@ -0,0 +1,396 @@
...
+			}
+
+		case replica.Sync_REPAIR:
+			// Receiving a repaired object, check if it's still later than our local
+			// replica's and if so, save it to disk with an update instead of a put.
+			//
+			// NOTE: we must check if it's later than our local version in case a
+			// concurrent Put (a stomp) or concurrent synchronization updated it.
+			// NOTE: honu.Update performs the version checking in a transaction.
+			if err := r.db.Update(sync.Object, options.WithNamespace(sync.Object.Namespace)); err != nil {
```
rebeccabilbro, line 257: Looks like this is a place where we'd increment skips or stomps? I guess we'll have to change honu to let us know if there was a PID tiebreaker (stomp) or not (skip)?

pkg/trtl/server_test.go
```diff
@@ -196,12 +195,10 @@ func (s *trtlTestSuite) setupConfig() (err error) {
 			ReindexOnBoot: false,
 		},
 		Replica: config.ReplicaConfig{
-			Enabled:        true,
-			PID:            metaPID,
-			Region:         metaRegion,
-			Name:           metaOwner,
-			GossipInterval: 1 * time.Second,
-			GossipSigma:    100 * time.Millisecond,
```
rebeccabilbro, line 204: Removed because we just want to use the default values here?

rebeccabilbro (approved):
This looks great @bbengfort! I really like the logic factored out into different files, specifically to help distinguish between client/initiator methods and server/remote methods. I also really appreciate the inline comments and helper function names, which make it really convenient to follow along with the bigger processes. 

I have left a handful of notes and ideas based partly on our discussion yesterday (and that weird version-match/value-mismatch bug we noticed) and partly on some possible remaining misconceptions I had about how some terms are used in the AE literature. Thank you in advance for any additional clarification you can provide on those, and for making such a complex process so accessible!

pkg/trtl/replica/replica.go
```diff
@@ -0,0 +1,396 @@
...
+	// messages concurrently without violating the grpc semantic that only one go
+	// routine can send messages on the stream.
+	// NOTE: newStreamSender calls wg.Add(1) and runs the sender go routine.
+	wg := new(sync.WaitGroup)
+	sender := newStreamSender(wg, logctx, stream)
+
+	// Start phase 1: receive object version vectors from the initiator. This go routine
+	// will kick off phase 2: sending unchecked versions back to the initiator.
+	wg.Add(1)
+	go r.remotePhase1(stream.Context(), wg, logctx, stream, sender)
```
pdeziel, line 103: I wish we had better names for phase1 and phase2. I feel like we have a mutual understanding of what phase1 and phase2 mean after our discussions but it might be difficult for others to understand without the accompanying diagram. Unfortunately, I don't have any better suggestions right now.

pkg/trtl/replica/replica.go
```diff
@@ -0,0 +1,396 @@
...
+		case replica.Sync_ERROR:
+			// Something went wrong on the initiator, log and continue
+			if sync.Object != nil {
+				log.Warn().Str("error", sync.Error).
+					Str("key", b64e(sync.Object.Key)).
+					Str("namespace", sync.Object.Namespace).
+					Msg("a replication error occurred")
+			} else {
+				log.Warn().Str("error", sync.Error).Msg("a replication error occurred")
+			}
```
pdeziel, line 275: Why do we continue if there is an error?

pkg/trtl/replica/sync.go
```diff
@@ -0,0 +1,491 @@
...
+	}
+	defer iter.Release()
+
+	for iter.Next() {
+		// Fetch the key and copy it into a new byte slice; otherwise on the next
+		// iteration the pointer to the key byte slice will reference the next key.
+		src := iter.Key()
+		dst := make([]byte, len(src))
+		copy(dst, src)
+		keys = append(keys, dst)
```
pdeziel, line 109: Is this the "duplicate" key fix?

pdeziel (approved):
I think this is a very good improvement over the previous iteration in terms of understandability, I appreciate the thorough docstrings and splitting it up by role (initiator, remote) makes sense to me since it's somewhat analogous to client/server interactions. I also like that we are passing along the log contexts, which seems like it will be helpful for tracking down the bugs.

pkg/trtl/peers_test.go
```diff
@@ -122,7 +122,6 @@ func (s *trtlTestSuite) TestPeers() {
 	require.Equal(network[3].Addr, rep.Peers[0].Addr)
 	require.Equal(network[3].Name, rep.Peers[0].Name)
 	require.Equal(network[3].Region, rep.Peers[0].Region)
-	require.NotEqual(rep.Peers[0].Created, rep.Peers[0].Modified)
```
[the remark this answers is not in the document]
bbengfort: The problem was that tests executed so fast that the created and modified timestamps were less than a second apart, which meant the test could fail occasionally since the timestamp is only to the second granularity.

pkg/trtl/replica/replica.go
```diff
@@ -0,0 +1,396 @@
...
+	return nil
+}
+
+// remotePhase1 is the counterpart to initiatorPhase2 and runs as long as messages will
+// be coming from the initiator. This is the only remote go routine that will receive
+// messages, so an intermediate read routine is not necessary. This phase has two parts
+// (so could be considered both phase 1 and phase 3). The difference is that the phase
+// is allowed to send messages in the first part (phase 1) but cannot send messages in
+// the second part (phase 3). In phase 1, CHECK messages from the initiator are
+// responded to with either a mirror CHECK if the initiator version is later, a REPAIR
```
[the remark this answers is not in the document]
bbengfort: It's more that the CHECK is reflected back to the client if the remote's version is earlier than the initiator's and it's requesting a repair. E.g. a mirror check is the following sequence:

initiator: CHECK v(8,4)
remote: I'm at v(8,3) ⟶ CHECK v(8,4)
initiator: REPAIR v(8,4)

pkg/trtl/replica/replica.go
```diff
@@ -0,0 +1,396 @@
...
+				sync.Object = local
+				sync.Error = ""
+				if ok := sender.Send(sync); ok {
+					atomic.AddUint64(&updates, 1)
+				}
+			case sync.Object.Version.IsLater(local.Version):
+				// Send a check request back to the initiating replica to fetch its version
+				sender.Send(sync)
+			default:
+				// The versions are equal, do nothing
```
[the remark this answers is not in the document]
bbengfort: That would be an unexpected and bad situation that would indicate there was a bug in the versioning system or in our deployment. E.g. this situation could occur if we accidentally assigned the same PID to two different replicas. If the versions are equal, that should imply that the values are equal because a new version is assigned every time the value changes.

pkg/trtl/replica/sync.go
bbengfort: I think you have it right. Just to clarify; I believe that push and pull refer to types of gossip where the synchronization is unidirectional:

1. Pull: initiator is synchronized with remote
2. Push: remote is synchronized by the initiator 
3. Bilateral: both initiator and remote are synchronized

pkg/trtl/replica/sync.go
bbengfort: Yes!

pkg/trtl/replica/sync.go
bbengfort: It is - and now that we have the update type functionality, I've added a TODO here for that tracking.

pkg/trtl/server_test.go
bbengfort: Yes, we're no longer testing the replica in `trtl` - the tests are in `replica`, so this configuration was unnecessary. Removed it for clarity.

pkg/trtl/replica/sync.go
bbengfort: It is!

pkg/trtl/replica/replica.go
bbengfort: This particular error is related to a single object - so if the object has been corrupted, an admin may be able to fix it, but we want to continue with the rest of the synchronization so that all other objects are replicated. If this was a critical error, then instead of sending a `Sync_ERROR`, the initiator would simply close the stream with a gRPC error.

PR #300, opened 2022-01-17
sc-2870 add DeleteVASP endpoint

pdeziel:
This adds the DeleteVASP endpoint to delete a VASP and its associated certificate requests if it is in the correct state. The correct state is defined as `PENDING_REVIEW` or earlier, or `ERRORED`. I thought it made sense to allow deleting errored VASPs given that we intend to use this API for cleaning up bad registrations and such and it might help us avoid editing the database manually.

### DIFF
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -175,6 +175,7 @@ func (s *Admin) setupRoutes() (err error) {
 			vasps.GET("", s.ListVASPs)
 			vasps.GET("/:vaspID", s.RetrieveVASP)
 			vasps.PATCH("/:vaspID", csrf, s.UpdateVASP)
+			vasps.DELETE("/:vaspID", csrf, s.DeleteVASP)
 			vasps.GET("/:vaspID/review", s.ReviewToken)
 			vasps.POST("/:vaspID/review", csrf, s.Review)
 			vasps.POST("/:vaspID/resend", csrf, s.Resend)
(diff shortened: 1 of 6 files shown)


pkg/gds/admin.go
```diff
@@ -1226,6 +1227,58 @@ func (s *Admin) updateVASPEndpoint(vasp *pb.VASP, commonName, endpoint, source s
...
+	}
+
+	// Delete the certificate requests
+	for _, id := range certReqIDs {
+		if err = s.db.DeleteCertReq(id); err != nil {
+			log.Warn().Err(err).Str("certreq_id", id).Msg("could not delete certificate request")
+			c.JSON(http.StatusInternalServerError, admin.ErrorResponse("could not delete associated certificate request"))
+			return
+		}
+	}
```
pdeziel, line 1270: I thought it made sense to delete the certreqs first, because if we somehow fail to delete one of them then we still have a way to remove the VASP from the API.

pkg/gds/admin.go
```diff
@@ -1226,6 +1227,58 @@ func (s *Admin) updateVASPEndpoint(vasp *pb.VASP, commonName, endpoint, source s
...
+	if vasp, err = s.db.RetrieveVASP(vaspID); err != nil {
+		log.Warn().Err(err).Msg("could not retrieve VASP from database")
+		c.JSON(http.StatusNotFound, admin.ErrorResponse("could not retrieve VASP record by ID"))
+		return
+	}
+
+	// Only allow deletions if the VASP has not been reviewed yet
+	if vasp.VerificationStatus > pb.VerificationState_PENDING_REVIEW && vasp.VerificationStatus < pb.VerificationState_ERRORED {
+		log.Warn().Str("status", vasp.VerificationStatus.String()).Msg("VASP is in invalid state for deletion")
+		c.JSON(http.StatusBadRequest, admin.ErrorResponse("cannot delete VASP in its current state"))
```
bbengfort, line 1252: `http.StatusBadRequest` is a good choice for this error code.

pkg/gds/admin.go
bbengfort: Makes sense to me!

pkg/gds/admin_test.go
```diff
@@ -135,6 +135,8 @@ func (s *gdsTestSuite) TestMiddleware() {
 		{"reviews", http.MethodGet, "/v2/reviews", true, false},
 		{"listVASPs", http.MethodGet, "/v2/vasps", true, false},
 		{"retrieveVASP", http.MethodGet, "/v2/vasps/42", true, false},
+		{"updateVASP", http.MethodPatch, "/v2/vasps/42", true, true},
```
bbengfort, line 138: Good catch!

pkg/gds/admin_test.go
```diff
@@ -694,6 +694,61 @@ func (s *gdsTestSuite) TestUpdateVASP() {
...
+		},
+		claims: &tokens.Claims{
+			Email: "admin@example.com",
+		},
+	}
+	c, w := s.makeRequest(request)
+	rep := s.doRequest(a.DeleteVASP, c, w, nil)
+	s.APIError(http.StatusNotFound, "could not retrieve VASP record by ID", rep)
+
+	// VASP is in an invalid state for deletion
```
bbengfort, line 724: Could we add a check to assert that delta is in the state we expect it to be? This is helpful to detect when fixtures change, but also tells us what state delta is in. 

Do you think we should test all states? We want to make really sure we do the right thing for deletes. I know that's excessive, but we are using `>` and `<` comparisons so maybe it's a good guard?

bbengfort (approved):
Looks good, a very complete implementation of the endpoint; I think we've got a pretty good structure set up for the Admin API! I just had one minor note, feel free to merge when ready.

pkg/gds/admin_test.go
pdeziel: Good point, I've pushed some changes to test all the states, mind taking another look?

PR #301, opened 2022-01-18
 update gds admin ui readme file

elysee15:
Fixes SC-2921


pdeziel (changes requested):
Thanks for making this change! While we're at it can we also fix the `TESTNET` misspelling that I mentioned in the story?

elysee15:
> Thanks for making this change! While we're at it can we also fix the `TESTNET` misspelling that I mentioned in the story?

I fixed it in #302

bbengfort:
@elysee15 does this PR need o be rebased or closed? It seems like the only change is whitespace.

elysee15:
> @elysee15 does this PR need o be rebased or closed? It seems like the only change is whitespace.

PR #303, opened 2022-01-18
sc-2606 add Sectigo params to CertificateRequest

pdeziel:
This adds the Sectigo params as a generic map to the CertificateRequest object. We were previously hardcoding the organization and locality values; this PR adds a function which populates these values from the VASP IVMS data upon VASP registration. The default values are still used if the information is not found in the VASP object.


pkg/gds/models/v1/models.go
```diff
@@ -173,6 +176,69 @@ func AppendCertReqID(vasp *pb.VASP, certreqID string) (err error) {
...
+	if len(vasp.Entity.GeographicAddresses) > 0 {
+		address := vasp.Entity.GeographicAddresses[0]
+		localityName = address.TownLocationName
+		stateOrProvinceName = address.CountrySubDivision
+		countryName = address.Country
+	}
+	if localityName != "" && stateOrProvinceName != "" && countryName != "" {
+		certRequest.Params["localityName"] = localityName
+		certRequest.Params["stateOrProvinceName"] = stateOrProvinceName
+		certRequest.Params["countryName"] = countryName
```
pdeziel, line 228: Should we allow partial location info? I don't think the TRISA address validation covers that case for us.

bbengfort: Oh man, this is going to be a pain in the butt. You're right - we should not have partial location info, and right now the front end uses `AddressLines` instead of `TownLocationName` and `CountrySubDivision` … You'll definitely have `countryName` but you won't have TownLocationName and CountrySubDivision. I don't want `Menlo Park, California, SG`, so I think the check you're doing right now is right.

Let's create a follow on story to see if locality and stateorprovince are optional in Sectigo. If so, we can move forward with just country. We'll then also probably have to create follow on stories to allow these fields to be edited from the Admin UI.

pkg/gds/models/v1/models.go
```diff
@@ -173,6 +176,69 @@ func AppendCertReqID(vasp *pb.VASP, certreqID string) (err error) {
...
+	} else {
+		log.Info().
```
bbengfort, line 230: Could we change this to `Debug` - it will likely be the common case for a while.

pkg/gds/models/v1/models_test.go
```diff
@@ -448,6 +449,79 @@ func TestVeriedContacts(t *testing.T) {
...
+			{
+				LegalPersonName:               "Charlie Inc.",
+				LegalPersonNameIdentifierType: ivms101.LegalPersonShort,
+			},
+		},
+	}
+	cr, err = NewCertificateRequest(vasp)
+	require.NoError(t, err)
+	require.Contains(t, cr.Params, "organizationName")
+	require.Equal(t, "TRISA Production", cr.Params["organizationName"])
```
bbengfort, line 481: It's not `"Charlie Inc."` because it's a short name not a legal name?

pkg/gds/models/v1/models_test.go
```diff
@@ -448,6 +449,79 @@ func TestVeriedContacts(t *testing.T) {
...
+	require.Contains(t, cr.Params, "stateOrProvinceName")
+	require.Equal(t, "California", cr.Params["stateOrProvinceName"])
+	require.Contains(t, cr.Params, "countryName")
+	require.Equal(t, "US", cr.Params["countryName"])
+
+	vasp.Entity.GeographicAddresses = []*ivms101.Address{
+		{
+			TownLocationName:   "Toronto",
+			CountrySubDivision: "Ontario",
+		},
```
bbengfort, line 500: Could we change the partial info test to having Country but not town or state? This is the common case in live data, so I'd like to make sure we're protecting from regression with that case.

pkg/gds/gds.go
```diff
@@ -228,12 +227,13 @@ func (s *GDS) Register(ctx context.Context, in *api.RegisterRequest) (out *api.R
...
+	var certRequest *models.CertificateRequest
 	password := secrets.CreateToken(16)
-	certRequest := &models.CertificateRequest{
-		Id:         uuid.New().String(),
-		Vasp:       vasp.Id,
-		CommonName: vasp.CommonName,
+	if certRequest, err = models.NewCertificateRequest(vasp); err != nil {
+		log.Error().Err(err).Str("vasp", vasp.Id).Msg("could not create certificate request")
+		return nil, status.Error(codes.Internal, "internal error with registration, please contact admins")
 	}
```
bbengfort, line 235: Much neater!

pkg/gds/certs.go
```diff
@@ -143,20 +143,17 @@ func (s *Service) submitCertificateRequest(r *models.CertificateRequest) (err er
...
-		params["localityName"] = "Menlo Park"
-		params["stateOrProvinceName"] = "California"
-		params["countryName"] = "US"
+		params = r.Params
+	} else {
+		params = make(map[string]string)
 	}
+	params["commonName"] = r.CommonName
+	params["dNSName"] = r.CommonName
+	params["pkcs12Password"] = string(pkcs12Password)
```
bbengfort, line 157: Also much neater! 

Do we have tests related to cert manager? I'd like to create a story that uses the sectigo mock to ensure that the params are created correctly for both `ProfileCipherTraceEndEntityCertificate` (includes all the subject information + common name, dns name, password) and for `ProfileCipherTraceEE` (just common name, dns name, password). If we don't have tests related to cert manager, I guess it will be two stories - one to create test stubs for cert manager and the other to implement the test above.

bbengfort (commented):
@pdeziel thank you for taking this on! This story is particularly difficult because we're trying to tie together a change in Sectigo with the original implementation of the VASP and the UI. I've some comments below - happy to discuss them as needed.

pkg/gds/models/v1/models.go
```diff
@@ -173,6 +176,69 @@ func AppendCertReqID(vasp *pb.VASP, certreqID string) (err error) {
...
+	}
+
+	certRequest = &CertificateRequest{
+		Id:         uuid.New().String(),
+		Vasp:       vasp.Id,
+		CommonName: vasp.CommonName,
+		Params:     make(map[string]string),
+	}
+
+	// Populate the organization name, if available.
```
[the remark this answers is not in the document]
pdeziel: Thanks for pointing that out, I'm definitely more comfortable using the pre-existing method.

pkg/gds/models/v1/models_test.go
pdeziel: That was the idea with the current name search, but I'll change this when I switch to using `VASP.Name()`.

pkg/gds/certs.go
pdeziel: We do have tests for the cert manager but I don't the Sectigo mock is checking the parameters we're building here, so I think that would be a good thing to add to validate this code.

proto/gds/models/v1/models.proto
```diff
@@ -21,22 +21,25 @@ message CertificateRequest {
     // Request pipeline status
     CertificateRequestState status = 4;
 
+    // Generic parameters used for making requests to certificate issuing services
+    map<string, string> params = 5;
```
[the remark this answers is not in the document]
pdeziel: That makes complete sense, it seems a lot easier to move the position of the params into the gap so I'll make that change.

pkg/gds/certs.go
bbengfort: Excellent! Would you mind creating a story for it?

bbengfort (approved):
@pdeziel thank you for making those changes!

PR #304, opened 2022-01-19
sc-916 improve contact handling

pdeziel:
This adds some utilities for dealing with the contacts. The main thing is a way to iterate through the contacts on a VASP more easily. Previously, we had to specify the contacts in a slice and then iterate through that slice every time a function wanted to iterate through the contacts. This PR adds two functions for iteration, `IterContacts` and `IterVerifiedContacts`, which both return anonymous functions which can be used to perform the iteration. This ensures that we are always iterating through the contacts in the same order and gets us access to the contact "kinds" if we need them ("technical", "administrative", "legal", and "billing").

### DIFF
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -444,17 +444,11 @@ func (s *Admin) Summary(c *gin.Context) {
 		out.VASPsCount++
 
 		// Count contacts
-		contacts := []*pb.Contact{
-			vasp.Contacts.Administrative, vasp.Contacts.Legal,
-			vasp.Contacts.Technical, vasp.Contacts.Billing,
-		}
-
-		for _, contact := range contacts {
-			if contact != nil && contact.Email != "" {
-				out.ContactsCount++
-				if _, verified, _ := models.GetContactVerification(contact); verified {
-					out.VerifiedContacts++
-				}
+		next := models.IterContacts(vasp.Contacts, true)
+		for contact, _ := next(); contact != nil; contact, _ = next() {
+			out.ContactsCount++
+			if _, verified, _ := models.GetContactVerification(contact); verified {
+				out.VerifiedContacts++
 			}
 		}
 
@@ -875,17 +869,9 @@ func (s *Admin) prepareVASPDetail(vasp *pb.VASP, log zerolog.Logger) (out *admin
 	// Must be done after verified contacts is computed
 	// WARNING: This is safe because nothing is saved back to the database!
 	vasp.Extra = nil
-	if vasp.Contacts.Administrative != nil {
-		vasp.Contacts.Administrative.Extra = nil
-	}
-	if vasp.Contacts.Legal != nil {
-		vasp.Contacts.Legal.Extra = nil
-	}
-	if vasp.Contacts.Technical != nil {
-		vasp.Contacts.Technical.Extra = nil
-	}
-	if vasp.Contacts.Billing != nil {
-		vasp.Contacts.Billing.Extra = nil
+	next := models.IterContacts(vasp.Contacts, false)
+	for contact, _ := next(); contact != nil; contact, _ = next() {
+		contact.Extra = nil
 	}
 
 	// Rewire the VASP from protocol buffers to specific JSON serialization context
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -444,9 +444,10 @@ func (s *Admin) Summary(c *gin.Context) {
 		out.VASPsCount++
 
 		// Count contacts
-		next := models.IterContacts(vasp.Contacts, true)
-		for contact, _ := next(); contact != nil; contact, _ = next() {
+		iter := models.NewContactIterator(vasp.Contacts, true, false)
+		for iter.Next() {
 			out.ContactsCount++
+			contact, _, _ := iter.Value()
 			if _, verified, _ := models.GetContactVerification(contact); verified {
 				out.VerifiedContacts++
 			}
@@ -835,10 +836,7 @@ func (s *Admin) prepareVASPDetail(vasp *pb.VASP, log zerolog.Logger) (out *admin
 	out = &admin.RetrieveVASPReply{
 		Traveler: models.IsTraveler(vasp),
 	}
-	if out.VerifiedContacts, err = models.VerifiedContacts(vasp); err != nil {
-		log.Warn().Err(err).Msg("could not get verified contacts")
-		return nil, err
-	}
+	out.VerifiedContacts = models.VerifiedContacts(vasp)
 
 	// Attempt to determine the VASP name from IVMS 101 data.
 	if out.Name, err = vasp.Name(); err != nil {
@@ -869,8 +867,9 @@ func (s *Admin) prepareVASPDetail(vasp *pb.VASP, log zerolog.Logger) (out *admin
 	// Must be done after verified contacts is computed
 	// WARNING: This is safe because nothing is saved back to the database!
 	vasp.Extra = nil
-	next := models.IterContacts(vasp.Contacts, false)
-	for contact, _ := next(); contact != nil; contact, _ = next() {
+	iter := models.NewContactIterator(vasp.Contacts, false, false)
+	for iter.Next() {
+		contact, _, _ := iter.Value()
 		contact.Extra = nil
 	}
 
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -16,6 +16,7 @@ import (
 	"github.com/gin-gonic/gin"
 	"github.com/golang-jwt/jwt/v4"
 	"github.com/google/uuid"
+	"github.com/hashicorp/go-multierror"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	"google.golang.org/api/idtoken"
@@ -447,8 +448,10 @@ func (s *Admin) Summary(c *gin.Context) {
 		iter := models.NewContactIterator(vasp.Contacts, true, false)
 		for iter.Next() {
 			out.ContactsCount++
-			contact, _, _ := iter.Value()
-			if _, verified, _ := models.GetContactVerification(contact); verified {
+			contact, kind := iter.Value()
+			if verified, err := models.ContactIsVerified(contact); err != nil {
+				log.Warn().Str("contact", kind).Err(err).Msg("could not retrieve verification status")
+			} else if verified {
 				out.VerifiedContacts++
 			}
 		}
@@ -780,9 +783,11 @@ func (s *Admin) ListVASPs(c *gin.Context) {
 			snippet.Name, _ = vasp.Name()
 
 			// Add verified contacts to snippet
-			if snippet.VerifiedContacts, err = models.ContactVerifications(vasp); err != nil {
-				log.Warn().Err(err).Msg("could not get contact verifications")
-				c.JSON(http.StatusInternalServerError, admin.ErrorResponse(err))
+			var errs *multierror.Error
+			if snippet.VerifiedContacts, errs = models.ContactVerifications(vasp); errs != nil {
+				for _, err := range errs.Errors {
+					log.Error().Err(err).Msg("could not get contact verifications")
+				}
 			}
 
 			// Append to list in reply
@@ -834,9 +839,9 @@ func (s *Admin) RetrieveVASP(c *gin.Context) {
 func (s *Admin) prepareVASPDetail(vasp *pb.VASP, log zerolog.Logger) (out *admin.RetrieveVASPReply, err error) {
 	// Create the response to send back
 	out = &admin.RetrieveVASPReply{
-		Traveler: models.IsTraveler(vasp),
+		Traveler:         models.IsTraveler(vasp),
+		VerifiedContacts: models.VerifiedContacts(vasp),
 	}
-	out.VerifiedContacts = models.VerifiedContacts(vasp)
 
 	// Attempt to determine the VASP name from IVMS 101 data.
 	if out.Name, err = vasp.Name(); err != nil {
@@ -869,7 +874,7 @@ func (s *Admin) prepareVASPDetail(vasp *pb.VASP, log zerolog.Logger) (out *admin
 	vasp.Extra = nil
 	iter := models.NewContactIterator(vasp.Contacts, false, false)
 	for iter.Next() {
-		contact, _, _ := iter.Value()
+		contact, _ := iter.Value()
 		contact.Extra = nil
 	}
 
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -785,7 +785,10 @@ func (s *Admin) ListVASPs(c *gin.Context) {
 			snippet.Name, _ = vasp.Name()
 
 			// Add verified contacts to snippet
-			snippet.VerifiedContacts = models.ContactVerifications(vasp)
+			if snippet.VerifiedContacts, err = models.ContactVerifications(vasp); err != nil {
+				log.Warn().Err(err).Msg("could not get contact verifications")
+				c.JSON(http.StatusInternalServerError, admin.ErrorResponse(err))
+			}
 
 			// Append to list in reply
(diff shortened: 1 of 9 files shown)


pkg/gds/models/v1/contacts.go
```diff
@@ -0,0 +1,202 @@
...
+		for i < len(all) {
+			contact = all[i].contact
+			kind = all[i].kind
+			i++
+			if contact != nil && (!requireEmail || ContactHasEmail(contact)) {
+				return contact, kind
+			}
+		}
+		return nil, ""
+	}
```
bbengfort, line 62: This is a very neat implementation, and I admit - not one I've seen before for iteration in Go. I'll be honest that it threw me and I had to read it a few times to fully grok what was going on. I've been playing with it a little bit and I think that I get everything that's happening. 

I honestly feel like this deserves some more commentary, but honestly I don't know what to say!

pkg/gds/models/v1/contacts.go
```diff
@@ -0,0 +1,202 @@
...
+	return []*contactType{
+		{contact: contacts.Technical, kind: TechnicalContact},
+		{contact: contacts.Administrative, kind: AdministrativeContact},
+		{contact: contacts.Legal, kind: LegalContact},
+		{contact: contacts.Billing, kind: BillingContact},
+	}
+}
+
+// Returns a function which iterates over the verified contacts in a Contacts object.
+func IterContacts(contacts *pb.Contacts, requireEmail bool) func() (*pb.Contact, string) {
```
bbengfort, line 48: Could we please type the return function? E.g. 

```go
type ContactIterator func() (*pb.Contact, string)

func IterContacts(contacts *pb.Contacts, requireEmail bool) ContactIterator
```

This will make it a bit easier to read, but also to put a doc string above `ContactIterator` so that we can describe how to use it in a `for` loop.

pkg/gds/models/v1/contacts.go
```diff
@@ -0,0 +1,202 @@
...
+			if contact != nil && (!requireEmail || ContactHasEmail(contact)) {
+				return contact, kind
+			}
+		}
+		return nil, ""
+	}
+}
+
+// Returns a function which iterates over the verified contacts in a Contacts object.
+func IterVerifiedContacts(contacts *pb.Contacts) func() (*pb.Contact, string, error) {
```
bbengfort, line 66: I'm a little uncomfortable with the error return here, so I'd like to discuss it. The loop terminal condition is when contact is `nil` - but that's true when an error is returned; so will the caller have the opportunity to handle an error? 

E.g. if the first contact errors:

```go
for contact, _, err := contacts(); contact != nil; contact, _ err = contacts() {
    // I don't think err will ever print, will it? 
   fmt.Println(err)
}
``

I think this is why the `iter.Next()` syntax works the way it does; to decouple the loop termination context from error handling. 

Moreover, as mentioned in later comments below - this iterator will not allow us to ignore errors; e.g. if one contact errors but the rest are fine, we still may want to send emails or count validation. What do you think?

**Review by bbengfort — COMMENTED** [review]
@pdeziel wow - this is a great PR that cleans up so much code and reduces a lot of technical debt! Also, you've definitely surprised me with your implementation - I very much enjoyed reviewing this PR! A couple of discussion comments below.

File: pkg/gds/models/v1/contacts.go [hunk 4]
```diff
@@ -0,0 +1,202 @@
...
+	extra.Token = token
+	if contact.Extra, err = anypb.New(extra); err != nil {
+		return err
+	}
+	return nil
+}
+
+// VerifiedContacts returns a map of contact type to email address for all verified
+// contacts, omitting any contacts that are not verified or do not exist.
+func VerifiedContacts(vasp *pb.VASP) (contacts map[string]string, err error) {
```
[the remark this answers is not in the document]
pdeziel: Ok, I get the reasoning behind ignoring the errors now; I didn't want to just throw away the error at the utility level so I agree that Option 2 is probably the way to go.

pkg/gds/models/v1/contacts.go
pdeziel: I see what you mean, I think in order to catch the error the caller has to declare the error outside the loop and check the error value after loop termination, which seems like a lot of work for the caller to do. I think the real solution is to replace the closure with an iterator object with Next() and Value(), this allows us to better distinguish between iterator and value errors and allows the errors to be handled independently and matches the `DirectoryIterator` pattern.

pdeziel:
@bbengfort I pushed some changes, let me know what you think of the new refactor. The goal was to make the caller code a lot simpler by using a Next() and Value() pattern for the iteration.

pkg/gds/models/v1/contacts.go
```diff
@@ -0,0 +1,216 @@
...
+	i.current = nil
+	return false
+}
+
+func (i *ContactIterator) Value() (*pb.Contact, string, error) {
+	if i.current != nil {
+		return i.current.contact, i.current.kind, i.current.err
+	}
+	return nil, "", errors.New("no more contacts")
+}
```
bbengfort, line 103: I really did like the iterator closure that you put together, but I think you're right that in this case, we need a more standard iterator struct to deal with error handling. This is a good implementation, but I'd like to recommend that we use recursion here in case we extend our filtering context. I'd also like to suggest we reduce some of our data storage by using the references we already have and we implement a multi-error handling functionality. 

The one big note is that for this recursive implementation to work, the index must be initialized to -1. 

```suggestion
type ContactIterator struct {
	email    bool
	verified bool
	index    int
	contacts []*contactType
}

// Returns a new ContactIterator which has Next() and Value() methods that can be used
// to iterate over contacts in a Contacts object.
func NewContactIterator(contacts *pb.Contacts, requireEmail bool, requireVerified bool) *ContactIterator {
	return &ContactIterator{
		email:    requireEmail,
		verified: requireVerified,
		index: -1,
		contacts: contactOrder(contacts),
	}
}

// Moves the ContactIterator to the next existing contact and returns true if there is
// one, false if the iteration is complete.
func (i *ContactIterator) Next() bool {
	// Advance the index (note, index must be initialized at -1) 
	i.index++
	
	// Stopping condition
	if i.index >= len(i.contacts) {
		return false
	}
	
	// Filter checks - contact must exist
	contact := i.contacts[i.index].contact
	if contact == nil {
		return i.Next()
	}
	
	// Filter if email required
	if i.email && !ContactHasEmail(contact) {
		return i.Next()
	}
	
	// Filter if verified is required
	if i.verified {
		if verified, err := ContactIsVerified(contact); err != nil || !verified {
			// Even in an error we're skipping the contact, errors have to be 
			// fetched as a multi-error after the iteration is complete.
			contact.err = err 
			return i.Next()
		}
	}
	
	return true
}

func (i *ContactIterator) Value() (*pb.Contact, string) {
	if i.index < len(i.contacts) {
		// Note that no checking of the contact occurs here to allow
		// us to allow iteration with different filtering mechanisms.
		current := i.contacts[i.index]
		return current.contact, current.kind
	}
	return nil, ""
}

// Error returns a multi-error that describes any errors that occurred during
// iteration for any contact. This will allow us to log all errors once.
func (i *ContactIterator) Error() error {
	bad := make([]string, 0, 4)
	for _, contact := range i.contacts {
		if contact.err != nil {
			bad = append(bad, contact.type)
		}
	}
	if len(bad) > 0 {
		bads := strings.Join(contacts, ", ")
		return fmt.Errorf("errors occurred processing the %s contacts", bads)
	}
	return nil
}
```

bbengfort: Note that we could actually use a multi-error if you wanted to track all errors across all contacts: https://github.com/hashicorp/go-multierror

pkg/gds/models/v1/contacts.go
```diff
@@ -0,0 +1,216 @@
...
+	iter := NewContactIterator(vasp.Contacts, false, false)
+	for iter.Next() {
+		if contact, kind, err := iter.Value(); err == nil {
+			if verified, err := ContactIsVerified(contact); err == nil {
+				contacts[kind] = verified
+			}
+		}
+	}
+	return contacts, nil
+}
```
bbengfort, line 169: Could we also add a multi-error here, e.g. something similar to the suggested `iter.Error()` above or using https://github.com/hashicorp/go-multierror -- basically in the for loop we have to collect errors, and if an error occurs we have to return the multi-error along with the contact.

pkg/gds/gds.go
```diff
@@ -470,7 +455,7 @@ func (s *GDS) VerifyContact(ctx context.Context, in *api.VerifyContactRequest) (
 		// Perform token check and if token matches, mark contact as verified
 		if token == in.Token {
 			found = true
-			log.Info().Str("vasp", vasp.Id).Int("index", idx).Msg("contact email verified")
+			log.Info().Str("vasp", vasp.Id).Str("contact", kind).Msg("contact email verified")
```
bbengfort, line 458: This is a much nicer log message!

pkg/gds/admin.go
```diff
@@ -444,17 +444,12 @@ func (s *Admin) Summary(c *gin.Context) {
...
-				out.ContactsCount++
-				if _, verified, _ := models.GetContactVerification(contact); verified {
-					out.VerifiedContacts++
-				}
+		iter := models.NewContactIterator(vasp.Contacts, true, false)
+		for iter.Next() {
+			out.ContactsCount++
+			contact, _, _ := iter.Value()
+			if _, verified, _ := models.GetContactVerification(contact); verified {
+				out.VerifiedContacts++
```
bbengfort, line 452: Perhaps we should use `ContactIsVerified` here?

pkg/gds/admin.go
```diff
@@ -785,7 +780,10 @@ func (s *Admin) ListVASPs(c *gin.Context) {
 			snippet.Name, _ = vasp.Name()
 
 			// Add verified contacts to snippet
-			snippet.VerifiedContacts = models.ContactVerifications(vasp)
+			if snippet.VerifiedContacts, err = models.ContactVerifications(vasp); err != nil {
+				log.Warn().Err(err).Msg("could not get contact verifications")
+				c.JSON(http.StatusInternalServerError, admin.ErrorResponse(err))
+			}
```
bbengfort, line 786: Instead of returning a 500 here, let's optimistically try to return the verified contacts we do have, and log the error:

```suggestion
			if snippet.VerifiedContacts, err = models.ContactVerifications(vasp); err != nil {
				log.Error().Err(err).Msg("could not get all verified contacts")
			}
```

pkg/gds/admin.go
```diff
@@ -836,9 +834,9 @@ func (s *Admin) RetrieveVASP(c *gin.Context) {
 func (s *Admin) prepareVASPDetail(vasp *pb.VASP, log zerolog.Logger) (out *admin.RetrieveVASPReply, err error) {
 	// Create the response to send back
 	out = &admin.RetrieveVASPReply{
-		VerifiedContacts: models.VerifiedContacts(vasp),
-		Traveler:         models.IsTraveler(vasp),
+		Traveler: models.IsTraveler(vasp),
 	}
+	out.VerifiedContacts = models.VerifiedContacts(vasp)
```
bbengfort, line 839: We could put this back into the struct now that we're changing the error handling.

pkg/gds/admin.go
```diff
@@ -869,17 +867,10 @@ func (s *Admin) prepareVASPDetail(vasp *pb.VASP, log zerolog.Logger) (out *admin
...
-	}
-	if vasp.Contacts.Technical != nil {
-		vasp.Contacts.Technical.Extra = nil
-	}
-	if vasp.Contacts.Billing != nil {
-		vasp.Contacts.Billing.Extra = nil
+	iter := models.NewContactIterator(vasp.Contacts, false, false)
+	for iter.Next() {
+		contact, _, _ := iter.Value()
+		contact.Extra = nil
```
bbengfort, line 873: This is so much nicer! Sorry - I'm very happy about these changes, it will make things so much easier!

pkg/gds/emails/client.go
```diff
@@ -218,55 +205,49 @@ func (m *EmailManager) SendRejectRegistration(vasp *pb.VASP, reason string) (sen
...
-		vasp.Contacts.Billing, vasp.Contacts.Legal,
-	}
-
 	// Attempt at least one delivery, don't give up just because one email failed
 	// Track how many emails and errors occurred during delivery.
 	var nErrors uint8
-	for idx, contact := range contacts {
-		// Skip any contacts that we can't send emails to
-		if contact == nil || contact.Email == "" {
+	iter := models.NewContactIterator(vasp.Contacts, false, true)
```
bbengfort, line 211: Should this be `NewContactIterator(vasp.Contacts, true, true)` -- e.g. both verified contacts and those contacts with emails?

pkg/gds/emails/client.go
```diff
@@ -218,55 +205,49 @@ func (m *EmailManager) SendRejectRegistration(vasp *pb.VASP, reason string) (sen
...
+	for iter.Next() {
+		var contact *pb.Contact
+		var kind string
+		if contact, kind, err = iter.Value(); err != nil {
+			log.Error().Err(err).Str("vasp", vasp.Id).Msg("could not retrieve contact info")
+			nErrors++
```
bbengfort, line 217: With the suggestion for multi-error below; we don't have to track nErrors, we can just log the error at the end of the iteration if it occurred.

pkg/gds/emails/client.go
```diff
@@ -288,60 +269,48 @@ func (m *EmailManager) SendDeliverCertificates(vasp *pb.VASP, path string) (sent
...
-		vasp.Contacts.Legal, vasp.Contacts.Billing,
-	}
-
 	// Attempt at least one delivery, don't give up just because one email failed
 	// Track how many emails and errors occurred during delivery.
 	var nErrors uint8
-	for idx, contact := range contacts {
-		// Skip any null contacts or contacts without email addresses
-		if contact == nil || contact.Email == "" {
+	iter := models.NewContactIterator(vasp.Contacts, false, true)
```
bbengfort, line 275: Should this be `NewContactIterator(vasp.Contacts, true, true)`?

bbengfort (commented):
Thank you for diving into this a second time @pdeziel! I think we're close - there are still some error handling things that I'd like to smooth over a bit. I'm learning about [error wrapping](https://go.dev/doc/go1.13#error_wrapping) which will be a handy technique in the future, but perhaps we'd like to use [go-multierr](https://github.com/hashicorp/go-multierror) for this use case?

pkg/gds/models/v1/models_test.go
```diff
@@ -436,6 +436,7 @@ func TestVeriedContacts(t *testing.T) {
 	require.NoError(t, err)
 
 	contacts = VerifiedContacts(vasp)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
pdeziel: Yes they did, good catch.

pdeziel:
@bbengfort I've pushed your recursion changes and the new multierror handling for your review.

pkg/gds/models/v1/contacts_test.go
```diff
@@ -0,0 +1,145 @@
...
+		contact, kind := iter.Value()
+		actualContacts = append(actualContacts, contact)
+		actualKinds = append(actualKinds, kind)
+	}
+	require.NoError(t, iter.Error())
+	require.Equal(t, expectedContacts, actualContacts)
+	require.Equal(t, expectedKinds, actualKinds)
+}
+
+func TestIterVerfiiedContacts(t *testing.T) {
```
bbengfort, line 92:
```suggestion
func TestIterVerifiedContacts(t *testing.T) {
```

bbengfort (approved):
It was a long road, but I think it was worth it - the contact handling code is so much cleaner and more readable with these helper functions. Thank you!

pkg/gds/models/v1/contacts_test.go
pdeziel: Agreed, the error handling was a bit of a headache but I'm pretty happy with how this turned out!

PR #305, opened 2022-01-20
sc-2540 change duplicate name warning to debug

pdeziel:
This changes the duplicate name warning on the autocomplete endpoint to a debug to avoid the alert spam.


rebeccabilbro (approved):
Small commit, huge impact 😉  -- thank you @pdeziel!

PR #306, opened 2022-01-20
sc-2938 fix sectigo testing behavior for dev

pdeziel:
This fixes two things related to the sectigo configuration for the dev environment. The first is that the `SECTIGO_TESTING` flag was not being picked up from the configuration. The second is that GDS was erroring on startup if the sectigo testing flag was set. This has been fixed by moving the localhost check to when we try to send requests instead of on startup. This allows us to boot up GDS in Sectigo testing mode without it crashing.


bbengfort (approved):
Looks good, thank you for making this fix @pdeziel!

PR #307, opened 2022-01-21
sc-848 DirectoryIterator improved error handling

pdeziel:
This changes the `DirectoryIterator` methods so that they return an error if we are unable to parse the object from the database. This allows us to better distinguish between iteration errors and parsing errors (corrupted data) when iterating over VASPs or CertReqs.

### DIFF
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -436,8 +436,9 @@ func (s *Admin) Summary(c *gin.Context) {
 	for iter.Next() {
 		// Fetch VASP from the database
 		var vasp *pb.VASP
-		if vasp = iter.VASP(); vasp == nil {
-			// VASP could not be parsed; error logged in VASP() method continue iteration
+		var err error
+		if vasp, err = iter.VASP(); err != nil {
+			log.Error().Err(err).Msg("could not parse VASP from database")
 			continue
 		}
 
@@ -480,8 +481,9 @@ func (s *Admin) Summary(c *gin.Context) {
 	for iter2.Next() {
 		// Fetch CertificateRequest from the database
 		var certreq *models.CertificateRequest
-		if certreq = iter2.CertReq(); certreq == nil {
-			// CertificateRequest could not be parsed; error logged in CertReq() method continue iteration
+		var err error
+		if certreq, err = iter2.CertReq(); err != nil {
+			log.Error().Err(err).Msg("could not parse CertificateRequest from database")
 			continue
 		}
 
@@ -520,8 +522,9 @@ func (s *Admin) Autocomplete(c *gin.Context) {
 	for iter.Next() {
 		// Fetch VASP from the database
 		var vasp *pb.VASP
-		if vasp = iter.VASP(); vasp == nil {
-			// VASP could not be parsed; error logged in VASP() method continue iteration
+		var err error
+		if vasp, err = iter.VASP(); err != nil {
+			log.Error().Err(err).Msg("could not parse VASP from database")
 			continue
 		}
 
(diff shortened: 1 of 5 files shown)


pkg/gds/admin.go
```diff
@@ -436,8 +436,9 @@ func (s *Admin) Summary(c *gin.Context) {
 	for iter.Next() {
 		// Fetch VASP from the database
 		var vasp *pb.VASP
-		if vasp = iter.VASP(); vasp == nil {
-			// VASP could not be parsed; error logged in VASP() method continue iteration
+		var err error
+		if vasp, err = iter.VASP(); err != nil {
+			log.Error().Err(err).Msg("could not parse VASP from database")
 			continue
 		}
```
bbengfort, line 443: The following comment has nothing to do with this PR - but I have been thinking about allocs and escapes in code recently; particularly with respect to the performance of Go code. Finding ways to prevent allocs by reusing the same register I think is a good practice that makes escape analysis easier in general; even if most of the time the value is `nil`. 

No change to the code is necessary, but I did want to make a comment about variable declaration, scoping, and allocs. If I have two variables that are both undeclared, I do prefer the following syntax:

```go
vasp, err := iter.VASP()
if err != nil {
    log.Error().Err(err).Msg("")
    continue
}
```

This shortens it by a line but I think it also makes it clearer which block the variable is local to; e.g. when compared to:

```go
if vasp, err := iter.VASP(); err != nil {
     // vasp and err are only local in this block and related else/else if blocks. 
}
```

That said, later on in line 471 and 496, we have another err assignment; so potentially this block might be more easily written as:

```go
	var err error
	for iter.Next() {
		// Fetch VASP from the database
		var vasp *pb.VASP
		if vasp, err = iter.VASP(); err != nil {
			log.Error().Err(err).Msg("could not parse VASP from database")
			continue
		}

		…
	}

	if err = iter.Error(); err != nil {
		iter.Release()
		log.Warn().Err(err).Msg("could not iterate over vasps in store")
		c.JSON(http.StatusInternalServerError, admin.ErrorResponse(err))
		return
	}

	...

	for iter2.Next() {
		// Fetch CertificateRequest from the database
		var certreq *models.CertificateRequest
		if certreq, err = iter2.CertReq(); err != nil {
			log.Error().Err(err).Msg("could not parse CertificateRequest from database")
			continue
		}
		...
	}

	...
	
	if err = iter2.Error(); err != nil {
		iter2.Release()
		log.Warn().Err(err).Msg("could not iterate over certreqs in store")
		c.JSON(http.StatusInternalServerError, admin.ErrorResponse(err))
		return
	}
```

pkg/gds/certs.go
```diff
@@ -62,8 +62,9 @@ func (s *Service) CertManager(stop <-chan struct{}) {
 		log.Debug().Msg("cert-manager checking certificate request pipelines")
 
 		for careqs.Next() {
-			req := careqs.CertReq()
-			if req == nil {
+			var req *models.CertificateRequest
+			if req, err = careqs.CertReq(); err != nil {
+				log.Error().Err(err).Msg("could not parse certificate request from database")
```
bbengfort, line 67: Thank you for adding the logging statements!

pkg/gds/members.go
```diff
@@ -197,7 +197,11 @@ func (s *Members) List(ctx context.Context, in *api.ListRequest) (out *api.ListR
 		}
 
 		// Collect the VASP from the iterator
-		vasp := iter.VASP()
+		var vasp *pb.VASP
+		if vasp, err = iter.VASP(); err != nil {
+			log.Warn().Err(err).Msg("could not parse VASP from database")
```
bbengfort, line 202: In certs.go you used `log.Error()` and here `log.Warn()` - any reason for the difference?

pkg/gds/admin.go
pdeziel: Thanks for bringing this up, definitely good to keep in mind for future PRs.

pkg/gds/members.go
pdeziel: Good point, I think I intended this to be a `log.Error`.

PR #308, opened 2022-01-24
Fix admin UI syntax errors

elysee15:
SC-3053


pdeziel (approved):
Looks good to me, thanks for addressing this so quickly! I will create a story to look into the codecov issue.

PR #310, opened 2022-01-24
sc-2466 add trtl store package

pdeziel:
This adds the beginnings of the trtl store package with unimplemented stubs for the `DirectoryStore` and `CertificateStore` methods.

### DIFF
--- a/pkg/gds/store/trtl/trtl.go
+++ b/pkg/gds/store/trtl/trtl.go
@@ -1,0 +1,92 @@
+package trtl
+
+import (
+	"errors"
+
+	"github.com/rs/zerolog/log"
+	"github.com/trisacrypto/directory/pkg/gds/client"
+	"github.com/trisacrypto/directory/pkg/gds/models/v1"
+	"github.com/trisacrypto/directory/pkg/gds/store/iterator"
+	"github.com/trisacrypto/directory/pkg/trtl/pb/v1"
+	gds "github.com/trisacrypto/trisa/pkg/trisa/gds/models/v1beta1"
+	"google.golang.org/grpc"
+)
+
+// Open a connection to the Trtl database.
+func Open(profile *client.TrtlProfile) (store *Store, err error) {
+	store = &Store{}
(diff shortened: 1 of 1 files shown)


pkg/gds/store/trtl/trtl.go
```diff
@@ -0,0 +1,92 @@
...
+
+// Open a connection to the Trtl database.
+func Open(profile *client.TrtlProfile) (store *Store, err error) {
+	store = &Store{}
+	if store.conn, err = profile.Connect(); err != nil {
+		return nil, err
+	}
+	store.client = pb.NewTrtlClient(store.conn)
+	return &Store{}, nil
+}
```
pdeziel, line 23: Do we plan on integrating this into the generic `store.Store.Open`? It seems like it might be a bit awkward, because that method is expecting to deal with embedded databases and we need to somehow pass in the mTLS parameters.

bbengfort: As discussed, we'll go with making the `store.Store.Open` a bit more explicit with the database config.

pkg/gds/store/trtl/trtl.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/trisacrypto/directory/pkg/gds/client"
+	"github.com/trisacrypto/directory/pkg/gds/models/v1"
+	"github.com/trisacrypto/directory/pkg/gds/store/iterator"
+	"github.com/trisacrypto/directory/pkg/trtl/pb/v1"
+	gds "github.com/trisacrypto/trisa/pkg/trisa/gds/models/v1beta1"
+	"google.golang.org/grpc"
+)
+
+// Open a connection to the Trtl database.
+func Open(profile *client.TrtlProfile) (store *Store, err error) {
```
bbengfort, line 16: I can see profile making sense here, but we will have to load the configuration from the environment on the GDS server; when we get to that point we should discuss if we want to decouple these or not.

bbengfort (approved):
Good start!

PR #311, opened 2022-01-25
sc-2407 Remove ignore of LegalPerson error

pdeziel:
This removes the ignore of the LegalPerson error now that we have patched the validation check in TRISA.


rebeccabilbro (approved):
Excellent, thank you for closing the loop on this @pdeziel!

PR #314, opened 2022-02-01
sc-2405 allow contact updates on UpdateVASP

pdeziel:
This modifies the `UpdateVASP` endpoint so that it allows updates to the VASP contacts. I ran into a lot of design decisions with this one so I have some questions and comments in the PR.


fixtures/datagen/templates/no_cert.json
```diff
@@ -47,42 +47,9 @@
...
-            "extra": {
-                "@type": "type.googleapis.com/gds.models.v1.GDSContactExtraData",
-                "verified": false,
-                "token": "legal_token",
-                "email_log": []
-            }
-        },
+        "technical": null,
+        "administrative": null,
+        "legal": null,
```
pdeziel, line 52: Changed so that the fixtures work with TRISA validation (it expects contacts which are not nil to have an email address).

pkg/gds/admin.go
```diff
@@ -1139,6 +1149,89 @@ func (s *Admin) updateVASPTRIXO(vasp *pb.VASP, data map[string]interface{}, log
...
+				continue
+			}
+
+			// Write the updated email log to the database.
+			if err := s.db.UpdateVASP(vasp); err != nil {
+				errs = multierror.Append(errs, err)
+				continue
+			}
+		}
+	}
```
pdeziel, line 1230: This code could be simplified a lot if we don't allow contacts to be added/removed on the update. Let me know what you think.

pkg/gds/admin_test.go
```diff
@@ -531,6 +529,7 @@ func (s *gdsTestSuite) TestListVASPs() {
 
 	// List VASPs with the specified status
 	request.path = "/v2/vasps?status=" + snippets[0].VerificationStatus
+	actual = &admin.ListVASPsReply{}
```
pdeziel, line 532: Chased a really weird bug around for an hour because we were not resetting the target pointer for the JSON marshal in `doRequest`. I guess we just got lucky and it wasn't causing us issues until now.

pkg/gds/emails/client.go
```diff
@@ -85,43 +85,20 @@ func (m *EmailManager) SendVerifyContacts(vasp *pb.VASP) (sent int, err error) {
...
-			sent++
-
-			if err = models.AppendEmailLog(contact, string(admin.ResendVerifyContact), msg.Subject); err != nil {
-				log.Error().Err(err).Str("vasp", vasp.Id).Str("contact", kind).Msg("could not log verify contact email")
+			if err := m.SendVerifyContact(vasp, contact); err != nil {
 				nErrors++
-				continue
+				log.Error().Err(err).Str("vasp", vasp.Id).Str("contact", kind).Msg("failed to send verify contact email")
+			} else {
+				sent++
```
pdeziel, line 101: Refactoring to reduce repeated code, originally I was sending emails to all the unverified contacts on the update but I thought about it some more and I think it only makes sense to send the email to only the contacts that were updated?

pkg/gds/admin_test.go
```diff
@@ -542,18 +541,21 @@ func (s *gdsTestSuite) TestListVASPs() {
...
 	request.path = "/v2/vasps?status=" + snippets[0].VerificationStatus + "&status=" + snippets[1].VerificationStatus
+	actual = &admin.ListVASPsReply{}
 	c, w = s.makeRequest(request)
 	rep = s.doRequest(a.ListVASPs, c, w, actual)
 	require.Equal(http.StatusOK, rep.StatusCode)
 	require.Equal(2, actual.Count)
 	require.Equal(1, actual.Page)
 	require.Equal(100, actual.PageSize)
 	require.Len(actual.VASPs, 2)
+	fmt.Println(actual.VASPs)
```
pdeziel, line 552:
```suggestion
```

pkg/gds/admin_test.go
```diff
@@ -643,6 +648,7 @@ func (s *gdsTestSuite) TestRetrieveVASP() {
 func (s *gdsTestSuite) TestUpdateVASP() {
 	s.LoadSmallFixtures()
 	defer s.ResetSmallFixtures()
+	defer emails.PurgeMockEmails()
```
bbengfort, line 651: Got left behind from refactor - I don't think UpdateVASP sends emails anymore - but not a big deal.

pkg/gds/admin_test.go
```diff
@@ -765,6 +772,244 @@ func (s *gdsTestSuite) TestDeleteVASP() {
...
+		Contact: map[string]interface{}{},
+	}
+	c, w = s.makeRequest(request)
+	rep = s.doRequest(a.ReplaceContact, c, w, nil)
+	s.APIError(http.StatusBadRequest, "contact data is required for ReplaceContact request", rep)
+
+	// Test removing a contact email is not allowed
+	contact = &pb.Contact{
+		Email: "",
+	}
```
bbengfort, line 840: Good test!

pkg/gds/emails/client.go
bbengfort: I appreciate it - it definitely simplifies things!

bbengfort (approved):
Quite a big PR - but I think everything was well organized and very clear. Also, the tests were quite extensive - which is great! 

Would you mind creating a follow-on PR for contact management in the UI please? 

Also, because there are such big changes, before our next deployment, I'd like to do a QA test as a group to make sure everything is working the way we expect. I'll create a story for that as well.

PR #315, opened 2022-02-01
Anti-entropy metrics with prometheus sc-2365 (v2)

rebeccabilbro:
This PR adds replication monitoring to Trtl with prometheus, moving metrics into a standalone package that can be accessed from the various places within `trtl` where we'd like to be tracking behavior.

Requirements:

- [x] number of initiated anti-entropy sessions per peer/per region (initiator perspective)
- [x] number of received anti-entropy sessions per peer/per region (remote perspective) 
- [x] phase 1 (initiator perspective) anti-entropy session latency per peer 
- [x] phase 2 (remote perspective) anti-entropy session latency per peer 
- [x] total (initiator perspective) anti-entropy session latency per peer 
- [x] number of skips (initiator and remote perspective)
- [x] number of stomps (initiator and remote perspective)
- [x] number of versions (initiator and remote perspective)
- [x] number of updates (initiator and remote perspective)
- [x] number of repairs (initiator and remote perspective)

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,16 +106,14 @@ Server side configuration is done with the environment. Please see the instructi
 
 Client-side configuration is setup using profiles. A profile is a set of related configurations for both development and production. For example, the default environments are "production" to connect to vaspdirectory.net, "testnet" to connect to trisatest.net, and "localhost" to connect to locally running development servers. The profiles are configured in a YAML file that is stored in an OS-specific configuration directory.
 
-You must first install the CLI Tools described in the section above. After that, install the profiles helper tools:
+You must first install the CLI Tools described in the section above. After that, install the profiles helper tool:
 
 
 ```
 $ gds profiles --install
-$ gdsutil profiles --install
-$ trtl profiles --install
 ```
 
-This will create YAML files in your OS-specific configuration directory. To view the path of the configuration files (e.g. for the GDS tool):
+This will create YAML files in your OS-specific configuration directory. To view the path of the configuration file:
 
 ```
 $ gds profiles --path
--- a/pkg/gds/client/profiles.go
+++ b/pkg/gds/client/profiles.go
@@ -50,37 +50,43 @@ const (
 func DefaultProfiles() *Profiles {
 	return &Profiles{
 		Version: ProfileVersion,
-		Active:  "testnet",
+		Active:  "localhost",
 		Profiles: map[string]*Profile{
 			"production": {
 				Directory: &DirectoryProfile{
 					Endpoint: "api.vaspdirectory.net:443",
+					Insecure: true,
 				},
 				Admin: &AdminProfile{
 					Endpoint: "https://api.admin.vaspdirectory.net",
 				},
 				Members: &MembersProfile{
 					Endpoint: "members.vaspdirectory.net:443",
+					Insecure: true,
 				},
 				TrtlProfiles: []*TrtlProfile{
 					{
 						Endpoint: "trtl.us.vaspdirectory.net:443",
+						Insecure: true,
 					},
 				},
 			},
 			"testnet": {
 				Directory: &DirectoryProfile{
 					Endpoint: "api.trisatest.net:443",
+					Insecure: true,
 				},
 				Admin: &AdminProfile{
 					Endpoint: "https://api.admin.trisatest.net",
 				},
 				Members: &MembersProfile{
 					Endpoint: "members.trisatest.net:443",
+					Insecure: true,
 				},
 				TrtlProfiles: []*TrtlProfile{
 					{
 						Endpoint: "trtl.us.trisatest.net:443",
+						Insecure: true,
 					},
 				},
 			},
--- a/pkg/trtl/metrics/metrics.go
+++ b/pkg/trtl/metrics/metrics.go
@@ -39,6 +39,10 @@ type MetricsService struct {
 	srv *http.Server
 }
 
+// New creates a metrics service and also initializes all of the prometheus metrics.
+// The trtl server *must* create the metrics service by calling New before any
+// metrics are logged to Prometheus. Even in the case of tests, the metrics service
+// must be created before the tests can be run.
 func New() (*MetricsService, error) {
 	initMetrics()
 	return &MetricsService{srv: &http.Server{}}, nil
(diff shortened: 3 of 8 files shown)


README.md
```diff
@@ -106,16 +106,14 @@ Server side configuration is done with the environment. Please see the instructi
 
 Client-side configuration is setup using profiles. A profile is a set of related configurations for both development and production. For example, the default environments are "production" to connect to vaspdirectory.net, "testnet" to connect to trisatest.net, and "localhost" to connect to locally running development servers. The profiles are configured in a YAML file that is stored in an OS-specific configuration directory.
 
-You must first install the CLI Tools described in the section above. After that, install the profiles helper tools:
+You must first install the CLI Tools described in the section above. After that, install the profiles helper tool:
```
rebeccabilbro, line 109: These were some changes we discussed a while back and not actually related to this PR, but didn't want to forget

pkg/gds/client/profiles.go
```diff
@@ -50,37 +50,43 @@ const (
 func DefaultProfiles() *Profiles {
 	return &Profiles{
 		Version: ProfileVersion,
-		Active:  "testnet",
+		Active:  "localhost",
```
rebeccabilbro, line 53: Another few changes not strictly related to this PR but just to devEx; hope you don't mind me including them here.

pkg/trtl/metrics/metrics.go
```diff
@@ -0,0 +1,239 @@
...
+	PmRPCLatency *prometheus.HistogramVec // the time it is taking for successful RPC calls to complete, labeled by RPC type, success, and failure
+	// PmObjects    *prometheus.CounterVec   // count of objects being managed by trtl, by namespace
+	// PmTombstones *prometheus.CounterVec   // count of tombstones per namespace; increases on delete, decrease on overwrite of tombstone
+
+	// Anti-Entropy Metrics
+	PmAESyncs         *prometheus.CounterVec   // count of anti entropy sessions per peer, per region, and by perspective (initiator/remote)
+	PmAESyncLatency   *prometheus.HistogramVec // duration of anti entropy sessions (initiator perspective), by peer
+	PmAEPhase1Latency *prometheus.HistogramVec // time phase 1 of anti-entropy is taking from the perspective of the initiator, by peer
+	PmAEPhase2Latency *prometheus.HistogramVec // time phase 2 of anti-entropy is taking from the perspective of the remote, by peer
+	PmAEVersions      *prometheus.HistogramVec // count of all observed versions, per peer and region
```
rebeccabilbro, line 29: potentially this makes more sense as a counter?

pkg/trtl/metrics/metrics.go
```diff
@@ -0,0 +1,239 @@
...
+	// 	Namespace: PmNamespace,
+	// 	Name:      "objects",
+	// 	Help:      "the count of trtl objects, labeled by namespace",
+	// }, []string{"namespace"})
+
+	// PmTombstones = prometheus.NewCounterVec(prometheus.CounterOpts{
+	// 	Namespace: PmNamespace,
+	// 	Name:      "tombstones",
+	// 	Help:      "the count of tombstones, labeled by namespace",
+	// }, []string{"namespace"})
```
rebeccabilbro, line 119: Need to remember to create a backlog story to work on these

pkg/gds/client/profiles.go
bbengfort: I think this is very helpful since the primary use case is development and the secondary use case is managing the network - thank you for including them here!

pkg/trtl/metrics/metrics.go
bbengfort: I see what you're saying - but that's a tough one. On the one hand, we could count all the version vectors we send, which is a proxy for baseline cost; on the other we measure the size of the database and how it grows over time. For now (e.g. until we start monitoring in earnest to figure out what we're looking for) I think the histogram is the most understandable so I agree with your choice here.

pkg/trtl/metrics/metrics.go
```diff
@@ -0,0 +1,239 @@
...
+
+)
+
+// A MetricsService manages Prometheus metrics
+type MetricsService struct {
+	srv *http.Server
+}
+
+func New() (*MetricsService, error) {
+	initMetrics()
```
bbengfort, line 43: By adding `initMetrics` here, the metrics are unusable (they're nil, which will cause a panic if they're accessed) until the code creates a new `MetricsService`. This makes sense to me as it's the intended use case - what's the point of metrics without prometheus? (The alternative is to make `initMetrics` simply `init` - which will be called automatically when the package is imported the first time and there is no need to call `New`). However, I think that we need to document this in the comment above new, e.g.

```go
// New creates a metrics service and also initializes all of the prometheus metrics.
// Ensure that the trtl server creates the metrics service by calling New before any
// metrics are logged to Prometheus. Even in the case of tests, the metrics service
// must be created before the tests can be run. 
func New() (*MetricsService, error)

```

pkg/trtl/metrics/metrics.go
```diff
@@ -0,0 +1,239 @@
...
+		if err := m.srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			log.Err(err).Msg("metrics server shutdown prematurely")
+		}
+	}()
+	return nil
+}
+
+// Gracefully shutdown the Prometheus metrics service
+func (m *MetricsService) Shutdown() error {
+	// Might want to share context from Trtl more globally?
```
bbengfort, line 65: I think we would, would you mind creating a 1 point story for this?

pkg/trtl/replica/replica.go
```diff
@@ -137,13 +142,19 @@ func (r *Service) remotePhase1(ctx context.Context, wg *sync.WaitGroup, log zero
 	// When phase 3 is complete (or if phase 1 ends early) log anti-entropy
 	defer func() {
 		nUpdates := atomic.LoadUint64(&updates)
-		rRepairs := atomic.LoadUint64(&repairs)
-		if nUpdates > 0 || rRepairs > 0 {
+		nRepairs := atomic.LoadUint64(&repairs)
```
bbengfort, line 145: :-) thanks for the typo fix!

bbengfort (approved):
Creating the new metrics package was a simple and effective way to decouple the prometheus tracking from the trtl code. This is an excellent foundation to get us started with tracing and monitoring and it will definitely enhance our experiments with anti-entropy moving forward!

PR #316, opened 2022-02-02
sc-2467 trtl DirectoryStore

pdeziel:
This implements the DirectoryStore methods for the trtl store. One thing absent from this PR is the index handling but I've added some TODO comments in the methods to capture that. This also adds a bug fix for the trtl `Iter()`, we were not checking the tombstone on the objects and so were accidentally returning deleted objects.


pkg/gds/store/trtl/iterator.go
```diff
@@ -0,0 +1,222 @@
...
+
+		var reply *trtlpb.IterReply
+		var err error
+		if reply, err = iter.client.Iter(ctx, request); err != nil {
+			iter.err = err
+			return
+		}
+		iter.values = reply.Values
+		iter.nextPageToken = reply.NextPageToken
+	}
```
pdeziel, line 82: My thought was that most of the time we are iterating through the VASPs we just want to get a materialized view of them for the summary endpoint etc. Although for the CertReqs it might make more sense to stream them from the cursor since we process them one by one in the cert manager. So currently the iterator is configurable with the snapshot bool here, let me know what you think.

pkg/gds/store/trtl/trtl.go
```diff
@@ -19,7 +33,7 @@ func Open(profile *client.TrtlProfile) (store *Store, err error) {
 		return nil, err
 	}
 	store.client = pb.NewTrtlClient(store.conn)
-	return &Store{}, nil
+	return store, nil
```
pdeziel, line 36: :facepalm:

pkg/gds/store/trtl/iterator.go
bbengfort: I think it's a good idea to have two different iterators for both pagination contexts, and I definitely agree with the use cases you mentioned. It feels like kind of a lot for a single type though - do you think breaking this into two types like `trtlBatchIterator` and `trtlStreamingIterator` would be helpful for testing and maintenance? Of course if this were two types, something like the `iterWrapper` would be necessary above, but we'd have to have the ability to select which of those types we wanted to use.

bbengfort, line 82: I recommend making the first request on `Next` instead of `New` so that the client can perform other actions (like Seek) before calling `Next`).

pkg/gds/store/trtl/iterator.go
```diff
@@ -0,0 +1,222 @@
...
+	iter.index = -1
+
+	return iter
+}
+
+func (i *trtlIterator) nextCursor() bool {
+	if i.cursor == nil {
+		i.err = errors.New("nil cursor encountered on Next()")
```
bbengfort, line 90: Could we change this to `errors.New("cursor not initialized correctly before calling next")`?

pkg/gds/store/trtl/iterator.go
```diff
@@ -0,0 +1,222 @@
...
+	i.index++
+	if i.index >= len(i.values) {
+		// Fetch a new value from the cursor
+		var value *trtlpb.KVPair
+		var err error
+		if value, err = i.cursor.Recv(); err != nil {
+			if err != io.EOF {
+				i.err = err
+			}
+			i.cancel()
```
bbengfort, line 103: The caller must call `Release` right? Because of that, I don't think the cancel is necessary here.

pkg/gds/store/trtl/iterator.go
```diff
@@ -0,0 +1,222 @@
...
+			return false
+		}
+		i.values = append(i.values, value)
```
bbengfort, line 106: I think if we separate the two mechanisms, "Batch" (Iter) and "Streaming" (Cursor), then batch should keep things in memory, but streaming should not. The reason you'd choose batch is for requests that materialize a smaller number of rows - hence you're not too worried about snapshot isolation between pages, likely at most one page will be fetched. With Cursor,  you might be fetching an arbitrarily large number of rows, hence the need for a snapshot, but also memory safety. 

This will prevent `Prev` from working with `Cursor` - but I think that's an acceptable tradeoff, if `Prev` is needed, use `Batch` instead. Similarly, `Seek` will only work with `Cursor`.

pkg/gds/store/trtl/iterator.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	return true
+}
+
+func (i *trtlIterator) nextIter() bool {
+	i.index++
+	if i.index >= len(i.values) && i.nextPageToken != "" {
+		ctx, cancel := getContext()
+		defer cancel()
```
bbengfort, line 116: If you create a new context here, it will be separated from the original context. Consider if the original context had a timeout of 5 seconds and it took 3 seconds to get to this point, this first context timeout will go off before this context has a chance to finish. The way to handle this is to make the original context the parent of the new context:

```
context.WithTimeout(originalCtx, time.Second*5)
```

Perhaps `getContext` can be adapted to take a `context` as an argument instead of defaulting to `context.Background`? or there can be a `withContext` function that is the same as `getContext` (e.g. configurable) that uses the passed in context?

pkg/gds/store/trtl/iterator.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+func (i *trtlIterator) Seek(key []byte) bool {
+	for current := i.Value(); bytes.Compare(current, key) < 0; current = i.Value() {
+		if !i.Next() {
+			return false
+		}
+	}
+	return true
+}
```
bbengfort, line 170: This is a pretty decent implementation of forward Seek, but there are a lot of hidden gotchas here. I propose that instead of seeking our array of values in memory, that we make Seek callable only once, before a call to Next and that we pass the seek argument into the CursorRequest. This offloads Seek to the `Trtl` database. Note that means Seek won't be available in `Iter` mode.

pkg/gds/store/trtl/iterator.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	if err = i.iter.Error(); err != nil {
+		return nil, err
+	}
+	return vasps, nil
+}
+
+func (i *vaspIterator) Id() string {
+	// The VASP ID is prefix + uuid so strip off the prefix and return the string
```
bbengfort, line 216: Stray comment from leveldb?

pkg/gds/store/trtl/mock.go
```diff
@@ -0,0 +1,16 @@
+package trtl
+
+import (
+	"github.com/trisacrypto/directory/pkg/trtl/pb/v1"
+	"google.golang.org/grpc"
+)
+
+// NewMock creates a mocked trtl store from a custom grpc client connection to enable
+// testing with bufconn.
+func NewMock(conn *grpc.ClientConn) (store *Store, err error) {
```
bbengfort, line 10: Nice!

pkg/gds/store/trtl/trtl.go
```diff
@@ -1,15 +1,29 @@
...
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
+)
+
+// Errors that may occur during Trtl operations.
+var (
+	ErrIncompleteRecord = errors.New("record is missing required fields")
+	ErrEntityNotFound   = errors.New("entity not found")
+	ErrProtocol         = errors.New("unexpected protocol error")
 )
```
bbengfort, line 27: Should we move these to a `store/errors` package so they can be shared between leveldb and trtl?

pkg/gds/store/trtl/trtl_test.go
```diff
@@ -0,0 +1,238 @@
...
+			Insecure: true,
+		},
+	}
+
+	// Mark config as processed
+	conf, err = conf.Mark()
+	require.NoError(err)
+	s.conf = &conf
+
+	trtl, err := trtl.New(*s.conf)
```
bbengfort, line 69: This is super cool - we're just creating a mock `trtl` server without doing any networking. You can't do that with postgres!

bbengfort (commented):
@pdeziel this is a very good start to the DirectoryStore interface! I had some questions and comments that I'd like to go over (don't worry, it won't be as bad as the Contacts endpoint PR!) Perhaps we could chat today?

pkg/gds/store/trtl/trtl.go
bbengfort: Good catch!

pkg/gds/store/trtl/iterator.go
```diff
@@ -0,0 +1,281 @@
...
+
+	"github.com/rs/zerolog/log"
+	trtlpb "github.com/trisacrypto/directory/pkg/trtl/pb/v1"
+	"github.com/trisacrypto/directory/pkg/utils/wire"
+	pb "github.com/trisacrypto/trisa/pkg/trisa/gds/models/v1beta1"
+	"google.golang.org/protobuf/proto"
+)
+
+type vaspIterator struct {
+	trtlIterator
```
bbengfort, line 16: I wasn't aware that you could embed an interface into a struct - that's very handy!

pkg/gds/store/trtl/iterator.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	if i.index >= len(i.values)-1 && i.nextPageToken == "" {
+		// No more values from the iterator
+		return false
+	}
+
+	i.index++
+	if i.index > 0 && i.index < len(i.values) {
+		// We have already loaded the next value
```
bbengfort, line 71: This is very clever logic, but really depends on the comparison operators (e.g. >= will not work). Just for future us, it might be helpful to comment this a bit more specifically:

```suggestion
		// If the index is not 0 and is in range of the current set of values, we do not have
		// to fetch the next page since the next value is already loaded.
```

pkg/gds/store/trtl/iterator.go
```diff
@@ -0,0 +1,281 @@
...
+
+func (i *trtlBatchIterator) Next() bool {
+	if i.err != nil {
+		return false
+	}
+
+	if i.index >= len(i.values)-1 && i.nextPageToken == "" {
+		// No more values from the iterator
+		return false
+	}
```
bbengfort, line 67: In the starting case, `i.index == -1 && len(i.values)-1 == -1` right? So wouldn't the starting case return false immediately? Am I missing something?

```suggestion
	// If our next index will not refer to a value from the current page and there
	// is no next page, then we're done iterating. Note that we must also check
	// the starting case to fetch the first page.
	if len(i.values) > 0 && i.index >= len(i.values)-1 && i.nextPageToken == "" {
		// No more values from the iterator
		return false
	}
```

Also I made an edit to above to make sure `Prev` works as expected. But now what happens if `Iter` returns nothing? I suppose in that case, the `i.index < len(i.values)` also guards against that; at the risk of multiple fetches for `Next()` calls that ignore `false`, which isn't normal.

pkg/gds/store/trtl/iterator.go
```diff
@@ -0,0 +1,281 @@
...
+
+	request := &trtlpb.IterRequest{
+		Namespace: i.namespace,
+	}
+	if i.index > 0 {
+		// We need to request the next page of values
+		request.Options = &trtlpb.Options{
+			PageToken: i.nextPageToken,
+		}
+	}
```
bbengfort, line 83: When I first read this, I assumed that this logic was checking if we'd already fetched the first page; but I think it's just guarding against adding a page token unnecessarily. However, in the first page case, the next page token should be empty, so the following should work, right?

```suggestion
	request := &trtlpb.IterRequest{
		Namespace: i.namespace,
		Options: &trtlpb.Options{
			PageToken: i.nextPageToken,		
		},
	}
```

pkg/gds/store/trtl/iterator.go
```diff
@@ -0,0 +1,281 @@
...
+
+	if reply, err = i.client.Iter(ctx, request); err != nil {
+		i.err = err
+		return false
+	}
+
+	// Add the new values to the slice
+	i.values = append(i.values, reply.Values...)
+	i.nextPageToken = reply.NextPageToken
+	return true
```
bbengfort, line 98: There is a slight edge case here where no new values are returned. E.g. because there is no snapshot isolation for paginated requests, the previous request could have had a next page token, but then everything on the next page was deleted before that page could be fetched. 

```suggestion
	return i.index < len(i.values)
```

pkg/gds/store/trtl/iterator.go
```diff
@@ -0,0 +1,281 @@
...
+	i.index--
+	if i.index < 0 {
+		i.index = 0
+		return false
+	}
+	return true
+}
+
+func (i *trtlBatchIterator) Seek(key []byte) bool {
+	// TODO: Figure out what to do about the batch seek method.
```
bbengfort, line 111: Let's make sure this TODO is noted in the follow on story.

pkg/gds/store/trtl/iterator.go
```diff
@@ -0,0 +1,281 @@
...
+		return false
+	}
+
+	if i.current != nil {
+		i.prev = i.current
+	}
+
+	i.current = val
+
+	return true
```
bbengfort, line 195: The `nil` check isn't necessary I don't think since we'd just be assigning `i.prev = nil`, right?

```suggestion
	// Bookkeeping to enable one `Prev` call 
	i.prev = i.current 
	i.current = val
	return true
```

bbengfort (approved):
@pdeziel thank you for making those changes! I had a couple of minor suggestions as I was working through the logic of the iterators - please read carefully, I'm not 100% sure if I was right or not, since I couldn't execute the tests in suggesting mode. 

We should also open a story for creating iterator tests for both Batch and Streaming, with the following cases:

- nothing is returned 
- 1 thing is returned
- 1 page is returned
- multiple pages are returned 

We'll want to look at `Next`, `Prev`, and `Seek` logic in those tests, since it is fairly advanced code that you've written here! Nicely done!

pkg/gds/store/trtl/iterator.go
pdeziel: Thanks for catching this, we definitely need dedicated iterator tests, especially since the batch iterator is currently uncovered.

PR #317, opened 2022-02-02
sc-2013 add code generation to workflow

pdeziel:
This adds a code generation step to the GitHub workflow to verify that the protocol buffers build correctly. This has a dependency on the trisacrypto/trisa repo so that repository is checked out to a separate directory before doing the generation.


bbengfort (approved):
Nice! At some point, it would be great if you can walk me through all of the incantations about working directory and what refs are what - but the build structure is pretty clear. Thank you for diving into this! s

PR #319, opened 2022-02-03
sc-3009 sectigo cert profiles tests

pdeziel:
This adds some tests for the certificate manager to verify that we are building the sectigo request parameters properly. In order to do this, some validation code was added to the sectigo mock which checks if the provided parameters match the required parameters for `ProfileCipherTraceEndEntityCertificate` and `ProfileCipherTraceEE`.

### DIFF
--- a/pkg/sectigo/mock/mock.go
+++ b/pkg/sectigo/mock/mock.go
@@ -246,15 +246,63 @@ func (s *Server) createSingleCertBatch(c *gin.Context) {
 		c.JSON(http.StatusBadRequest, err)
 		return
 	}
-	c.JSON(http.StatusOK, &sectigo.BatchResponse{
-		BatchID:      42,
-		CreationDate: time.Now().Format(time.RFC3339),
-		Status:       sectigo.BatchStatusReadyForDownload,
-		Active:       false,
-		BatchName:    in.BatchName,
-		OrderNumber:  23,
-		Profile:      "profile",
-	})
+
+	subjectParams := map[string]bool{
+		"organizationName":    false,
+		"localityName":        false,
+		"stateOrProvinceName": false,
+		"countryName":         false,
+	}
+	nameParams := map[string]bool{
+		"commonName":     false,
+		"dNSName":        false,
+		"pkcs12Password": false,
+	}
+
+	for k, v := range in.ProfileParams {
+		if v == "" {
+			c.JSON(http.StatusBadRequest, fmt.Errorf("empty profile parameter %s", k))
+			return
+		}
+
+		if _, ok := subjectParams[k]; ok {
+			subjectParams[k] = true
+		} else if _, ok := nameParams[k]; ok {
+			nameParams[k] = true
+		} else {
+			c.JSON(http.StatusBadRequest, fmt.Errorf("invalid profile parameter %s", k))
+			return
+		}
+	}
+
+	hasSubject, hasName := true, true
+	for _, ok := range subjectParams {
+		if !ok {
+			hasSubject = false
+		}
+	}
+
+	for _, ok := range nameParams {
+		if !ok {
+			hasName = false
+		}
+	}
+
+	// ProfileCipherTraceEndEntityCertificate: includes subject and name parameters
+	// ProfileCipherTraceEE: only includes name parameters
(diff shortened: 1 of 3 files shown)


pkg/sectigo/mock/mock.go
```diff
@@ -246,15 +246,63 @@ func (s *Server) createSingleCertBatch(c *gin.Context) {
...
+		} else {
+			c.JSON(http.StatusBadRequest, fmt.Errorf("invalid profile parameter %s", k))
+			return
+		}
+	}
+
+	hasSubject, hasName := true, true
+	for _, ok := range subjectParams {
+		if !ok {
+			hasSubject = false
```
bbengfort, line 281: You could also `break` here - no need to continue the loop once we've detected a missing subject param; same with the missing name param (though it doesn't really matter since the maps are so short).

pkg/sectigo/mock/mock.go
```diff
@@ -246,15 +246,63 @@ func (s *Server) createSingleCertBatch(c *gin.Context) {
...
+			CreationDate: time.Now().Format(time.RFC3339),
+			Status:       sectigo.BatchStatusReadyForDownload,
+			Active:       false,
+			BatchName:    in.BatchName,
+			OrderNumber:  23,
+			Profile:      "profile",
+		})
+	} else {
+		c.JSON(http.StatusBadRequest, fmt.Errorf("invalid profile parameters"))
+	}
```
bbengfort, line 305: One of the input parameters is `AuthorityID` - which relates directly to the certificate profiles, e.g. when you call `sectigo.UserAuthorities` you get back something like this: 

```json
[
  {
    "id": 7,
    "ecosystemId": 18,
    "signerCertificateId": 0,
    "ecosystemName": "Testing",
    "balance": 25,
    "enabled": true,
    "profileId": 7,
    "profileName": "CipherTrace EE"
  }
]
```

`submitCertificateRequest` is doing this on [line 136](https://github.com/trisacrypto/directory/blob/main/pkg/gds/certs.go#L136) and then passing the `7` along with the request. 

What I propose is that if `AuthorityID` == 7; we test ProfileCipherTraceEE and if `AuthorityID == 24` or something we test the other profile. That way we can ensure based on the request what profile is returned. We'll probably also want to edit the `userAuthorities` handler and the `profiles`, `profileParameters`, and `profileDetail` endpoints to respond correctly based on the profile configured in the Mock … which is also something the mock might need to do.

bbengfort (approved):
Nicely done! I downloaded the profile raw configurations and verified that you have included all of the parameters with respect to the profiles (more info on that in the story). This test nicely handles both cases for both profiles!

I started thinking about the logic of the hander - and I wanted to add that here for discussion.  One of the input parameters to create single cert batch is `AuthorityID` - which relates directly to the certificate profiles, e.g. when you call `sectigo.UserAuthorities` you get back something like this: 

```json
[
  {
    "id": 7,
    "ecosystemId": 18,
    "signerCertificateId": 0,
    "ecosystemName": "Testing",
    "balance": 25,
    "enabled": true,
    "profileId": 7,
    "profileName": "CipherTrace EE"
  }
]
```

`submitCertificateRequest` is doing this on [line 136](https://github.com/trisacrypto/directory/blob/main/pkg/gds/certs.go#L136) and then passing the `7` along with the request. 

What I propose is that if `AuthorityID` == 7; we test ProfileCipherTraceEE and if `AuthorityID == 24` or something we test the other profile. That way we can ensure based on the request what profile is returned. We'll probably also want to edit the `userAuthorities` handler and the `profiles`, `profileParameters`, and `profileDetail` endpoints to respond correctly based on the profile configured in the Mock … which is also something the mock might need to do. 

That all is more extensive mocking of Sectigo than this story considered. Do you think we should create a follow on story to have the Mock differentiate logic based on a configured profile setting?

pkg/sectigo/mock/mock.go
pdeziel: That makes sense, I was looking for something like that to make the mock handling logic easier but I couldn't find which Sectigo endpoint was doing it. I think we should create another story to configure the profile in the mock like you mentioned.

PR #324, opened 2022-02-10
sc-2465 trtl BackupManager

pdeziel:
This adds a BackupManager to trtl in the same vein as the GDS levelDB one. Most of the code is ported over, although I did add a common utility function that both use to simplify the code a bit.


pkg/trtl/server.go
```diff
@@ -193,6 +204,14 @@ func (t *Server) Shutdown() (err error) {
 	log.Info().Msg("gracefully shutting down trtl server")
 	t.srv.GracefulStop()
 
+	// Shutdown the backup manager
+	if t.conf.Backup.Enabled {
+		if err = t.backup.Shutdown(); err != nil {
+			log.Error().Err(err).Msg("could not shutdown backup manager")
+			errs = append(errs, err)
+		}
+	}
```
pdeziel, line 213: Do we have to add more concurrency control here? This is signaling the BackupManager to shutdown but I don't think it's waiting for it to finish up, which could cause problems (although unlikely) if the BackupManager is in the middle of doing a backup and the DB gets closed out from under it.

pkg/trtl/backup.go
```diff
@@ -0,0 +1,190 @@
...
+			log.Warn().Msg("trtl backup manager received stop signal")
+			return
+		default:
+		}
+	}
+}
+
+func (m *BackupManager) Shutdown() error {
+	if m.conf.Backup.Enabled && m.stop != nil {
+		close(m.stop)
```
bbengfort, line 98: When we call `Shutdown` shouldn't we send a stop signal rather than close the channel? I'm sorry, I'm just not remembering what we intended for this - was it primarily to stop tests?

bbengfort (approved):
@pdeziel looks good and it was a good chance for me to review the back up manager code! A couple of things came up for discussion, but I wanted to approve it anyway so that you could address them and move forward. Thanks for taking this one on this sprint!

pkg/trtl/backup_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	honuldb "github.com/rotationalio/honu/engines/leveldb"
+	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/trisacrypto/directory/pkg/trtl"
+	"github.com/trisacrypto/directory/pkg/trtl/config"
+	"github.com/trisacrypto/directory/pkg/utils"
+)
+
+// Test that the backup manager periodically creates backups.
+func (s *trtlTestSuite) TestBackupManager() {
```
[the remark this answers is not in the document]
pdeziel: In leveldb we also had a test to make sure the BackupManager doesn't do anything if it's not enabled; however in trtl that logic exists in `trtl.New()` so I've omitted that test for now.

PR #326, opened 2022-02-10
fix issues on status component of dashboard

masskoder:


bbengfort (approved):
Sorry about all the hassle with signed commits - but now that's out of the way we should be good to go for the future! 

It appears that the linter fixes are missing from this commit; but since you mentioned you'll be preparing a GitHub action for that, I don't think it's a big deal.

PR #329, opened 2022-02-10
sc-3104 github action for web/node tests

masskoder:


.github/workflows/tests.yaml
```diff
@@ -75,7 +75,102 @@ jobs:
...
+          cache-dependency-path: ./web/gds-admin-ui/yarn.lock
+
+      - name: Install Dependencies
+        working-directory: ./web/gds-admin-ui
+        run: yarn 
+
+      - name: Run Tests
+        working-directory: ./web/gds-admin-ui
+        env: 
+          #maybe we can make it dynamic like ${{secrets.REACT_APP_GDS_API_ENDPOINT}}
```
bbengfort, line 103: We can if you'd like - it's not exactly a secret, just a test configuration. If you feel it's hidden in the actions configuration, we could make these values the defaults in the code if you'd prefer.

.github/workflows/tests.yaml
```diff
@@ -75,7 +75,102 @@ jobs:
           # Make sure to add to GitHub secrets!
           token: ${{secrets.CODECOV_TOKEN}}
           file: ./directory/coverage.txt
+ 
+  gds_admin_test:
+    name: GDS Admin UI Tests
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [14.x, 16.x]
```
bbengfort, line 84: You have a matrix for node-version here, however, you're only using node-version 16.8 below in the Setup Node step. I think we could probably skip the matrix and only test the node version that we're using in our container builds, `BUILDER_IMAGE=node:16` -- is there a way we can specify the latest 16 version to catch any changes that might occur?

.github/workflows/tests.yaml
```diff
@@ -75,7 +75,102 @@ jobs:
...
+        with:
+          # Make sure to add to GitHub secrets!
+          token: ${{secrets.CODECOV_TOKEN}}
+  
+  gds_ui_test:
+    name: GDS UI Tests
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [14.x, 16.x]
```
bbengfort, line 119: See matrix comment above

.github/workflows/tests.yaml
```diff
@@ -75,7 +75,102 @@ jobs:
...
+          REACT_APP_GDS_IS_TESTNET : true
+        run: yarn test
+
+      - name: Upload Coverage report to CodeCov
+        uses: codecov/codecov-action@v1.0.0
+        with:
+          # Make sure to add to GitHub secrets!
+          token: ${{secrets.CODECOV_TOKEN}}
+
+  lingui_gds_ui:
```
bbengfort, line 149: So the point of this test is to ensure the messages for i18n are correct and that lingui builds correctly right? What would you think about adding rumi to this so that we have a report of our translation coverage? 

https://github.com/rotationalio/rumi

bbengfort (approved):
Looks good! Just some minor changes related to the matrix and node version; but after those changes are made this should be good to go!

PR #332, opened 2022-02-11
Port index handlers to their own package

bbengfort:
This PR ports the index functionality from the `leveldb` store to its own package in order to have the indices used by both the `trtl` and the `leveldb` stores. The amount of code was doubled in the port, but most of that is related to the interfaces created to allow the packages to import the indices successfully, as well as more extensive tests to ensure no bugs occurred in the transfer of the implementation. 

The code ports the code from `pkg/gds/store/leveldb/index.go` to `pkg/gds/store/index` 

In the original code there were three types of indices:

- `uniqueIndex` - maps a unique value to a UUID
- `containerIndex` - maps a unique value to a list of UUIDs
- `sequence` - a utility for auto-incrementing ids (not used anymore)

These were all stored in the same file, `index.go` so to organize things better, I moved them to `unique.go`, `container.go`, and `sequence.go` along with their tests to individual files. 

The index.go functionality also contained normalizers - functions that performed string processing to make sure that whitespace was removed and that the search terms were case-insensitive. There were three normalizers:

- Normalizer: strip space and lower case
- NormalizeURL: extract just the hostname of a valid URL
- NormalizeCountry: convert a country name to its ISO Alpha-2 code 

I moved these to `search.go` along with the `ParseQuery` function since they are intended to facilitate search behavior via deduplication. 

In order to unify the code between trtl and leveldb, I created type-specific indices for names (VASP company names), websites, countries, and categories. These can be found in `index.go` and tested as extensively as I can in `index_test.go`. This required utilities to combine a normalizer and search functionality with the correct index; hence `normalizeUnique`, and `normalizeContainer`. I've also created interfaces `Index`, `Serializer`, `SingleIndex` and `MultiIndex` to simplify the code. 

The `ExactMatch`, `PrefixMatch`, and `ContainsRecords` methods are factories that implement the types of search required by the leveldb package. These are also pretty gnarly, but I've tried to test them as best I could. 

In the end, the index functionality requires a drastic overhaul; but hopefully this intermediate step will move us in the right direction.


pkg/gds/store/index/container.go
```diff
@@ -0,0 +1,198 @@
...
+	"compress/gzip"
+	"encoding/json"
+	"fmt"
+	"sort"
+)
+
+// Container indices map a value to a set of values, most commonly an index to a set of
+// record IDs that are related to that value. The container index is maintained in
+// sorted order with no duplicates to save space and enhance ordered search.
+type Container map[string][]string
```
bbengfort, line 14: Moved to its own file to help with organization.

pkg/gds/store/index/container.go
```diff
@@ -0,0 +1,198 @@
...
+
+	decoder := json.NewDecoder(gz)
+	if err = decoder.Decode(&c); err != nil {
+		return fmt.Errorf("could not decode index: %s", err)
+	}
+	return nil
+}
+
+// Normalized container maintains the normalizer with the container index.
+type normalizedContainer struct {
```
bbengfort, line 154: Added to unify functionality between trtl and leveldb

pkg/gds/store/index/container_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+import (
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+	"github.com/trisacrypto/directory/pkg/gds/store/index"
+)
+
+func TestContainerIndex(t *testing.T) {
```
bbengfort, line 11: These tests are moved from the original pacakge

pkg/gds/store/index/index.go
```diff
@@ -0,0 +1,83 @@
...
+	Contains(key, value string) bool
+}
+
+// Serializer allows indices to be loaded and dumped to disk.
+type Serializer interface {
+	Load([]byte) error
+	Dump() ([]byte, error)
+}
+
+func NewNamesIndex() SingleIndex {
```
bbengfort, line 45: These named indexes are new; designed so that both trtl and leveldb would have the same functionality.

pkg/gds/store/index/index_test.go
```diff
@@ -0,0 +1,230 @@
+package index_test
```
bbengfort, line 1: All of these tests are new

pkg/gds/store/index/search.go
```diff
@@ -0,0 +1,169 @@
...
+	"strings"
+
+	"github.com/rs/zerolog/log"
+	"github.com/trisacrypto/trisa/pkg/iso3166"
+)
+
+const searchPrefixMinLength = 3
+
+// Normalize specifies how to modify a string value in preparation for search.
+type Normalizer func(s string) string
```
bbengfort, line 15: Normalize functions are moved from the original package.

pkg/gds/store/index/search.go
```diff
@@ -0,0 +1,169 @@
...
+	if i < len(arr) && arr[i] == item {
+		// value is already in the array
+		return arr
+	}
+
+	arr = append(arr, "")
+	copy(arr[i+1:], arr[i:])
+	arr[i] = item
+	return arr
+}
```
bbengfort, line 169: This is new - it ensures that search results are returned without duplicates and sorted.

pkg/gds/store/index/search.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+}
+
+// Queries are maps that hold an index name (e.g. "name" or "country") and map it to an
+// indexable key in the index. The query can be either a single string or a list of
+// strings; the parse function extracts the appropriate type and returns a list of
+// strings if the query for the particular index name exists.
+//
+// All values are normalized by the normalize function if provided.
+func ParseQuery(index string, query map[string]interface{}, norm Normalizer) ([]string, bool) {
```
bbengfort, line 123: Ported from the original.

pkg/gds/store/index/search_test.go
```diff
@@ -0,0 +1,205 @@
+package index_test
```
bbengfort, line 1: these tests are all new.

pkg/gds/store/index/sequence.go
```diff
@@ -0,0 +1,29 @@
+package index
```
bbengfort, line 1: Ported from the original; not in use anymore.

pkg/gds/store/index/sequence_test.go
```diff
@@ -0,0 +1,47 @@
+package index_test
```
bbengfort, line 1: Ported from the original

pkg/gds/store/index/unique.go
```diff
@@ -0,0 +1,164 @@
...
+
+	decoder := json.NewDecoder(gz)
+	if err = decoder.Decode(&c); err != nil {
+		return fmt.Errorf("could not decode index: %s", err)
+	}
+	return nil
+}
+
+// Normalized unique maintains the normalizer with the unique index.
+type normalizedUnique struct {
```
bbengfort, line 120: Added to unify functionality between trtl and leveldb

pkg/gds/store/index/unique_test.go
```diff
@@ -0,0 +1,97 @@
+package index_test
```
bbengfort, line 1: all of these tests were ported from the original

pkg/gds/store/index/search.go
pdeziel: Perhaps in a future refactor we can make this a utility function? In the container we have some similar code that's also doing an "insort" type operation.

pdeziel (approved):
The indexing logic looks good to me and it seems like we have a decent amount of test coverage. I definitely agree that there is room for refactoring but I like the shared interface and I think this is a good start.

pkg/gds/store/index/search.go
bbengfort: That's a good idea - though a utility function would have to be used with care. Slice operations are notoriously difficult because the underlying array is immutable (in fact, even with this function I went through several revisions with pointer slices and different types of addressing - I opted for likely the least performant version, but the most understandable). 

A slice is really just a low level pointer (really a header with a length, capacity, and pointer to a location in memory). When the slice is changed (e.g. if an append or a copy exceeds the capacity of the underlying array) a new array is allocated and the slice variable is modified. That means that a function cannot modify a slice in place (and you get a lot of weird bugs with empty slices). In this function, by returning the modified slice, we're guaranteeing the caller gets access to it, but at the cost of many escapes (because the array is in the heap, not the stack). 

Another option is to keep a sorted array type around or something like that.

PR #335, opened 2022-02-14
Fix docker compose for grpc-web

bbengfort:
This is a fix for the `"unknown service /trisa.gds.api.v1beta1.DirectoryService"` error when running the GDS UI with grpc-web in docker compose. The problem was a trailing slash in the `REACT_APP_GDS_API_ENDPOINT: http://localhost:8080/` configuration - no trailing slashes allowed!

Thanks to: https://stackoverflow.com/questions/54994392/how-do-i-set-up-an-envoy-proxy-in-minikube-for-a-grpc-service for helping me finally unlock things. 

In my attempt to fix things, I think I've ended up with a newer (maybe) envoy config and docker image. I still need to verify that this will work on k8s.

### DIFF
--- a/containers/docker-compose.yaml
+++ b/containers/docker-compose.yaml
@@ -71,6 +71,7 @@ services:
     ports:
       - 8080:8080
     profiles:
+      - proxy
       - api
       - all
 
@@ -79,7 +80,7 @@ services:
       context: ../
       dockerfile: ./containers/gds-ui/Dockerfile
       args:
-        REACT_APP_GDS_API_ENDPOINT: http://localhost:8080/
+        REACT_APP_GDS_API_ENDPOINT: http://localhost:8080
         REACT_APP_GDS_IS_TESTNET: "true"
         REACT_APP_ANALYTICS_ID: ${REACT_APP_ANALYTICS_ID}
     image: trisa/gds-ui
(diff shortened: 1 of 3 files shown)


containers/docker-compose.yaml
```diff
@@ -79,7 +80,7 @@ services:
       context: ../
       dockerfile: ./containers/gds-ui/Dockerfile
       args:
-        REACT_APP_GDS_API_ENDPOINT: http://localhost:8080/
+        REACT_APP_GDS_API_ENDPOINT: http://localhost:8080
```
bbengfort, line 83: 😵‍💫

containers/grpc-proxy/Dockerfile
```diff
@@ -1,2 +1,2 @@
-FROM envoyproxy/envoy-dev:933e67a9ecac814839f514fb5c31c78f3d33e5fe
```
bbengfort, line 1: The old docker image (`933e67a`) was pushed a year ago, whereas the newer docker image (`cc089aa`) was pushed 18 hours ago and is the image listed in their [docker-compose documentation](https://www.envoyproxy.io/docs/envoy/latest/start/docker) (they must have a documentation and image build system thats coordinated, which is pretty cool. 

It's probably better to go with the newer docker image because it's probably a later version of Envoy (though I couldn't tell you which version just by looking at the Dockerfile) - later versions tend to have security updates. Unfortunately, the old configuration didn't work with the newer image, hence the change to the `envoy.yaml` configuration below. This works on docker compose, but it is untested on k8s; so when we deploy we need to check to make sure the proxy is working and if it doesn't we can roll back to the old config/image.

containers/docker-compose.yaml
rebeccabilbro: 🙃

PR #336, opened 2022-02-14
sc-3354 sectigo mock verification

pdeziel:
This makes the Sectigo mock a bit smarter by adding known profiles that it checks against for requests. The mock now takes a parameter on `Start` which is the Sectigo profile name to use for the logic in the mock handlers.

### DIFF
--- a/pkg/gds/certs_test.go
+++ b/pkg/gds/certs_test.go
@@ -21,7 +21,7 @@ import (
 // Test that the certificate manger correctly moves certificates across the request
 // pipeline.
 func (s *gdsTestSuite) TestCertManager() {
-	s.setupCertManager()
+	s.setupCertManager(sectigo.ProfileCipherTraceEE)
 	defer s.teardownCertManager()
 	require := s.Require()
 
@@ -139,7 +139,7 @@ func (s *gdsTestSuite) TestCertManager() {
 
 // Test that the certificate manager is able to process an end entity profile.
 func (s *gdsTestSuite) TestCertManagerEndEntityProfile() {
-	s.setupCertManager()
+	s.setupCertManager(sectigo.ProfileCipherTraceEndEntityCertificate)
 	defer s.teardownCertManager()
 	defer s.loadReferenceFixtures()
 	require := s.Require()
@@ -180,7 +180,7 @@ func (s *gdsTestSuite) TestCertManagerEndEntityProfile() {
 
 // Test that the certificate manager is able to process a CipherTraceEE profile.
 func (s *gdsTestSuite) TestCertManagerCipherTraceEEProfile() {
-	s.setupCertManager()
+	s.setupCertManager(sectigo.ProfileCipherTraceEE)
 	defer s.teardownCertManager()
 	defer s.loadReferenceFixtures()
 	require := s.Require()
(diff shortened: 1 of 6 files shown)


pkg/sectigo/mock/mock.go
```diff
@@ -30,31 +31,84 @@ import (
...
+
+	// TODO: We are only loading one authority so that we can get consistent behavior
+	// from the tests.
+	authority := 1
+	for _, p := range validProfiles {
+		mockServer.profileLookup[p.ID] = p
+		if p.Name == profile {
+			mockServer.authorityLookup[authority] = p
+		}
+		authority++
```
pdeziel, line 111: This got a little hairy, originally I wanted to have the mock know about both authorities but I had forgotten that the cert manager code is just taking the first authority available and the profile ID from that, which made the tests unreliable. So, now `Start()` takes in a profile name and only loads one authority into the lookup map, which allows the tests to be able to configure which profile they want to test against.

pkg/gds/certs_test.go
```diff
@@ -492,11 +514,12 @@ func (s *gdsTestSuite) TestProcessBatchNoSuccess() {
 	require.Equal("automated", cert.AuditLog[3].Source)
 }
 
-func (s *gdsTestSuite) setupCertManager() {
+func (s *gdsTestSuite) setupCertManager(profile string) {
```
bbengfort, line 517: I like this - it makes it much simpler to test the two different profiles!

pkg/sectigo/mock/mock.go
```diff
@@ -30,31 +31,84 @@ import (
...
+	"pkcs12Password",
+}
+
+// All the profiles that the mock knows about.
+var validProfiles = []Profile{
+	{
+		ID:      sectigo.ProfileIDCipherTraceEE,
+		Name:    sectigo.ProfileCipherTraceEE,
+		Balance: 50,
+		Params:  nameParams[:],
```
bbengfort, line 60: Thanks for using the fixed array and converting it to a slice here; I think that's the better style for array "constants".

pkg/sectigo/endpoints_test.go
```diff
@@ -39,7 +39,7 @@ func TestModifyBaseURL(t *testing.T) {
 	require.Equal(t, "https://iot.sectigo.com/api/v1/certificates/find", ep.String())
 
 	// Create a mock Server
-	require.NoError(t, mock.Start())
+	require.NoError(t, mock.Start(""))
```
bbengfort, line 42: I may find this out later on - but is there a default if you pass empty string or does it just work with an empty profile? If there is a default we may just want to add a comment here.

pkg/sectigo/mock/mock.go
bbengfort: That makes sense and is closer to how Sectigo behaves anyway.

bbengfort: So to answer my earlier question, if empty string is passed in, then `profileLookup` and `authorityLookup` will be empty, therefore the mock will always return 404s on those endpoints?

bbengfort (approved):

pkg/sectigo/mock/mock.go
pdeziel: Exactly

PR #338, opened 2022-02-15
Increase expiration time for cookies

rebeccabilbro:
This PR attempts to resolve the "no csrf token cookie in request" warning message. After a bit of research, our best guess is that it is an expiration issue, so we are slightly increasing the time allotment to see if it has an impact.


bbengfort (approved):
🕐

PR #339, opened 2022-02-15
sc-3472 storage for mock emails

pdeziel:
This adds support for writing the mock emails to a directory. The functionality is behind the `GDS_EMAIL_STORAGE` variable and is only valid if `GDS_EMAIL_TESTING` is also enabled. This enables emails to be manually verified in QA. In order to make it easier to distinguish emails, each email is written to a subdirectory with the name of the recipient and a unique filename which includes the timestamp.


rebeccabilbro (approved):
Looks good to me @pdeziel; this will definitely help with our smoke testing process!

PR #342, opened 2022-02-16
Finalize Trtl/GDS Integration

bbengfort:


containers/docker-compose.yaml
```diff
@@ -20,8 +22,11 @@ services:
       - GDS_SECRET_KEY=supersecretsquirrel
       - GDS_LOG_LEVEL=debug
       - GDS_CONSOLE_LOG=true
-      - GDS_DATABASE_URL=leveldb:////data/db
+      - GDS_DATABASE_URL=trtl://trtl:4436/
```
bbengfort, line 25: works with trailing slash, doesn't work without :-\

containers/docker-compose.yaml
```diff
@@ -52,7 +57,7 @@ services:
       - GDS_ADMIN_REVIEW_URL=http://localhost:3001/vasps/
       - GDS_CERTMAN_INTERVAL=5m
       - GDS_CERTMAN_STORAGE=/data/certs
-      - GDS_BACKUP_ENABLED=true
+      - GDS_BACKUP_ENABLED=false
```
bbengfort, line 60: when configured with trtl, this needs to be false otherwise GDS will exit FATAL (can't backup)

PR #349, opened 2022-02-21
sc-3100 initial trtl grafana dashboard

pdeziel:
This adds a trtl grafana dashboard, which currently displays statistics about Gets/Puts/Deletes/Iters. In a future story we should add prometheus metrics for tracking Cursors so that we can display those as well. Here is a screenshot of the dashboard as it rendered on my end:
Screen Shot 2022-02-21 at 2 26 16 PM

### DIFF
--- a/containers/monitor/.gitignore
+++ b/containers/monitor/.gitignore
@@ -1,1 +1,2 @@
 grafana/*
+!grafana/dashboards/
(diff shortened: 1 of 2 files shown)


bbengfort (approved):
Looks great!

PR #351, opened 2022-02-22
Configuration via StatefulSet Replica Strategies

bbengfort:
This PR implements alternative configuration strategies for StatefulSets in Kubernetes:

- `HostnamePID`: parses a hostname in the form of `pod-n` to update the name and PID from the hostname that k8s configures on the pod. For example if the environment configuration is: `PID=40` `Name="foo"` and the hostname is `trtl-4` then this strategy will change the PID to 44 and the name to `trtl-44`; defaulting to the k8s StatefulSet configuration but still allowing for globally unique ocnfiguration. 
- `FilePID`: reads a PID from a file on disk, e.g. `/var/run/replica.pid` - this is kind of old school, but I felt like it might help with testing and debugging.
- `JSONConfig`: updates the configuration based on a JSON file that maps PID to the rest of the config. The idea here is that the `StatefulSet` can't configure different pods, but we can load a `ConfigMap` that has all the configurations. If the `HostnamePID` strategy comes first, then the pod can lookup its configuration in the `ConfigMap`.

These strategies are configured from the environment (sheesh 🐢  all the way down) - so if the following config is in the statefulset:

```
TRTL_REPLICA_STRATEGY_HOSTNAME_PID=true
TRTL_REPLICA_STRATEGY_JSON_CONFIG=/data/replicas.json
```

Then Trtl will first use the `HostnamePID` strategy then the `JSONConfig` strategy. The hardcoded order of strategies is HostnamePID, FilePID, JSONConfig. We can add other configs, e.g. `DatabaseConfig` or `KahuConfig` in the future; I've designed the strategies to be flexible; though specifying the strategies from the environment is tricky; but we can tackle that later.


pkg/trtl/config/config.go
```diff
@@ -36,12 +37,19 @@ type DatabaseConfig struct {
...
+	GossipInterval time.Duration `split_words:"true" default:"1m" json:"gossip_interval"`
+	GossipSigma    time.Duration `split_words:"true" default:"5s" json:"gossip_sigma"`
+}
+
+type ReplicaStrategyConfig struct {
+	HostnamePID bool   `split_words:"true" default:"false"` // Set to true to use HostnamePID
+	Hostname    string `envconfig:"TRTL_REPLICA_HOSTNAME"`  // configure the hostname from the environment
+	FilePID     string `split_words:"true"`                 // Set to the PID filename path to use FilePID
+	JSONConfig  string `split_words:"true"`                 // Set to the config map path to use JSONConfig
 }
```
bbengfort, line 53: I don't like this, but it's the quickest way to define a strategy. In the future, I hope we can have "strategy names" e.g. "k8s", "metal", etc. that automatically configure different approaches.

pkg/trtl/config/config.go
```diff
@@ -131,6 +144,24 @@ func (c *ReplicaConfig) Validate() error {
...
+	if c.FilePID != "" {
+		strategies = append(strategies, FilePID(c.FilePID))
+	}
+
+	if c.JSONConfig != "" {
+		strategies = append(strategies, JSONConfig(c.JSONConfig))
+	}
+
+	return strategies
+}
```
bbengfort, line 163: Note the fixed ordering of strategies - it's not very flexible, but punting this for the future.

pkg/trtl/config/config_test.go
```diff
@@ -23,10 +25,20 @@ var testEnv = map[string]string{
...
+	"TRTL_BACKUP_INTERVAL":          "1h",
+	"TRTL_BACKUP_STORAGE":           "fixtures/backups",
+	"TRTL_BACKUP_KEEP":              "7",
+}
+
+var strategyEnv = map[string]string{
+	"TRTL_REPLICA_STRATEGY_HOSTNAME_PID": "true",
+	"TRTL_REPLICA_HOSTNAME":              "pizza-36",
+	"TRTL_REPLICA_STRATEGY_FILE_PID":     "",
+	"TRTL_REPLICA_STRATEGY_JSON_CONFIG":  "testdata/replicas.json",
```
bbengfort, line 41: Separated this from the main tests since it's just a k8s thing. Everything should work without this.

pkg/trtl/config/testdata/replicas.json
```diff
@@ -0,0 +1,34 @@
+{
+    "40": {
+        "enabled": true,
+        "pid": 40,
+        "region": "brooklyn",
+        "name": "raphael",
+        "gossip_interval": 1200000000000,
```
bbengfort, line 7: Unfortunately Go `time.Duration` serializes as number of nanoseconds by default and it won't be fixed until Go 2.0 - https://github.com/golang/go/issues/10275

I could have written a custom type that deserialized a string such as "2m"  - but this type would not have played well easily with envconfig.

proto/trtl/v1/trtl.proto
```diff
@@ -130,6 +130,16 @@ message ServerStatus {
...
+}
+
+message ReplicaStatus {
+    bool enabled = 1;
+    uint64 pid = 2;
+    string region = 3;
+    string name = 4;
+    string interval = 5;
+    string sigma = 6;
 }
```
bbengfort, line 143: Added to have visibility into what the configuration is on k8s -- I can remove if you don't like it, but needed it for testing the deployment.

containers/trtl-init/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+
+if [[ -f $CERT && -f $CHAIN ]]; then
+    cp $CERT /data/certs/mtls_cert.pem
+    cp $CHAIN /data/certs/mtls_chain.pem
+    echo "mtls certificates configured"
+    exit 0
+else
+    echo "could not find certificate and/or chain for configuration"
+    exit 1
+fi
```
bbengfort, line 17: This script runs before a container is booted in an [init container](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) and copies the correct certificates, e.g. `ninja-1` certs into the `/data/certs` volume so that the pods only have access to their correct certs. Unfortunately it means the init-container has access to all the certs, but there isn't any getting around that.

pkg/trtl/config/config.go
```diff
@@ -36,12 +37,19 @@ type DatabaseConfig struct {
...
+	Enabled        bool          `split_words:"true" default:"true" json:"enabled"`
+	PID            uint64        `split_words:"true" required:"false" json:"pid"`
+	Region         string        `split_words:"true" required:"false" json:"region"`
+	Name           string        `split_words:"true" required:"false" json:"name"`
+	GossipInterval time.Duration `split_words:"true" default:"1m" json:"gossip_interval"`
+	GossipSigma    time.Duration `split_words:"true" default:"5s" json:"gossip_sigma"`
+}
+
+type ReplicaStrategyConfig struct {
+	HostnamePID bool   `split_words:"true" default:"false"` // Set to true to use HostnamePID
```
rebeccabilbro, line 49: Should we consider making this default to true given that the present approach will default to using this strategy ?

pkg/trtl/config/replica_test.go
```diff
@@ -0,0 +1,213 @@
...
+
+func TestReplicaConfigureNoModification(t *testing.T) {
+	s1 := makeRandomTestStrategy()
+	s2 := makeRandomTestStrategy()
+
+	conf := config.ReplicaConfig{
+		Enabled:        true,
+		PID:            53,
+		Region:         "ibelin",
+		Name:           "balin",
```
rebeccabilbro, line 23: i think it's "balian" 😉

proto/trtl/v1/trtl.proto
rebeccabilbro: Happy to have you leave this in; might be useful for configuration diagnostics/tuning in the future?

rebeccabilbro (approved):
Thank you @bbengfort for your research into configuration options that will allow replicas to consistently address each other by name! I know it took a bit of experimentation, but I'm confident that we'll be glad we did it since this feature is pretty foundation to well, you know, building distributed systems  😄 

The PR looks good; just checking -- might we also want to update anything in [`trtl/main`](https://github.com/trisacrypto/directory/blob/main/cmd/trtl/main.go) or our [env template](https://github.com/trisacrypto/directory/blob/main/.env.template)?

pkg/trtl/config/replica_test.go
bbengfort: Hehe it definitely is - darn you cold fog brain!

pkg/trtl/config/config.go
bbengfort: In a replicated trtl deployed as a StatefulSet, this has to be true - but for local development, single node trtl, and Docker compose this has to be false. I figured that right now at least, the last three use cases are more routine/normal.

PR #353, opened 2022-02-24
Trtl benchmark testing

rebeccabilbro:
This PR introduces benchmark testing for the Trtl RPC methods. 

Still to do:
- export data from each benchmark
- do some kind of data visualization to compare to honu/leveldb


pkg/trtl/bench_test.go
```diff
@@ -0,0 +1,347 @@
...
+	ctx := context.Background()
+
+	// Manually add some fixtures
+	var err error
+	for i := 0; i < b.N; i++ {
+		_, err = cc.Put(ctx, &pb.PutRequest{
+			Key:   []byte(fmt.Sprintf("t%d", i)),
+			Value: []byte(strconv.Itoa(i)),
+		})
+	}
```
bbengfort, line 200: I'm not 100% sure this will work - I think that `b.N` will change during the benchmark depending on how long the op takes. I recommend using stop and start timer in the benchmark loop as follows:

```go
	// Reset the timer to focus only on the Delete
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		// Stop timer and put key/value to delete 
		b.StopTimer()
		_, err = cc.Put(ctx, &pb.PutRequest{
			Key:   []byte(fmt.Sprintf("t%d", i)),
			Value: []byte(strconv.Itoa(i)),
		})
		require.NoError(b, err)
		
		// Start timer and perform the Delete 
		b.StartTimer()
                 _, err = cc.Delete(ctx, &pb.DeleteRequest{
			Key: []byte(fmt.Sprintf("t%d", i)),
		})		
	}
```

pkg/trtl/bench_test.go
```diff
@@ -0,0 +1,347 @@
...
+
+	// Start the gRPC client.
+	require.NoError(b, bench.grpc.Connect())
+	defer bench.grpc.Close()
+	cc := pb.NewTrtlClient(bench.grpc.Conn)
+	ctx := context.Background()
+
+	// Manually add some fixtures
+	var err error
+	for i := 0; i < b.N; i++ {
```
bbengfort, line 230: Recommend:

```go
for i:= 0; i < 1000; i++
```

1000 key/value pairs feels like a good number for the iter benchmark; it's not an insignificant number of iterations  (and is above the page size) but it's not so big that it will artificially delay the request.

pkg/trtl/bench_test.go
```diff
@@ -0,0 +1,347 @@
...
+
+	// Start the gRPC client.
+	require.NoError(b, bench.grpc.Connect())
+	defer bench.grpc.Close()
+	cc := pb.NewTrtlClient(bench.grpc.Conn)
+	ctx := context.Background()
+
+	// Manually add some fixtures
+	var err error
+	for i := 0; i < 3; i++ {
```
bbengfort, line 267: Recommend making this 1000 terminators (😨 )

bbengfort (approved):
Looks great! These benchmarks are well written and target exactly the operation that we're looking to measure. The setup and cleanup methods are compact and effective, making them easy to maintain and add new benchmarks. I especially like the batch request benchmark because that will benchmark our versioning system, a nice contrast to the Put that will add different keys. 

I also appreciate the Terminator references!

PR #355, opened 2022-02-28
Adds certificate expiration to admin list API

bbengfort:
CipherTrace wants to know when certificates are expiring, so I've added the expiration to the admin list API.

### DIFF
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -794,6 +794,7 @@ func (s *Admin) ListVASPs(c *gin.Context) {
 			// Add certificate serial number if it exists
 			if vasp.IdentityCertificate != nil {
 				snippet.CertificateSerial = fmt.Sprintf("%X", vasp.IdentityCertificate.SerialNumber)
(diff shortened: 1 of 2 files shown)


pkg/gds/admin.go
```diff
@@ -794,6 +794,7 @@ func (s *Admin) ListVASPs(c *gin.Context) {
 			// Add certificate serial number if it exists
 			if vasp.IdentityCertificate != nil {
 				snippet.CertificateSerial = fmt.Sprintf("%X", vasp.IdentityCertificate.SerialNumber)
+				snippet.CertificateExpiration = vasp.IdentityCertificate.NotAfter
```
pdeziel, line 797: I don't remember what we're checking in the tests, is it easy to add a check in the test to verify that we are populating this value?

bbengfort: Unfortunately it's not easy - we don't have certificate data in our fixtures to test this (or at least in the small fixtures).

PR #359, opened 2022-03-01
Trtl Anti-Entropy refinement

rebeccabilbro:
In the interest of assisting with some anti entropy debugging, this PR adds new sync message for each case inside sync.go as well as version logging and a "perspective" parameter for version, update, and repair metrics in prometheus.


bbengfort (approved):
Looks good! Let's get this merged and I can run the simulator on shellback for an hour or two just to make sure we're getting the metrics we need.

PR #362, opened 2022-03-01
sc-3672 Integrate trtl with GDS test suite

pdeziel:
This adds trtl to the GDS test suite. The direction I decided to try with this is to have the suite support either trtl or leveldb, which is configured with the `stype` parameter. So currently the GDS test suite is run twice, once with trtl and once with leveldb. It's possible to run them independently by referencing the specific test suite (e.g., `go test ^TestGDSLevelDB$` or `go test ^TestGDSTrtl$`). There are a few things that I think we could improve but I wanted to checkpoint here since the PR is already quite large; maybe we should create separate stories for these:

- Add flag to make it easier to run either the LevelDB test suite or the Trtl test suite
- Perhaps these should be separate test suites so that we don't have `stype` checks everywhere and we could easily deprecate the LevelDB store if we decide to remove LevelDB from GDS in the future
- On my machine the entire GDS test suite takes about 18 seconds to run, are we okay with this?
- We have a lot of log output when we run the tests


pkg/gds/service_test.go
```diff
@@ -38,16 +40,14 @@ const (
 )
 
 var (
-	update             = flag.Bool("update", false, "update the gzipped test databases")
-	fakesFixturePath   = filepath.Join("testdata", "fakes.tgz")
-	fullDBFixturePath  = filepath.Join("testdata", "db.tgz")
-	smallDBFixturePath = filepath.Join("testdata", "smalldb.tgz")
-	smallDBSubset      = map[string]map[string]struct{}{
+	fakesFixturePath = filepath.Join("testdata", "fakes.tgz")
+	smallDBSubset    = map[string]map[string]struct{}{
```
pdeziel, line 44: Removed the multiple fixture archives in favor of at-test-time DB generation.

pkg/gds/service_test.go
```diff
@@ -83,25 +85,19 @@ func (s *gdsTestSuite) ResetConfig() {
...
-
 	// Load the reference fixtures into memory for use in testing
 	s.loadReferenceFixtures()
 
-	// Generate the databases if required (depends on the loaded reference fixtures)
-	if *update || !pathExists(fullDBFixturePath) || !pathExists(smallDBFixturePath) {
-		s.generateDB()
+	// Start the Trtl server if required
+	if s.stype == storeTrtl {
+		s.SetupTrtl()
```
pdeziel, line 95: `SetupTrtl` has to be called before `generateDB` because we use the GDS storage interface `CreateVASP`, `UpdateCertReq` to generate the fixtures in the database, and Trtl has to be running for those functions to work.

pkg/gds/service_test.go
```diff
@@ -255,10 +345,13 @@ func (s *gdsTestSuite) CompareFixture(namespace, key string, obj interface{}, re
...
 	case vasps:
 		var a *pb.VASP
+		var data pb.VASP
 		for _, f := range s.fixtures[namespace] {
 			ref := f.(*pb.VASP)
 			if ref.Id == key {
-				a = ref
+				// Avoid modifying the object in the fixtures map
+				data = *ref
+				a = &data
```
pdeziel, line 354: Fixes a bug I ran into with one of the tests calling `CompareFixtures` which inadvertently altered one of the VASP fixtures and caused a subsequent test to fail.

rebeccabilbro (approved):
Thank you @pdeziel for ensuring that Trtl is well-covered by the GDS test suite! And thank you for the live review today and walking me and @bbengfort  through all the changes. I think this PR is nearly good to go - we just talked about updating the gitignore and adding a bit of commentary around that test fixture code. Nice work!

PR #367, opened 2022-03-03
fix: change date format to MM-DD-YYYY

masskoder:


bbengfort:
@masskoder there is no equivalent of `toLocaleDateString()` that might format it in the correct representation based on the user's browser settings?

bbengfort (approved):
@masskoder thank you for making that change!

PR #375, opened 2022-03-04
sc-3973 Legal person customer number basic component

elysee15:


web/gds-user-ui/src/components/section/CustomerNumer.stories.tsx
```diff
@@ -0,0 +1,16 @@
+import { Meta, Story } from "@storybook/react";
```
masskoder, line 1: i guess the filename is CustomerNumber.stories.tsx instead of `CustomerNumer`

masskoder (approved):

PR #378, opened 2022-03-06
Defines BFF API v1

bbengfort:
This PR defines the GDS BFF v1 API. The purpose of this package is to define the JSON data structures for requests and responses in the v1 API. It also implements a BFFClient which is used in testing and can also be used from a CLI implementation, though we have no plans to implement a CLI except for debugging and development. 

This package mirrors the `admin/v2` package, though it does omit the authentication and credentials components. The BFF epic currently has no authenticated endpoints, so I figured we'd save this until it was required for the GDS User UI.

### DIFF
--- a/pkg/bff/api/v1/client.go
+++ b/pkg/bff/api/v1/client.go
@@ -1,0 +1,160 @@
+package api
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/cookiejar"
+	"net/url"
+	"time"
+)
+
+// New creates a new api.v1 API client that implements the BFF interface.
+func New(endpoint string) (_ BFFClient, err error) {
+	c := &APIv1{
+		client: &http.Client{
+			Transport:     nil,
+			CheckRedirect: nil,
+			Timeout:       30 * time.Second,
+		},
+	}
+
+	// Create cookie jar
+	if c.client.Jar, err = cookiejar.New(nil); err != nil {
+		return nil, fmt.Errorf("could not create cookiejar: %s", err)
+	}
+
+	if c.endpoint, err = url.Parse(endpoint); err != nil {
+		return nil, fmt.Errorf("could not parse endpoint: %s", err)
+	}
+	return c, nil
+}
+
+// APIv1 implements the BFFClient interface.
+type APIv1 struct {
+	endpoint *url.URL
+	client   *http.Client
+}
+
+// Ensure the API implments the BFFClient interface.
+var _ BFFClient = &APIv1{}
+
+//===========================================================================
+// Client Methods
+//===========================================================================
+
+func (s *APIv1) Status(ctx context.Context) (out *StatusReply, err error) {
+	//  Make the HTTP request
+	var req *http.Request
+	if req, err = s.NewRequest(ctx, http.MethodGet, "/v1/status", nil, nil); err != nil {
+		return nil, err
+	}
+
+	// Execute the request and get a response
+	// NOTE: cannot use s.Do because we want to parse 503 Unavailable errors
+	var rep *http.Response
+	if rep, err = s.client.Do(req); err != nil {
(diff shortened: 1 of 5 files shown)


pkg/bff/api/v1/client.go
```diff
@@ -0,0 +1,160 @@
...
+	accept       = "application/json"
+	acceptLang   = "en-US,en"
+	acceptEncode = "gzip, deflate, br"
+	contentType  = "application/json; charset=utf-8"
+)
+
+// NewRequest creates an http.Request with the specified context and method, resolving
+// the path to the root endpoint of the API (e.g. /v2) and serializes the data to JSON.
+// This method also sets the default headers of all GDS Admin API v2 client requests.
+func (s *APIv1) NewRequest(ctx context.Context, method, path string, data interface{}, params *url.Values) (req *http.Request, err error) {
```
rebeccabilbro, line 93: `method` is a request type? e.g. GET, POST, DELETE?

pkg/bff/api/v1/client.go
```diff
@@ -0,0 +1,160 @@
...
+
+// Do executes an http request against the server, performs error checking, and
+// deserializes the response data into the specified struct if requested.
+func (s *APIv1) Do(req *http.Request, data interface{}, checkStatus bool) (rep *http.Response, err error) {
+	if rep, err = s.client.Do(req); err != nil {
+		return rep, fmt.Errorf("could not execute request: %s", err)
+	}
+	defer rep.Body.Close()
+
+	// Detect errors if they've occurred
```
rebeccabilbro, line 132: Would it be a good idea for us to strongly encourage BFF users to set `checkStatus` to true for debugging/introspection purposes?

pkg/bff/api/v1/errors.go
```diff
@@ -0,0 +1,52 @@
...
+var (
+	unsuccessful = Reply{Success: false}
+	notFound     = Reply{Success: false, Error: "resource not found"}
+	notAllowed   = Reply{Success: false, Error: "method not allowed"}
+)
+
+// ErrorResponse constructs an new response from the error or returns a success: false.
+func ErrorResponse(err interface{}) Reply {
+	if err == nil {
+		return unsuccessful
```
rebeccabilbro, line 20: What are some ways we think `unsuccessful` might come up?

rebeccabilbro (approved):
Looks like we're off to the races @bbengfort! I had a few small questions mostly w/r/t expected usage, though this seems ready to go to me 👍

pkg/bff/api/v1/client.go
bbengfort: Yes, this refers to the http request method and should be a valid http verb passed to `http.NewRequestWithContext`.

pkg/bff/api/v1/client.go
bbengfort: I'm hoping it will be rare for users of this library to directly call `Do` - instead the library methods that map to the endpoints should be our focus. These methods will generally set `checkStatus` to true, however there are cases where special handling is required (e.g. for a 404 Not Found error). In particular, the `Status` client method will set `checkStatus` to false so that it can handle a 503 Unavailable error, which will still have a body returned from the server.

pkg/bff/api/v1/errors.go
bbengfort: I remember this being added to prevent a panic in the Admin API when we were accidentally passing `nil` into `ErrorResponse` (I believe nil was being sent on an error channel, which is what caused the confusion). 

In terms of the use case, I could imagine this response being sent when the error code matters more, for example, `c.JSON(http.StatusPayloadTooLarge, ErrorResponse(nil))` -- here, it's largely the client's responsibility to break the payload into chunks using a multipart upload. The JSON body doesn't matter. However, even in this case, I'd prefer `c.JSON(http.StatusPayLoadTooLarge, ErrorResponse("payload is too large, upload in chunks"))` -- so mostly it is just panic prevention.

PR #382, opened 2022-03-07
Create BFF Binary

bbengfort:
Implements a urfave v2 CLI binary for running the BFF server.

### DIFF
--- a/cmd/bff/main.go
+++ b/cmd/bff/main.go
@@ -1,0 +1,71 @@
+package main
+
+import (
+	"log"
+	"os"
+
+	"github.com/joho/godotenv"
+	"github.com/trisacrypto/directory/pkg"
+	"github.com/trisacrypto/directory/pkg/bff"
+	"github.com/trisacrypto/directory/pkg/bff/config"
+	"github.com/urfave/cli/v2"
+)
+
+func main() {
+	// Load the dotenv file if it exists
+	godotenv.Load()
+
+	// Create the CLI application
+	app := &cli.App{
+		Name:    "gds-bff",
+		Version: pkg.Version(),
+		Usage:   "a backend for front-end for the GDS service",
+		Flags:   []cli.Flag{},
+		Commands: []*cli.Command{
+			{
+				Name:     "serve",
+				Usage:    "run the gds bff server",
+				Category: "server",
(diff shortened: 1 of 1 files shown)


cmd/bff/main.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Serve the GDS BFF service
+func serve(c *cli.Context) (err error) {
+	var conf config.Config
+	if conf, err = config.New(); err != nil {
+		return cli.Exit(err, 1)
+	}
+
+	if addr := c.String("addr"); addr != "" {
+		conf.BindAddr = addr
```
rebeccabilbro, line 59: Out of curiosity, at what stage is it typical to validate something like this -- eg if the user gives address w/o port are we figuring that the server error bubbling up is sufficient to communicate that the input wasn't correct? I realize this isn't super important for our own internal cli, so just asking more generally.

rebeccabilbro (approved):
Straightforward enough! This CLI will give us a convenient way to test out functionality as we build up the BFF; looks good @bbengfort !

cmd/bff/main.go
bbengfort: That's such a good question - it relates to the "trailing slash" problems I've had with different endpoint, DSN, URL parsing in the past. There are multiple ways to parse endpoints, addrs, URLs, etc. and that leads to ambiguity about what is required for what service. 

Unfortunately, while I'm invested in an answer for this, we haven't spent much time thinking about it in GDS. Right now we're mostly not validating that the addr is correct, we're passing whatever the user gives us to `net.Listen` and if that errors, that's when we return the error (e.g. in `Serve`). The only exception is in trtl/leveldb when we do some parsing to try to figure out where the DSN is pointing us. 

The options for validating are:

0. On the CLI -- but this could skip over validating environment variables or configuration files 
1. In config.Validate -- e.g. do we have a valid configuration?
2. In server.New -- when we create the server, do we have enough info to listen?
3. In server.Serve -- when we're ready to serve, should we check if we have what we need?
4. Try and see if there is an error (our current methods)

For other validation, we've mostly preferred option 1 combined with option 4. Sounds like it would be a good thing to chat about over 🍷  though!

PR #383, opened 2022-03-07
GDS BFF Dockerfile

bbengfort:
Implements the GDS BFF Dockerfile, adds the BFF to the docker-compose container and to the GitHub actions for build and push when the version is incremented. 

Files to review:

- containers/bff/Dockerfile
- containers/build.sh
- containers/docker-compose.yaml
- .workflows/github/containers.yaml

Blocked by #382

### DIFF
--- a/containers/docker-compose.yaml
+++ b/containers/docker-compose.yaml
@@ -120,6 +120,21 @@ services:
       - api
       - all
 
+  gds-bff:
+    build:
+      context: ../
+      dockerfile: ./containers/bff/Dockerfile
+    image: trisa/gds-bff
+    init: true
+    depends_on:
+      - gds
+    ports:
+      - 4437:4437
+    profiles:
+      - bff
+      - api
+      - all
+
   gds-ui:
     build:
       context: ../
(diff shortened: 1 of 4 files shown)


containers/docker-compose.yaml
```diff
@@ -120,6 +120,27 @@ services:
...
       - all
 
+  gds-bff:
+    build:
+      context: ../
+      dockerfile: ./containers/bff/Dockerfile
+    image: trisa/gds-bff
+    init: true
+    depends_on:
+      - gds
```
rebeccabilbro, line 130: Will we also (at some point) need to update the container directives for `gds-ui` and `gds-admin-ui` to depend on the BFF?

rebeccabilbro (approved):
Looks good @bbengfort - one small question w/r/t the depends-on relationships for the ui containers, but probably a future thing. Should be gtg when you are ready to merge!

containers/docker-compose.yaml
bbengfort: Oh good point - we'll have to remember when we add the `gds-user-ui` that it will depend on the BFF - `gds-ui` should go away; I don't think we have any plans right now for `gds-admin-ui` to depend on the BFF - but happy to discuss that.

PR #384, opened 2022-03-07
Configure BFF to GDS Client

bbengfort:
This PR adds a client to both the TestNet and MainNet GDS services. 

Files for review:

- pkg/bff/gds.go 
- pkg/bff/server.go
- pkg/bff/config/config.go
- pkg/bff/config/config_test.go

Blocked by #383


rebeccabilbro (approved):
Looks good @bbengfort - I think implementing specific connections to testnet and mainnet was the right way to go here; it allows us to get the BFF up and operational for the front end and will facilitate a key feature of the new UI, which is a seamless experience between testnet and mainnet from the user's perspective.

PR #391, opened 2022-03-09
BFF Tests and Mock GDS

bbengfort:
This PR creates the BFF test suite and implements a Mock GDS for those tests. 

The goal of the BFF tests is test the expected contract between the UI and the BFF. Therefore the following decisions were made:

1. The BFF server is run on a live, local-loopback socket similar to the Go builtin [httptest.Server](https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/net/http/httptest/server.go;l=28;drc=refs%2Ftags%2Fgo1.17.8;bpv=1;bpt=1). This allows end-to-end HTTP testing and also allows us to use the api.v1.BFFClient for tests, which should make developement and testing simpler. To ensure that this is possible, instead of implementing `ListenAndServe`, we've implemented separate `Listen` and `Serve` steps so that we can get the URL of the server and create the client in tests.
2. Rather than attempt to run the actual GDS for both TestNet and MainNet (and have to deal with leveldb, fixtures, etc.) we've opted to mock the GDS. The MockGDS allows the BFF to specify what it expects to be returned from GDS and simplifies the testing process. This means that we may need to create integration tests between the BFF and the actual GDS in addition to unit testing, but will allow us to focus on the client to BFF interactions rather than the BFF to GDS interactions. 

Minor changes may be needed as we start to implement tests, but I think this is a solid foundation with which to start building our RPC endpoints on.

### DIFF
--- a/pkg/bff/mock/gds.go
+++ b/pkg/bff/mock/gds.go
@@ -1,0 +1,125 @@
+package mock
+
+import (
+	"context"
+
+	"github.com/trisacrypto/directory/pkg/bff/config"
+	"github.com/trisacrypto/directory/pkg/utils/bufconn"
+	gds "github.com/trisacrypto/trisa/pkg/trisa/gds/api/v1beta1"
+	"google.golang.org/grpc"
+)
+
+const bufSize = 1024 * 1024
+
+func NewGDS(conf config.DirectoryConfig) (g *GDS, err error) {
+	g = &GDS{
+		srv:   grpc.NewServer(),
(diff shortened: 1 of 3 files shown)


rebeccabilbro (approved):
Looks good @bbengfort, and thank you for the live walkthrough earlier; especially glad to see  @pdeziel's hard work on the GDS tests coming to fruition here as we set up the bff test suite, and look forward to seeing the rest of the tests evolve from this!

PR #392, opened 2022-03-09
Status Endpoint

bbengfort:
Implements the Status endpoint and our first BFF API Tests that interact with the GDS nodes.

TODO: 

- [x] Implement `nogds` flag
- [x] Make status requests to GDS in two go routines
- [x] Test with Mock GDS status requests

### DIFF
--- a/cmd/bff/main.go
+++ b/cmd/bff/main.go
@@ -1,12 +1,17 @@
 package main
 
 import (
+	"context"
+	"encoding/json"
+	"fmt"
 	"log"
 	"os"
+	"time"
 
 	"github.com/joho/godotenv"
 	"github.com/trisacrypto/directory/pkg"
 	"github.com/trisacrypto/directory/pkg/bff"
+	"github.com/trisacrypto/directory/pkg/bff/api/v1"
 	"github.com/trisacrypto/directory/pkg/bff/config"
 	"github.com/urfave/cli/v2"
 )
@@ -33,6 +38,21 @@ func main() {
 						Aliases: []string{"a"},
 						Usage:   "the address and port to bind the server on",
 						EnvVars: []string{"GDS_BFF_BIND_ADDR"},
+					},
+				},
+			},
+			{
+				Name:     "status",
+				Usage:    "send a status check to the BFF server",
+				Category: "client",
+				Action:   status,
+				Flags: []cli.Flag{
+					&cli.StringFlag{
+						Name:    "url",
+						Aliases: []string{"u", "endpoint"},
+						Usage:   "specify the URL to connect to the BFF server on",
+						EnvVars: []string{"GDS_BFF_CLIENT_URL"},
+						Value:   "http://localhost:4437",
 					},
 				},
 			},
@@ -67,5 +87,37 @@ func serve(c *cli.Context) (err error) {
 	if err = srv.Serve(); err != nil {
 		return cli.Exit(err, 1)
 	}
+	return nil
+}
+
+// Status checks if the GDS BFF is up
+func status(c *cli.Context) (err error) {
+	var client api.BFFClient
+	if client, err = api.New(c.String("url")); err != nil {
+		return cli.Exit(err, 1)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	var rep *api.StatusReply
+	if rep, err = client.Status(ctx); err != nil {
+		return cli.Exit(err, 1)
+	}
+
+	return printJSON(rep)
+}
+
+//===========================================================================
+// Helper Functions
+//===========================================================================
+
+func printJSON(msg interface{}) (err error) {
+	var data []byte
+	if data, err = json.MarshalIndent(msg, "", "  "); err != nil {
+		return cli.Exit(err, 1)
+	}
+
+	fmt.Println(string(data))
 	return nil
 }
--- a/cmd/bff/main.go
+++ b/cmd/bff/main.go
@@ -54,6 +54,12 @@ func main() {
 						EnvVars: []string{"GDS_BFF_CLIENT_URL"},
 						Value:   "http://localhost:4437",
 					},
+					&cli.BoolFlag{
+						Name:    "nogds",
+						Aliases: []string{"no-gds", "G"},
+						Usage:   "health check the BFF without requesting GDS status",
+						Value:   false,
+					},
 				},
 			},
 		},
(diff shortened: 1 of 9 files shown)


cmd/bff/main.go
```diff
@@ -69,3 +89,35 @@ func serve(c *cli.Context) (err error) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	var rep *api.StatusReply
+	if rep, err = client.Status(ctx); err != nil {
+		return cli.Exit(err, 1)
+	}
+
+	return printJSON(rep)
+}
```
bbengfort, line 109: This is usually very helpful for devops and debugging, so I wanted to make sure it was in the CLI program.

pkg/bff/api/v1/api.go
```diff
@@ -25,7 +24,7 @@ type Reply struct {
 
 // StatusReply is returned on status requests. Note that no request is needed.
 type StatusReply struct {
-	Status    string    `json:"status"`
-	Timestamp time.Time `json:"timestamp,omitempty"`
-	Version   string    `json:"version,omitempty"`
+	Status  string `json:"status"`
+	Uptime  string `json:"uptime,omitempty"`
```
bbengfort, line 28: I prefer uptime to timestamp - it does the same job of busting the cache, but is a little more information which is nice.

pkg/bff/server_test.go
```diff
@@ -73,15 +73,25 @@ func (s *bffTestSuite) SetupSuite() {
 	// state doesn't change between tests.
 	go s.bff.Serve()
 
+	// Wait for 500 ms to ensure the BFF starts serving
+	time.Sleep(500 * time.Millisecond)
+
 	// Create the BFF client for making requests to the server
+	require.NotEmpty(s.bff.GetURL(), "no url to connect to client on")
```
bbengfort, line 80: Was running into a race condition between `s.bff.Serve` and the `GetURL` function; added a sleep to prevent the race … which isn't great, but we don't have a "ready" signal.

pkg/bff/server_test.go
```diff
@@ -73,15 +73,25 @@ func (s *bffTestSuite) SetupSuite() {
...
+}
+
 func (s *bffTestSuite) TearDownSuite() {
 	require := s.Require()
-	err := s.bff.Shutdown()
-	require.NoError(err, "could not shutdown the BFF server after tests")
+	require.NoError(s.bff.Shutdown(), "could not shutdown the BFF server after tests")
+	s.testnet.Shutdown()
+	s.mainnet.Shutdown()
 }
```
bbengfort, line 95: Forgot these in the last PR.

cmd/bff/main.go
```diff
@@ -69,3 +95,42 @@ func serve(c *cli.Context) (err error) {
...
+		}
+	}
+
+	var rep *api.StatusReply
+	if rep, err = client.Status(ctx, params); err != nil {
+		return cli.Exit(err, 1)
+	}
+
+	return printJSON(rep)
+}
```
bbengfort, line 122: This is usually very helpful for devops and debugging, so I wanted to make sure it was in the CLI program.

pkg/bff/api/v1/api.go
```diff
@@ -23,9 +22,16 @@ type Reply struct {
...
+	Version string `json:"version,omitempty"`
+	TestNet string `json:"testnet,omitempty"`
+	MainNet string `json:"mainnet,omitempty"`
```
bbengfort, line 36: TestNet and MainNet will be one of:

- "unavailable"
- "healthy" 

☝🏽  the two most common options

- "danger"
- "unhealthy"
- "maintenance"

bbengfort: "unavailable" is added when we cannot connect to GDS; the others are strings to lower case of the ServiceStatus enum in the TRISA protocol buffers. 

Unfortunately, version information for GDS is not available since it is not part of the TRISA protocol buffers.

pkg/bff/mock/gds.go
```diff
@@ -15,7 +15,7 @@ func NewGDS(conf config.DirectoryConfig) (g *GDS, err error) {
 	g = &GDS{
 		srv:   grpc.NewServer(),
 		sock:  bufconn.New(bufSize),
-		Calls: make(map[string]uint32),
+		Calls: make(map[string]int),
```
bbengfort, line 18: Switched to int to make it easier to do `require` assertions.

pkg/bff/server.go
```diff
@@ -172,12 +174,26 @@ func (s *Server) SetHealth(health bool) {
 	log.Debug().Bool("healthy", health).Msg("server health set")
 }
 
+func (s *Server) SetURL(url string) {
+	s.Lock()
+	s.url = url
+	s.Unlock()
+	log.Debug().Str("url", url).Msg("server url set")
+}
```
bbengfort, line 182: Wrapping in the lock prevents a race condition in order to get our tests to pass.

pkg/bff/status_test.go
```diff
@@ -0,0 +1,147 @@
...
+	// Set the Status RPC for the mocks
+	healthy := func(ctx context.Context, in *gds.HealthCheck) (*gds.ServiceState, error) {
+		return &gds.ServiceState{
+			Status:    gds.ServiceState_HEALTHY,
+			NotBefore: time.Now().Add(1 * time.Hour).Format(time.RFC3339Nano),
+			NotAfter:  time.Now().Format(time.RFC3339Nano),
+		}, nil
+	}
+	s.testnet.OnStatus = healthy
+	s.mainnet.OnStatus = healthy
```
bbengfort, line 30: An example of setting a handler on the GDS mock

pkg/bff/api/v1/client.go
```diff
@@ -47,10 +49,16 @@ var _ BFFClient = &APIv1{}
 // Client Methods
 //===========================================================================
 
-func (s *APIv1) Status(ctx context.Context) (out *StatusReply, err error) {
+func (s *APIv1) Status(ctx context.Context, in *StatusParams) (out *StatusReply, err error) {
+	// Create the query params from the input
+	var params url.Values
+	if params, err = query.Values(in); err != nil {
+		return nil, fmt.Errorf("could not encdoe query params: %s", err)
```
rebeccabilbro, line 56: encdoe -> encode

rebeccabilbro (approved):
Looks good @bbengfort! I agree that switching from timestamp to uptime is nicer.
I just noticed one tiny typo in an error message; otherwise should be good to go.

PR #402, opened 2022-03-10
BFF Test Fixtures

bbengfort:
This PR implements fixture loading for the GDS mocks to make implementing tests easier.

The basic use case is as follows:

1. Create a test fixture as JSON data in `pkg/bff/testdata` (e.g. `search_test_valid.json`) 
2. Call `s.testnet.UseFixture(mock.SearchRPC, "testdata/search_test_valid.json") - this will set `OnSearch` with a return type of the data that is in the fixture, returning an error if the fixture cannot be loaded or parsed. 
3. Execute a BFF request, verifying the response is expected from the fixture

Notes:

- the first parameter, `rpc` is the name of the RPC to use the fixture for, I've created string constants for each of the GDS RPCs. 
- the second parameter `path` must be a path to a JSON file that can be unmarshalled into the protocol buffer of the reply type of the RPC
- `UseFixture` can be called multiple times during the test to change fixtures in between. 

I've also created a `UseError` method to have the RPC mock return an error a bit easier. It follows the same pattern as `UseFixture`, but the user specifies a code and message for the gRPC error return. 

Open question:

Should I create `UseRegisterFixture`, `UseSearchFixture`, `UseRegisterError`, `UseSearchError`, etc. which are aliases for `UseFixture(RegisterRPC)`? 

This PR follows #401 so includes files from that PR, the files to review are:

- `pkg/bff/mock/gds.go`

### DIFF
--- a/pkg/bff/api/v1/api.go
+++ b/pkg/bff/api/v1/api.go
@@ -10,6 +10,7 @@ import (
 
 type BFFClient interface {
 	Status(ctx context.Context, in *StatusParams) (out *StatusReply, err error)
+	Lookup(ctx context.Context, in *LookupParams) (out *LookupReply, err error)
 }
 
 //===========================================================================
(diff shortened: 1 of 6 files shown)


pkg/bff/status_test.go
```diff
@@ -63,9 +63,7 @@ func (s *bffTestSuite) TestStatus() {
 	require.Equal(s.mainnet.Calls["Status"], 2)
 
 	// Test Status with TestNet Error
-	s.testnet.OnStatus = func(ctx context.Context, in *gds.HealthCheck) (*gds.ServiceState, error) {
-		return nil, status.Error(codes.Unavailable, "unreachable host")
-	}
+	s.testnet.UseError(mock.StatusRPC, codes.Unavailable, "unreachable host")
```
bbengfort, line 66: Here is an example of `UseError` - I will use `UseFixture` in the Lookup endpoint tests.

rebeccabilbro (approved):
This should make fixture loading for tests very convenient and flexible! I don't think the aliases are really necessary (they'd only save a few characters of typing each), and semantically are pretty similar to the `UseFixture(ThisRPC)` style. Nice work @bbengfort!

PR #403, opened 2022-03-10
GDS BFF: Lookup Tests

bbengfort:
This PR implements tests for the BFF Lookup Endpoint. I think this PR also highlights how successful our mock framework is and how easy it is to test different interactions with the GDS server! Writing these tests felt good and felt like things made sense. 

This PR follows #401 the files for review are:

- pkg/bff/gds_test.go
- pkg/bff/testdata/testnet/lookup_reply.json 
- pkg/bff/testdata/mainnet/lookup_reply.json

PR #405, opened 2022-03-11
GDS BFF: Register

bbengfort:
This PR implements the BFF Register endpoint. We've made the decision in v1.4 that the user specifies which directory they're registering to, multi-directory registration will happen in v1.5. This endpoint is therefore essentially a passthrough to the GDS specified by the client. 

Blocked by #403

Note: the files changed in this PR are the same as those changed in previous PRs, I'm happy to rebase before review.


rebeccabilbro (approved):
LGTM @bbengfort! Thank you for rebasing to make the review files a bit more clearcut 🙏

PR #422, opened 2022-03-22
BFF Verify Contact Endpoint

bbengfort:
Implements the initial verify email endpoint for the 1.4 GDS and BFF by directing the user to a URL that contains a `registered_directory` param so that the BFF can identify the directory to forward the request to. 

This PR contains both the endpoint implementation and the tests. The files and mechanisms changed should be similar to the Register endpoint. 

TODO:

- [x] change GDS verify contact email to include the additional param
- [x] test the GDS verify contact email change

### DIFF
--- a/pkg/bff/gds.go
+++ b/pkg/bff/gds.go
@@ -24,8 +24,10 @@ import (
 )
 
 const (
-	testnet = "testnet"
-	mainnet = "mainnet"
+	testnet       = "testnet"
+	mainnet       = "mainnet"
+	trisatest     = "trisatest.net"
+	vaspdirectory = "vaspdirectory.net"
 )
 
 // ConnectGDS creates a gRPC client to the TRISA Directory Service specified in the
@@ -233,7 +235,7 @@ func (s *Server) Register(c *gin.Context) {
 		return
 	}
 
-	// Make the gds request
+	// Make the GDS request
 	log.Debug().Str("network", network).Msg("issuing GDS register request")
 	ctx, cancel := context.WithTimeout(c.Request.Context(), 25*time.Second)
 	var rep *gds.RegisterReply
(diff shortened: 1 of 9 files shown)


pkg/gds/emails/client.go
```diff
@@ -116,9 +116,10 @@ func (m *EmailManager) SendVerifyContacts(vasp *pb.VASP) (sent int, err error) {
 // SendVerifyContact sends a verification email to a contact.
 func (m *EmailManager) SendVerifyContact(vasp *pb.VASP, contact *pb.Contact) (err error) {
 	ctx := VerifyContactData{
-		Name:    contact.Name,
-		VID:     vasp.Id,
-		BaseURL: m.conf.VerifyContactBaseURL,
+		Name:        contact.Name,
+		VID:         vasp.Id,
+		BaseURL:     m.conf.VerifyContactBaseURL,
+		DirectoryID: m.conf.DirectoryID,
```
bbengfort, line 122: Adding this directory ID from the configuration should be enough information for the BFF to route the request to the correct GDS.

pkg/bff/server.go
```diff
@@ -196,6 +196,7 @@ func (s *Server) setupRoutes() (err error) {
 		// GDS public routes (no authentication required)
 		v1.GET("/lookup", s.Lookup)
 		v1.POST("/register/:network", s.Register)
+		v1.GET("/verify-contact", s.VerifyContact)
```
bbengfort, line 199: Endpoint matches the current URL - happy to change to `/verify` if that seems better.

pkg/bff/gds.go
```diff
@@ -284,3 +286,82 @@ func (s *Server) Register(c *gin.Context) {
...
+		case codes.NotFound:
+			c.JSON(http.StatusNotFound, api.ErrorResponse(serr.Message()))
+		case codes.Aborted:
+			c.JSON(http.StatusConflict, api.ErrorResponse(serr.Message()))
+		default:
+			log.Error().Err(err).Str("code", serr.Code().String()).Str("registered_directory", params.Directory).Msg("could not verify contact")
+			c.JSON(http.StatusInternalServerError, api.ErrorResponse(serr.Message()))
+		}
+		return
+	}
```
bbengfort, line 350: I researched the errors that GDS `VerifyContact` will return; in addition to `InvalidArgument`, `NotFound`, and `Aborted`, it will also return `Internal` and `FailedPrecondition` - but these are both 500 errors. GDS returns a user-friendly error, so it is safe to just return the error message from the server.

pkg/bff/gds.go
```diff
@@ -24,8 +24,10 @@ import (
 )
 
 const (
-	testnet = "testnet"
-	mainnet = "mainnet"
+	testnet       = "testnet"
+	mainnet       = "mainnet"
+	trisatest     = "trisatest.net"
+	vaspdirectory = "vaspdirectory.net"
```
rebeccabilbro, line 30: Just confirming; are we figuring that internally to the bff we refer to the test network as `testnet` but that users of the bff will reference it as `trisatest.net`? Just want to make sure we document our reason for referencing the two networks in multiple ways so that we don't get mixed up later!

pkg/bff/server.go
rebeccabilbro: `/verify` does seem a bit nicer; though if the ship has sailed already and we're comfortable with `/verify-contact` I won't make a fuss

pkg/gds/emails/client.go
rebeccabilbro: Got it - and `ID` here means the string `testnet` or `mainnet`? Or the urls `trisatest.net` & `vaspdirectory.net`?

rebeccabilbro (approved):
Looks good @bbengfort -- I had a small question about how we reference the two networks internally vs externally, but otherwise this should be good to go!

pkg/gds/emails/client.go
bbengfort: In this case it will be `trisatest.net` or `vaspdirectory.net`.

pkg/bff/gds.go
bbengfort: This is sort of a weird situation: the GDS thinks of itself as `trisatest.net` or `vaspdirectory.net` - so when the GDS emails users, that's how it will represent itself. This makes sense from the perspective of the GDS since there is the possibility of multiple directory services that are respected by TRISA. Users (e.g. VASPs/TRISA implementers) should also think about the directories as `trisatest.net` and `vaspdirectory.net` since those are the endpoints they will be connecting to (though there is a lot of use of the colloquial "MainNet" and "TestNet" conversationally - but this tends to always require clarification, even now). 

The BFF internally thinks about `testnet` and `mainnet`; but unfortunately BFF users (e.g. the UI) will have to understand both just like the BFF has to understand both (though `testnet` and `mainnet` will be the common case). This is for two reasons:

1. In URLs with path parameters, e.g. `/v1/register/:network` we have to use `testnet` and `mainnet` because `vaspdirectory.net` and `trisatest.net` are not valid. They are valid in query parameters, e.g. `/v1/verify?network=trisatest.net` 
2. In copy the UI will have to make sure it uses "TestNet" and "MainNet" to describe the different networks/certificates but "trisatest.net" and "vaspdirectory.net" as the directories, and recommend to users that they should connect to the directories via "api.trisatest.net:443" and "api.vaspdirectory.net:443" respectively. 

Most of this will be made easier in BFF v1.5 - we plan to have a single verification token which is sent from the BFF - that way the contact is verified for both the TestNet and the MainNet with a single email; that will eliminate case #1 - but not case #2. 

As an alternative we could replace "testnet" with the "trisatest" and "mainnet" with "vaspdirectory", making the .net TLD optional for things like case #1. I'm certainly open to that.

PR #423, opened 2022-03-23
BFF GDS Parallel Request Helper

bbengfort:
This PR implements a utility helper to make parallel requests to the GDS for Status, Lookup, Search, and Verification. 

Blocked by #422


rebeccabilbro (approved):
Looks good @bbengfort; this will definitely streamline how we make rpc calls to mainnet and testnet in parallel in the BFF! Thank you for the walkthrough earlier; I don't have any additional comments so I think we should be good to go to merge 👍

PR #425, opened 2022-03-24
sc-3670 POST to resend results in 405 Not Allowed

elysee15:
Fixes SC-3670

### DIFF
--- a/web/gds-admin-ui/src/components/ResendEmail.js
+++ b/web/gds-admin-ui/src/components/ResendEmail.js
@@ -8,6 +8,7 @@ import { getCookie } from '../utils';
 import axios from 'axios';
 import toast from 'react-hot-toast';
 import { actionType, useModal } from 'contexts/modal';
+import { APICore } from 'helpers/api/apiCore';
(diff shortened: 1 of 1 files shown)


web/gds-admin-ui/src/components/ResendEmail.js
```diff
@@ -96,7 +99,7 @@ function ResendEmail() {
             reason: ""
         }
         setIsSubmitting(true)
-        axios.post(`/vasps/${vasp?.id}/resend`, payload, {
+        api.create(`/vasps/${vasp?.id}/resend`, payload, {
```
bbengfort, line 102: @elysee - this makes sense that this would fix the issue; I'm assuming that it wasn't correctly redirecting to `/v2` before. 

Before we merge this in, could you remove the `gds-user-ui` assets or rebase as needed?

elysee15: sure

masskoder (approved):
Looks good to me

PR #426, opened 2022-03-25
sc-3985 ResetFixtures cleanup

pdeziel:
This cleans up the ResetFixtures functions which were no longer necessary in favor of a single function.


bbengfort (approved):
@pdeziel thank you for taking care of this technical debt so we could wrap up the v1.4 epic!

PR #434, opened 2022-03-28
sc-3986 prevent certreq fixture modification

pdeziel:
In the GDS test suite `CompareFixture` method we are unintentionally overwriting the reference fixtures when we clear out the Modified and Created timestamps. Since the tests assume that the reference fixtures remain constant, this PR makes sure that we are only modifying a copy of the CertReq (and VASP, but that was done in a previous PR).


bbengfort (approved):
Thank you!

PR #441, opened 2022-03-31
Fix CI by specifying versions

bbengfort:
The latest version of `staticcheck` is failing to build. This PR sets the staticcheck version to v0.2.0 which was released in November 2021. We should update to v0.3.0 when they fix their build problems. 

While we're here, I've also changed the `generate.go` files to use absolute instead of relative paths based on `$GOPATH` and regenerated our files.


pdeziel (approved):
Looks good, does this change anything for generating the protocol buffers locally? For example, does the user need to make sure that their GOPATH is set correctly?

bbengfort:
@pdeziel yes, the user should have their `$GOPATH` set correctly in order to run `go generate` - if they don't, likely a `protoc` file not found error will be printed. Although `go modules` reduces the dependency on `$GOPATH` - I do think it's unlikely that a GDS developer wouldn't have this set correctly and still be able to run `go generate`.

PR #442, opened 2022-03-31
Reduce Test Logging Output

bbengfort:
This PR adds two utilities to our `pkg/util/logger` package:

1. `Testing` - sets the logger to use the `testing.T` logger
2. `Discard` - sets the logger to discard all output 

The logger package tracks the original logger and has a `Reset` functionality to cleanup after the test. 

Right now I've updated the test suites to use `Discard` in `SetupSuite` and to `Reset` in `TearDownSuite` -- for this `ConsoleLog` _must_ be false, otherwise the discard logger is overridden. This significantly reduces the amount of logging output during tests and will hopefully make it easier for us to figure out where tests are failing and handle them.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/prometheus/client_golang v1.11.0
 	github.com/rogpeppe/go-internal v1.8.1 // indirect
 	github.com/rotationalio/honu v0.3.0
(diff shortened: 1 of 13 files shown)


pdeziel (approved):
Thanks, this is definitely a pain point of GDS development so I'm glad some progress is being made here.

PR #443, opened 2022-04-01
Connect BFF to Trtl

bbengfort:
This PR implements a database configuration for BFF to connect to trtl, implements a stub `bff/db` package for the BFF to interact with trtl, and implements a mock trtl connecting to a temporary directory leveldb for testing.

### DIFF
--- a/pkg/gds/backup.go
+++ b/pkg/gds/backup.go
@@ -29,7 +29,7 @@ func (s *Service) BackupManager(stop <-chan bool) {
 		return
 	}
 
-	// Check backup directory
+	// Check backup directory, creating it as necessary
 	backupDir, err := s.getBackupStorage()
 	if err != nil {
 		log.Fatal().Err(err).Msg("backup manager cannot access backup directory")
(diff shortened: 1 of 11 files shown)


pkg/gds/backup.go
```diff
@@ -51,40 +50,61 @@ backups:
...
+		// Execute the backup - error messages are logged in the Backup function so they
+		// are ignored here and are only returned from Backup for testing purposes.
+		s.Backup(backupDir)
+	}
+}
+
+// Backup performs a backup on behalf of the service. BackupManager calls this function
+// at a periodic interval to take a snapshot of the store to disk and to keep only the
+// configured number of archives.
+func (s *Service) Backup(path string) (err error) {
```
bbengfort, line 62: This wasn't part of this PR - but the tests were breaking, and this was required to fix the tests.

rebeccabilbro (approved):
This all makes sense to me @bbengfort - I like that we've abstracted trtl slightly with the `db` package to capture the bff use case, and thank you for fixing that backup issue that was causing the tests to error!

PR #459, opened 2022-04-07
sc-3471: add pull request template

masskoder:


.github/pull_request_template.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Acceptance criteria
+
+Describe how reviewers can test this change to be sure that it works correctly. Add a checklist if possible.
+
+### Author checklist
+
+- [ ] I have manually tested the change and/or added automation in the form of unit tests or integration tests
+- [ ] I have added migrations for any changes to the database
+- [ ] My migrations can be run before the code is deployed without causing any issues
```
rebeccabilbro, line 20: Not sure if these two make sense for our project since we don't have a relational database backend. Perhaps we could replace these with something more like:
- [ ] I have updated the dependencies list
- [ ] I have recompiled and included new protocol buffers to reflect changes I made
- [ ] I have added new test fixtures as needed to support added tests

etc

.github/pull_request_template.md
```diff
@@ -0,0 +1,29 @@
...
+- [ ] Check this box if a reviewer can merge this pull request after approval (leave it unchecked if you want to do it yourself)
```
rebeccabilbro, line 21: Let's remove this one since our current protocol is that everyone merges their own PRs after they have been approved

.github/pull_request_template.md
```diff
@@ -0,0 +1,29 @@
...
+- [ ] I have moved the associated ShortCut or Jira ticket to "Ready for Review"
+
+### Reviewer(s) checklist
+
+- [ ] The code in the PR has been reviewed for architectural concerns and correctness
+- [ ] Migrations work without causing API downtime
```
rebeccabilbro, line 27: Again, the migrations language doesn't make as much sense in the context of this project -- could we update this to make it more relevant to the kinds of things we check for in the front and back end during PR reviews. For example, for backend PRs, this list might be something like checking for race conditions/concurrency issues, good error handling and logging, maintaining backwards compatibility, etc. Maybe you could also list some of the things you check for in frontend PRs?

.github/pull_request_template.md
```diff
@@ -0,0 +1,29 @@
...
+- [ ] The acceptance criteria have been reviewed and tested
```
rebeccabilbro, line 28: Maybe we should also add a checklist item for QA? e.g. 

.github/pull_request_template.md
```diff
@@ -0,0 +1,29 @@
+### Type of change
+
+- [ ] bug fix
+- [ ] new feature
+- [ ] documentation
+- [ ] other (describe)
+
+### Scope of changes
+
+Briefly describe the scope of changes. Include links to ShortCut or Jira ticket if applicable.
```
rebeccabilbro, line 10: Perhaps we should add a note here that our practice is to actually name our PR branches after the Shortcut stories they resolve and to include the `sc-XXXX` in the PR title , since we have automations set up to support workflow management in Shortcut

rebeccabilbro (approved):
Thank you so much @masskoder for stepping up to work on this PR template; I really like the idea of using effective PR titles and body text to facilitate the PR review process! 

I have added a few comments and thoughts with the intention of tailoring this template to the Rotational workflow. Would you mind taking a look and addressing those before merging? Let me know if you have any questions or if you'd like to discuss more together!

PR #460, opened 2022-04-08
sc-4690 TRISA endpoint validation doesn't provide helper text to explain validation failures

elysee15:
Screen Shot 2022-04-08 at 07 29 33


masskoder (approved):
@elysee15  If the latest PR changes are not required in the current branch, could you create your branch from the main branch? 
Also for all your current PR could you rename `components/testnetProgress/CertificateStepLabel` -> `components/TestnetProgress/CertificateStepLabel`

And `modules/dashboard/Certificate` -> `modules/dashboard/certificate`

You can use these commands to make sure that the folder is renamed correctly 
``` git mv casesensitive tmp```
  ```    git mv tmp CaseSensitive```

PR #477, opened 2022-04-14
sc-4849 add country of registration when it is not LEI

elysee15:
Screen Shot 2022-04-14 at 23 31 36
Screen Shot 2022-04-14 at 23 31 50


masskoder (approved):
Looks good.

PR #492, opened 2022-04-18
Contact Verification Checks

bbengfort:
This PR updates the GDS to use trisa version 0.3.3 - which updates the validation mechanisms from IVMS101 - many test fixtures had to be updated to exclude country as a result. The new handlers add `contact.IsZero` checks to ensure the contact is not empty; e.g. does not have a name, email, or phone number. Valid contacts are now required throughout the code.

On registration, zero-valued contacts are set to nil to allow the front end to send empty valued data to the backend without error.


rebeccabilbro (approved):
Looks good @bbengfort -- thank you for updating GDS with the latest gRPC, JWT, and TRISA deps!

PR #495, opened 2022-04-19
Fixes panic when trtl becomes unavailable and adds panic recovery

bbengfort:
Fixes a panic that occurred when `trtl` was running when the `gds` server started, but came down after bootup. Adds tests for the panic case and also adds panic recovery so that the node doesn't come down even when it panics.

### DIFF
--- a/cmd/gds/main.go
+++ b/cmd/gds/main.go
@@ -1420,6 +1420,9 @@ func loadProfile(c *cli.Context) (err error) {
 
 // helper function to create the GRPC client with default options
 func initClient(c *cli.Context) (err error) {
+	if profile.Directory == nil {
+		return cli.Exit("current profile does not contain directory configuration", 1)
+	}
 	if client, err = profile.Directory.Connect(); err != nil {
 		return cli.Exit(err, 1)
 	}
@@ -1427,6 +1430,9 @@ func initClient(c *cli.Context) (err error) {
 }
 
 func initAdminClient(c *cli.Context) (err error) {
+	if profile.Admin == nil {
+		return cli.Exit("current profile does not contain admin configuration", 1)
+	}
 	if adminClient, err = profile.Admin.Connect(); err != nil {
 		return cli.Exit(err, 1)
 	}
@@ -1434,6 +1440,9 @@ func initAdminClient(c *cli.Context) (err error) {
 }
 
 func initMembersClient(c *cli.Context) (err error) {
+	if profile.Members == nil {
+		return cli.Exit("current profile does not contain members configuration", 1)
+	}
 	if membersClient, err = profile.Members.Connect(); err != nil {
 		return cli.Exit(err, 1)
 	}
@@ -1442,7 +1451,6 @@ func initMembersClient(c *cli.Context) (err error) {
 
 // helper function to print JSON response and exit
 func printJSON(msg interface{}) (err error) {
-
 	var data []byte
 	switch m := msg.(type) {
 	case proto.Message:
--- a/pkg/gds/interceptor.go
+++ b/pkg/gds/interceptor.go
@@ -2,8 +2,11 @@ package gds
 
 import (
 	"context"
+	"fmt"
+	"runtime/debug"
 	"time"
 
+	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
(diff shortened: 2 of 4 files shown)


cmd/gds/main.go
```diff
@@ -1420,20 +1420,29 @@ func loadProfile(c *cli.Context) (err error) {
...
 	if adminClient, err = profile.Admin.Connect(); err != nil {
 		return cli.Exit(err, 1)
 	}
 	return nil
 }
 
 func initMembersClient(c *cli.Context) (err error) {
+	if profile.Members == nil {
+		return cli.Exit("current profile does not contain members configuration", 1)
+	}
```
bbengfort, line 1445: These were an unrelated panic - but figured I'd fix the panic with a reasonable error message to make debugging easier in the future.

pkg/gds/interceptor.go
```diff
@@ -21,6 +37,7 @@ func (s *Service) serverInterceptor(ctx context.Context, in interface{}, info *g
 
 	// Call the handler to finalize the request and get the response.
 	out, err = handler(ctx, in)
+	panicked = false
```
bbengfort, line 40: This is the primary panic recovery middleware. We could use the recovery middleware from grpc-ecosystem and chain things together, however we're not quite at the point where we need chained middleware and to keep our dependency footprint light, I just added the panic `recover` function. If we do move to chained middleware in the future, we should use the provided `Recovery` interceptor rather than role our own.

pkg/gds/store/trtl/iterator.go
```diff
@@ -279,7 +279,12 @@ func (i *trtlStreamingIterator) Error() error {
...
 
 func (i *trtlStreamingIterator) Release() {
-	i.cursor.CloseSend()
+	// If the cursor is nil it's likely that the connection erred and i.err is not nil.
+	// Release() should be called safely before Error() to ensure any resources are
+	// cleaned up (e.g. canceling the context).
+	if i.cursor != nil {
+		i.cursor.CloseSend()
+	}
 	i.cancel()
```
bbengfort, line 288: The problem was that on connect in `Next()` an error (`i.err`) was on the cursor but `Release` was called before the error check, causing a nil panic on `i.cursor`.

pkg/gds/interceptor.go
```diff
@@ -13,6 +16,19 @@ import (
...
+			log.WithLevel(zerolog.PanicLevel).
+				Err(fmt.Errorf("%v", r)).
+				Str("stack_trace", string(debug.Stack())).
+				Msg("grpc server has recovered from a panic")
+			err = status.Error(codes.Internal, "an unhandled exception occurred")
+		}
+	}()
 
 	// Check if we're in maintenance mode - status method should still return a full response.
 	if s.conf.Maintenance && info.FullMethod != "/trisa.gds.api.v1beta1.TRISADirectory/Status" {
```
pdeziel, line 34: If I'm reading this right, does this cause the `recover` handler to get called on a status call in maintenance mode?

pkg/gds/store/trtl/iterator_test.go
```diff
@@ -345,3 +348,57 @@ func (s *trtlStoreTestSuite) TestStreamingIterator() {
 	require.Nil(iter.Error())
 	iter.Release()
 }
+
+func (s *trtlStoreTestSuite) TestStreamingIteratorError() {
+	require := s.Require()
+	iter := trtl.NewTrtlStreamingIterator(&trtlErrorClient{}, "")
+
+	// Call interactive methods on iter should do nothing but not return an error.
```
pdeziel, line 356:
```suggestion
	// Calling interactive methods on iter should do nothing but not return an error.
```

pdeziel (approved):
Looks good, I just had a comment edit and one question below.

pkg/gds/store/trtl/iterator_test.go
bbengfort: Ah shoot, I went too fast and I didn't manage to merge the suggestion; will update in the next PR - sorry.

pkg/gds/interceptor.go
bbengfort: It will, because the `defer` statement is before the maintenance mode check.

PR #515, opened 2022-05-05
sc-5100 bff performance tracking

pdeziel:
### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Scope of changes

sc-5100: This adds a few more environment variables for configuring sentry performance tracking and adds some instrumentation to the bff request handlers to track performance, if enabled.

### Acceptance criteria

- [ ] Set environment variables `SENTRY_DSN`, `SENTRY_ENVIRONMENT`,  and `GDS_BFF_SENTRY_TRACK_PERFORMANCE` and rebuild local dockers
- [ ] Use directory search functionality from the GDS UI and confirm that transactions get logged in Sentry dashboard
- [ ] Use directory register functionality from the GDS UI and confirm that the transaction gets logged
- [ ] Use verify contact functionality from the GDS Admin UI and confirm that the transaction gets logged

### Author checklist

- [x]  I have moved the associated ShortCut or Jira ticket to "Ready for Review"

### Reviewer(s) checklist

### DIFF
--- a/containers/docker-compose.yaml
+++ b/containers/docker-compose.yaml
@@ -137,6 +137,9 @@ services:
       - GDS_BFF_DATABASE_INSECURE=true
       - GDS_BFF_DATABASE_CERT_PATH
       - GDS_BFF_DATABASE_POOL_PATH
+      - GDS_BFF_SENTRY_TRACK_PERFORMANCE
+      - SENTRY_DSN
+      - SENTRY_ENVIRONMENT
     profiles:
       - api
       - testnet
--- a/pkg/bff/gds.go
+++ b/pkg/bff/gds.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/getsentry/sentry-go"
 	"github.com/gin-gonic/gin"
 	"github.com/rs/zerolog/log"
 	"github.com/trisacrypto/directory/pkg/bff/api/v1"
@@ -108,15 +107,8 @@ func (s *Server) Lookup(c *gin.Context) {
 		return rep, nil
 	}
 
-	ctx := c.Request.Context()
-	if s.conf.Sentry.TrackPerformance {
-		// Track Lookup request performance
-		span := sentry.StartSpan(ctx, "Lookup", sentry.TransactionName(req.String()))
-		defer span.Finish()
-	}
-
 	// Execute the parallel GDS lookup request, ensuring that flatten is true
-	results, errs := s.ParallelGDSRequests(ctx, lookup, true)
+	results, errs := s.ParallelGDSRequests(c.Request.Context(), lookup, true)
 
 	// If there were multiple errors, return a 500
 	if len(errs) == 2 {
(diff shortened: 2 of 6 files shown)


pkg/bff/gds.go
```diff
@@ -107,8 +108,15 @@ func (s *Server) Lookup(c *gin.Context) {
 		return rep, nil
 	}
 
+	ctx := c.Request.Context()
+	if s.conf.Sentry.TrackPerformance {
+		// Track Lookup request performance
+		span := sentry.StartSpan(ctx, "Lookup", sentry.TransactionName(req.String()))
+		defer span.Finish()
+	}
```
pdeziel, line 116: This is to verify that the perf tracking works, but apparently we can nest spans for more fine-grained perf tracking if we need it.

bbengfort: Very cool! Is it possible to embed this as a middleware to make it easy to track all API requests, or is it something we want to control at a pretty fine granularity?

containers/docker-compose.yaml
```diff
@@ -137,6 +137,9 @@ services:
       - GDS_BFF_DATABASE_INSECURE=true
       - GDS_BFF_DATABASE_CERT_PATH
       - GDS_BFF_DATABASE_POOL_PATH
+      - GDS_BFF_SENTRY_TRACK_PERFORMANCE
+      - SENTRY_DSN
+      - SENTRY_ENVIRONMENT
```
bbengfort, line 142: Would you mind adding all of these environment variables to `.env.template` as well?

pkg/bff/gds.go
pdeziel: We can definitely do this, the question is what do we want to label the transactions with? It's a bit difficult to use the request body from the middleware but should we just use a hard coded label or the content length for now?

bbengfort: Is the label the `"Lookup"` string in this code or the `sentry.TransactionName(req.String())` argument? 

If you're looking for the `"Lookup"` label, then the path of the request, e.g. `/v1/lookup` would be a good label -- and this should be available in the header, I haven't played around with the Gin context too much, but it's definitely in the HTTP metadata, so I assume it's easily available via Gin. 

I'm not 100% sure I understand the `req.String()` part of this, but if a unique, per-request ID is required; one thing that I used to do when I was writing raw HTTP servers was to add a UUID to the request context as it was coming into the server -- inner middleware and the handler could access that request ID if they wanted, and then you can use it on the way out of the handler. E.g. most middleware is structured as:

```go
func Middleware(req *http.Request, next http.HandlerFunc) (*http.Response, error) {
    // Handle incoming request before other middleware and handler
    // Add unique ID here
    
    rep, err = next(req) 

    // Handle outgoing request 
}
```

I know Gin complicates this a bit with its context mechanism and I don't recall the structure of Gin middleware off the top of my head, but that could be a solution. 

Also, I seem to remember that Sentry allows the front end to specify Transaction IDs so that we can track an error on both the front-end and the backend. E.g. if there is something like an `X-SENTRY-XFER-ID` header that the front-end sets we could use that as the unique request ID to match the profiling across services. 

Sorry if I didn't totally get the question - hope this is close to what you were asking at least.

pdeziel: So when you log a transaction you have to give it an "operation" name which would correspond to the request path in our case, and optionally you can pass in a label that appears in the Sentry dashboard. It seems that Sentry is generating transaction ids anyway so I think that part is just a mapping from the human-interpretable label to the id. So I think I'll just leave that option blank for now until we have a better idea for it.

Screen Shot 2022-05-05 at 4 48 15 PM

bbengfort: ok sounds good; sorry I didn't totally understand what `req.String()` was doing, but it sounds like we have a path forward and we can review when we start getting actual traces.

bbengfort (approved):

PR #545, opened 2022-05-19
Cleanup CertMan

bbengfort:
### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [x] other (technical debt deprecation)

### Scope of changes

Moves the `stop` signal in the CertMan routine to be selected with the ticker to ensure the routine is stopped immediately rather than after a specified number of ticks.

Refactors the body of the CertMan loop to a `HandleCertifcateRequests` method; refactors tests to use this method directly rather than relying on a certain number of intervals to pass; I'm hoping this helps reduce the occasional races detected by our CI. At the very least, it will make testing easier. 

Demotes the "could not find VASP by ID" and "multiple VASPs returned" log messages from the Warn level to the Debug level to reduce the number of alerts we get.

Cleans up the `.env.template` and trtl config.go files with some minor fixes.

Removes hotfix code: In v1.3 we modified the certificate requests to contain params, however certificate requests that were created during v1.2 needed special handling since they didn't have params. We added some hotfix code to deal with it. I've removed this hotfix code and gone through the current registrations in vaspdirectory.net and trisatest.net and don't believe that any old certificate requests will cause a problem. 

Fixes SC-5699
Fixes SC-3944

### Acceptance criteria

CI Tests pass, the changes make logical sense, no typos in doc strings, code style review. 

### Author checklist

- [x]  I have moved the associated ShortCut or Jira ticket to "Ready for Review"

### Reviewer(s) checklist


pkg/gds/gds.go
```diff
@@ -294,7 +294,7 @@ func (s *GDS) Lookup(ctx context.Context, in *api.LookupRequest) (out *api.Looku
 	case in.Id != "":
 		// TODO: add registered directory to lookup
 		if vasp, err = s.db.RetrieveVASP(in.Id); err != nil {
-			log.Warn().Err(err).Str("id", in.Id).Str("registered_directory", in.RegisteredDirectory).Msg("could not find VASP by ID")
+			log.Debug().Err(err).Str("id", in.Id).Str("registered_directory", in.RegisteredDirectory).Msg("could not find VASP by ID")
```
pdeziel, line 297: We were definitely feeling alert fatigue with these so thanks for taking the opportunity to make this change!

.env.template
```diff
@@ -105,9 +105,9 @@ TRTL_REPLICA_GOSSIP_SIGMA=1500ms
 
 # Trtl: Replica Configuration Strategy
 TRTL_REPLICA_STRATEGY_HOSTNAME_PID=false
-TRTL_REPLICA_STRATEGY_HOSTNAME=
-TRTL_REPLICA_STRATEGY_FILE_PID=false
-TRTL_REPLICA_STRATEGY_JSON_CONFIG=
+TRTL_REPLICA_STRATEGY_HOSTNAME=""
+TRTL_REPLICA_STRATEGY_FILE_PID=""
+TRTL_REPLICA_STRATEGY_JSON_CONFIG=""
```
pdeziel, line 110: Is this a functional change or a cosmetic one?

pdeziel (approved):
Looks good to me, the decoupling of the cert processing makes sense and seems like it makes this a lot more compatible with tests.

.env.template
bbengfort: It's hard to say - I was getting an error that it couldn't load the PID file, and when I added the quotes it seemed to fix it; however I didn't have that problem before, so I'm not sure what is going on. I couldn't reproduce just now - so it must be unrelated.

PR #569, opened 2022-06-06
Implement BFF Interactions with the Auth0 Management API

bbengfort:
### Scope of changes

Implements functionality so that the BFF can connect to the Auth0 Management API and set permissions on sign up.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

TODO: Describe how reviewers can test this change to be sure that it works correctly. Add a checklist if possible

### Author checklist

### Reviewer(s) checklist


.env.template
```diff
@@ -132,8 +132,10 @@ GDS_BFF_CONSOLE_LOG=true
 GDS_BFF_ALLOW_ORIGINS="http://localhost:3000,http://localhost:3003"
 GDS_BFF_COOKIE_DOMAIN=localhost
 
-GDS_BFF_AUTH0_ISSUER=
+GDS_BFF_AUTH0_DOMAIN=
```
bbengfort, line 135: The domain is required to compute the management API endpoints - so we could no longer call this the issuer. This might change some .env files for folks who have been running BFF in docker compose.

go.mod
```diff
@@ -1,35 +1,23 @@
 module github.com/trisacrypto/directory
 
-go 1.16
+go 1.17
```
bbengfort, line 3: @rebeccabilbro ok, so now this PR is officially out of hand - looks like I need Go 1.17 to finish this task since I'm relying on something in the `http` standard library. Should we just push through and go with this or should we discuss as a group?

DanielSollis (approved):
It took about three cups of coffee but I managed to get through the PR, I just have a few minor comments, overall it looks good.

pkg/bff/users.go
```diff
@@ -0,0 +1,82 @@
...
+	if roles, err = s.auth0.User.Roles(*user.ID); err != nil {
+		log.Error().Err(err).Msg("could not fetch roles associated with the user")
+		c.JSON(http.StatusInternalServerError, "could not complete user login")
+		return
+	}
+
+	if len(roles.Roles) == 0 {
+		// Assign the user the organization collaborator role
+		var collaborator *management.Role
+		if collaborator, err = s.FindRoleByName(DefaultRole); err != nil {
```
pdeziel, line 51: This requires us to define the `DefaultRole` in Auth0, correct?

pkg/bff/auth/auth_test.go
```diff
@@ -63,11 +65,12 @@ func TestClaimsContext(t *testing.T) {
...
 	// Test context with values
 	c.Set(auth.ContextBFFClaims, vclaims.CustomClaims.(*auth.Claims))
-	c.Set(auth.ContextRegisteredClaims, vclaims.RegisteredClaims)
+	c.Set(auth.ContextRegisteredClaims, &vclaims.RegisteredClaims)
 
 	bclaims, err := auth.GetClaims(c)
 	require.NoError(t, err, "could not fetch bff claims")
-	require.Equal(t, "eee784b5-49b3-452e-97d5-1b01e79f5e62", bclaims.VASP)
+	require.Equal(t, "6f0d943d-6cd7-4745-bc9d-6d65e32c70e9", bclaims.OrgID)
+	require.Equal(t, "eee784b5-49b3-452e-97d5-1b01e79f5e62", bclaims.VASP["testnet"])
```
pdeziel, line 73: Do we actually define the `testnet` and `mainnet` constants anywhere or have we just been relying on being consistent with the spelling and capitalization?

pkg/bff/auth/authtest/server.go
```diff
@@ -0,0 +1,141 @@
...
+			ExpiresAt: jwt.NewNumericDate(time.Now().Add(15 * time.Minute)),
+		},
+		Email: Email,
+		OrgID: OrgID,
+		VASPs: map[string]string{
+			"testnet": TestNetVASP,
+			"mainnet": MainNetVASP,
+		},
+		Scope:       Scope,
+		Permissions: permissions,
```
pdeziel, line 109: Am I understanding this right that the claim "structure" is defined somewhere in Auth0?

pdeziel (approved):
Looks good to me, I had a few questions and comments but overall I think the "live" test package is a good thing to have, even if we aren't running those often.

containers/docker-compose.yaml
```diff
@@ -152,8 +152,10 @@ services:
       - GDS_BFF_SENTRY_DSN
       - GDS_BFF_SENTRY_ENVIRONMENT
       - GDS_BFF_SENTRY_TRACK_PERFORMANCE
-      - GDS_BFF_AUTH0_ISSUER=dev-bu-hbv3o.us.auth0.com
+      - GDS_BFF_AUTH0_DOMAIN=dev-bu-hbv3o.us.auth0.com
       - GDS_BFF_AUTH0_AUDIENCE=https://bff.vaspdirectory.net
+      - GDS_BFF_AUTH0_CLIENT_ID
+      - GDS_BFF_AUTH0_CLIENT_SECRET
```
bbengfort, line 158: @masskoder you will have to update your `.env` file with these values from auth0 after we merge this PR.

pkg/bff/auth/auth.go
```diff
@@ -183,6 +186,42 @@ func Authorize(permissions ...string) gin.HandlerFunc {
...
+	return func(c *gin.Context) {
+		claims, err := GetRegisteredClaims(c)
+		if err != nil || claims.Subject == "" {
+			// c.Error will panic if err is nil so wrap in a guard
+			if err != nil {
+				c.Error(err)
+			}
+			c.AbortWithStatusJSON(http.StatusUnauthorized, api.ErrorResponse("could not identify authenticated user in request"))
+			return
+		}
```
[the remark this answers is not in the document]
bbengfort: I can't think of any issues checking `err` twice; the semantics are fundamentally different between the first if block and the second; e.g. the first block says that we want to do some shared handling (abort with status json) in the two cases where there is an error or the subject is empty; the second if block is simply the case where err is not nil not caring about the subject. I think extending this only duplicates code:

```go
if err != nil {
    c.Error(er)
    c.AbortWithStatusJSON(http.StatusUnauthorized, api.ErrorResponse("could not identify authenticated user in request"))
    return
} else if claims.Subject == "" {
    c.AbortWithStatusJSON(http.StatusUnauthorized, api.ErrorResponse("could not identify authenticated user in request"))
    return
}
```

It's essentially the same number of lines, but without the duplication of the AbortWithStatus. 

For @pdeziel's suggestion, the if/else if structure is better if we want to have a different abort with status message if `claims.Subject` is empty, but I don't think it's necessary. 

Was there some other code snippet that I'm missing?

pkg/bff/auth/auth.go
```diff
@@ -202,6 +241,16 @@ func GetRegisteredClaims(c *gin.Context) (*validator.RegisteredClaims, error) {
...
+}
+
+// GetUserInfo fetches the user info from the gin context. Returns an error if no user
+// exists on the context or if the user value is nil. Panics if user is incorrect type.
+func GetUserInfo(c *gin.Context) (*management.User, error) {
+	user, exists := c.Get(ContextUserInfo)
+	if !exists || user == nil {
+		return nil, ErrNoUserInfo
+	}
+	return user.(*management.User), nil
```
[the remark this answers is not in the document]
bbengfort: That's a good question - in the docstring I mention that this function will cause a panic if the user is an incorrect type, and I think that's important here. I tend to differentiate between user errors (e.g. if the user can change the request and have it succeed) vs system or developer errors (that have to be changed in code). If this is the wrong type, it's the latter kind of error and we need to be alerted to it. 

The GDS BFF code has two things that deal with the panic:

1. panic recovery middleware that will return a 500 to the user 
2. sentry logging that will alert us to the panic and create a shortcut story for us 

So in this case, I'm very comfortable with allowing the panic. In the past, if we wanted to prevent the panic while still being able to track problems in production we would do something like this:

```
log.WithLevel(zerolog.Panic).Err(err).Str("type", fmt.Sprintf("%T", user).Msg("unexpected user type")
```

This would log the problem and send a slack message to our #gds-critical channel; but now that we have Sentry, just panicking is a bit easier.

pkg/bff/auth/auth_test.go
bbengfort: https://app.shortcut.com/rotational/story/6417/unify-bff-string-constants-in-config

pkg/bff/auth/auth_test.go
```diff
@@ -131,6 +135,38 @@ func TestAuthenticate(t *testing.T) {
...
+	success := func(c *gin.Context) {
+		c.JSON(http.StatusOK, api.Reply{Success: true})
+	}
+
+	// Create a valid token to authenticate
+	tks, err := srv.NewToken()
+	require.NoError(t, err, "could not create valid token")
+
+	// Execute request expecting success
+	c, s, w := createTestContext(http.MethodGet, "/", nil, authenticate, success)
```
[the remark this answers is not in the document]
bbengfort: `c` is pretty standard for the [`gin.Context`](https://github.com/gin-gonic/gin#quick-start) and I don't want to confuse it with a `context.Context` which is usually `ctx`. Similarly, `w` is standard for [`http.HandlerFunc`](https://pkg.go.dev/net/http#HandlerFunc.ServeHTTP). 

`s` is not standard for the router so I could change that to `srv` or `mux` if it would make it easier to read and add a comment to make clearer. Do you have other suggestions for other names? We can create a new PR to make the code more self-documenting if you do.

pkg/bff/auth/authtest/server.go
```diff
@@ -0,0 +1,141 @@
...
+	claims := &Claims{
+		RegisteredClaims: jwt.RegisteredClaims{
+			Issuer:  issuer,
+			Subject: UserID,
+			Audience: jwt.ClaimStrings{
+				issuer,
+				Audience,
+			},
+			IssuedAt:  jwt.NewNumericDate(time.Now()),
+			ExpiresAt: jwt.NewNumericDate(time.Now().Add(15 * time.Minute)),
```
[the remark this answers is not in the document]
bbengfort: I'm not sure there is any standard for the amount of time claims expire, it mostly depends on the expected use case. Generally these systems hand out two tokens, an access token and a refresh token. The access token ExpiresAt and refresh token NotBefore overlap so that a user can use the refresh token to get a new access token without having to resubmit their username/password. 

The access token expiration acts as an "automatic logout" - so you can combine timings to get a desired application outcome. For example, if you have a high request frequency (maybe something like sensors) you could do an access token expiration of 16 minutes and a refresh token not before of 12 minutes and you're guaranteed that the system will stay online and authenticated but constantly recycling access tokens to prevent replay attacks. 

For web applications, I've generally stuck with 1 hour and 45 minutes so that idle users will get logged out in a reasonable time but if you're actively using the app you'll stay logged in. If the app isn't high security though, then a 24 hour access token would probably work well.

pkg/bff/auth/authtest/server.go
bbengfort: That's correct; Auth0 has to be configured to put specific claims into the access token.

pkg/bff/config/config.go
```diff
@@ -125,36 +129,38 @@ func (c AuthConfig) Validate() error {
...
-			return u, nil
-		}
-
-		// If an invalid URL is passed in with an unexpected scheme or path, error.
-		if u.Scheme != "" && (u.Path != "" || u.Host == "") {
-			return nil, errors.New("invalid configuration: could not parse issuer url")
-		}
+	// Do not allow the domain to be a URL -- this is a very basic check
+	if strings.HasSuffix(c.Domain, "/") || strings.HasPrefix(c.Domain, "http://") || strings.HasPrefix(c.Domain, "https://") {
+		return nil, errors.New("invalid configuration: auth0 domain must not be a url or have a trailing slash")
```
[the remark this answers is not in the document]
bbengfort: That's actually a good point, the domain should not have slashes at all -- URLs with and without slashes are constructed from the domain throughout the code base; it should just be `example.us.auth0.com` without a scheme, slash, etc.

pkg/bff/users.go
bbengfort: Yes, the roles have to be configured in Auth0 with the names specified. Auth0 also doesn't have any notion of a default role, unfortunately.

PR #572, opened 2022-06-08
sc-5170 gds apis sentry integration

pdeziel:
### Scope of changes

Integrate Sentry with the GDS API, Admin API, and Members API

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

To test this change:
1. Set required environment variables for each component; the sentry DSN values for each can be found on the settings page in the sentry dashboard
2. Trigger calls via the gds CLI and verify panics and perf tracing shows up in the sentry dashboard

### Author checklist

### Reviewer(s) checklist


pkg/gds/config/config.go
```diff
@@ -21,7 +22,7 @@ type Config struct {
 	Maintenance bool                `split_words:"true" default:"false"`
 	LogLevel    logger.LevelDecoder `split_words:"true" default:"info"`
 	ConsoleLog  bool                `split_words:"true" default:"false"`
-	GDS         GDSConfig
+	API         GDSConfig
```
pdeziel, line 25: This is to avoid weird environment variables such as `GDS_GDS_`, luckily I didn't have to update many other places.

pkg/gds/admin.go
```diff
@@ -48,6 +50,22 @@ func NewAdmin(svc *Service) (a *Admin, err error) {
...
+			Environment:      a.conf.Sentry.Environment,
+			Release:          fmt.Sprintf("gds-admin-api@%s", a.conf.Sentry.GetReleaseVersion()),
+			AttachStacktrace: true,
+			Debug:            a.conf.Sentry.Debug,
+			TracesSampleRate: a.conf.Sentry.SampleRate,
+		}); err != nil {
+			return nil, fmt.Errorf("could not initialize sentry: %w", err)
+		}
+
+		log.Info().Bool("track_performance", a.conf.Sentry.TrackPerformance).Float64("sample_rate", a.conf.Sentry.SampleRate).Msg("GDS admin api sentry tracing is enabled")
```
bbengfort, line 66: Does this log message duplicate the log message here?

https://github.com/trisacrypto/directory/blob/main/pkg/bff/trace.go#L11

pdeziel: In this PR I've moved the log message from that function to here.

bbengfort: That's fine - I'm not seeing that in the PR though; does something stray still need to be pushed up?

pkg/gds/config/config.go
bbengfort: Makes sense, the other two environment variables, `Enabled` and `BindAddr` had directly specified environment variables, but this is probably more extensible.

pkg/gds/interceptor.go
```diff
@@ -36,7 +64,14 @@ func (s *Service) serverInterceptor(ctx context.Context, in interface{}, info *g
 	}
 
 	// Call the handler to finalize the request and get the response.
+	var span *sentry.Span
+	if trackPerfomance {
+		span = sentry.StartSpan(ctx, info.FullMethod)
```
bbengfort, line 69: How does Sentry differentiate between the members DSN and the API DSN here?

bbengfort (commented):
Some comments inline below - I think I might be missing some code in the PR? Is everything pushed up?

pkg/gds/admin.go
pdeziel: I moved `TrackPerformance` from `trace.go` to the utils package `sentry.go`.

pkg/gds/admin.go
```diff
@@ -137,19 +155,48 @@ func (s *Admin) Shutdown() (err error) {
...
+		// Maintenance mode handling - does not require authentication
+		s.Available(),
+	}
+
+	// Add the middleware to the router
+	for _, middleware := range middlewares {
+		if middleware != nil {
+			s.router.Use(middleware)
+		}
+	}
```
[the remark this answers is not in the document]
pdeziel: We did it in the BFF.

pkg/gds/interceptor.go
```diff
@@ -18,10 +21,35 @@ func (s *Service) serverInterceptor(ctx context.Context, in interface{}, info *g
...
+		if s.conf.Members.Sentry.Enabled {
+			trackPanics = true
+			if s.conf.Members.Sentry.TrackPerformance {
+				trackPerfomance = true
+			}
+		}
+	default:
+		log.WithLevel(zerolog.PanicLevel).Err(fmt.Errorf("unknown service type: %T", info.Server))
+		return nil, status.Error(codes.Unimplemented, "unknown service type for request")
+	}
```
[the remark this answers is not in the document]
pdeziel: I think that's reasonable, this part of the code is definitely a bit weird and since we're using the same interceptor anyway for these it seems that they should be considered together in terms of error and performance tracking.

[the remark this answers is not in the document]
pdeziel: Actually scratch that, I think the best thing to do for now is to put the monolith (GDS API, Admin API, Members API) into its own Sentry. It simplifies the configuration a lot and it looks like it's more difficult than I originally thought to separate them. If we feel we need to separate them then we can create a separate story for that.

pkg/gds/interceptor.go
pdeziel: I don't think it actually does, at least the way Sentry is configured in the current state of this PR, so I think that's another reason to combine them for now.

bbengfort (approved):
A couple of minor comments here and there - and apparently I'm reviewing things without my coffee kicking in because I was having a hard time tracking file changes! I'm glad we did the research to see what the code looked like with the per-component sentry, it makes my much more confident in this Sentry configuration!

PR #578, opened 2022-06-09
sc-4310 Initial bff overview endpoint

pdeziel:
### Scope of changes

This changes the bff overview endpoint to call the members API instead of returning the mock data. In order to do this, I created a `Summary` members endpoint which returns some summary information. In a future PR we need to parse the VASP IDs out of the user claims and do specific VASP lookups, as well as implement the overview test in the bff when authentication is fully integrated.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

I haven't manually tested the endpoint yet but it should now return real data from the configured testnet and mainnet.

### Author checklist

### Reviewer(s) checklist


pkg/bff/members.go
```diff
@@ -1,39 +1,134 @@
...
+// configuration. This method is used to connect to both the TestNet and the MainNet and
+// to connect to mock GDS services in testing using buffconn.
+func ConnectMembers(conf config.DirectoryConfig) (_ members.TRISAMembersClient, err error) {
+	// Create the Dial options with required credentials
+	var opts []grpc.DialOption
+	if conf.Insecure {
+		opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	} else {
+		opts = append(opts, grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{})))
+	}
```
bbengfort, line 31: The Members server should be mTLS authenticated, so I don't think the only reason to connect insecure is to connect to a bufconn; but I've preferred to do that recently by adding dial opts to the connect members function so that in the test path we simply pass in our own options. The connect function checks if the passed in dial opts are empty, and if so, adds the mTLS connection from the configuration.

pkg/bff/mock/members.go
```diff
@@ -0,0 +1,160 @@
...
+// storage and managing fixtures.
+//
+// To set the response of the mock for a particular test, update the Members OnRPC method.
+// e.g. to mock the Summary RPC, set the OnSummary method. The number of calls to the
+// RPC will be recorded to verifiy that the service is being called correctly. Use the
+// Reset method to remove all RPC handlers and set calls back to 0.
+//
+// NOTE: this mock is not safe for concurrent use.
+// NOTE: if the OnRPC function is not set, the test will panic
+type Members struct {
```
bbengfort, line 48: Nice mock :-)

pkg/bff/parallel.go
```diff
@@ -36,8 +37,46 @@ func (s *Server) ParallelGDSRequests(ctx context.Context, rpc RPCFunc, flatten b
...
+// MembersRPC allows the BFF to issue arbitrary members client methods in parallel to both the
+// testnet and the mainnet. The members client and network are passed into the function,
+// allowing the RPC to make any GDS RPC call and log with the associated network.
+type MembersRPC func(ctx context.Context, client members.TRISAMembersClient, network string) (proto.Message, error)
+
+// ParallelMembersRequest makes concurrent requests to both the testnet and the mainnet,
+// storing the results and errors in a slice of length 2 ([testnet, mainnet]). If the
+// flatten bool is true, then nil values are removed from the slice (though this will
+// make which network returned the result ambiguous).
+func (s *Server) ParallelMembersRequests(ctx context.Context, rpc MembersRPC, flatten bool) (results []proto.Message, errs []error) {
```
bbengfort, line 60: Is the only difference here the client interface? If so, we could probably just create a super client interface combining both the interfaces into one, then embedding the clients into the `NetworkClient` so that it acted as clients to both.

bbengfort (approved):
Wow that was quite a lot of work to get the members client integrated into the BFF and add new members endpoints! I think this is a great foundation for this code, but before we can actually put it into use, we'll have to add mTLS configuration. I think there may be a way we can simplify the code a bit as well, but that can be for the future. 

Since this code is conflicting with the PR I'm working on, I'm going to go ahead and merge so I can deal with the merge conflicts. Would you mind creating follow on stories from both of our comments in the code?

PR #582, opened 2022-06-14
sc-6285 additional sentry tracing metadata

pdeziel:
### Scope of changes

Since GDS runs a few different servers from the same program, we previously didn't have an easy way of distinguishing errors and performance traces from the different components. This adds some tags to the sentry collections which are searchable from the dashboard. This also adds some additional metadata for HTTP requests such as the method and path and sets the transaction name properly for performance tracing.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Transactions should have a name in the Sentry dashboard now, and if we receive a panic from any of the components there should be a populated `service` tag which indicates which of the components the error originated.

### Author checklist

### Reviewer(s) checklist


pkg/bff/server.go
```diff
@@ -227,15 +227,25 @@ func (s *Server) SetURL(url string) {
...
 
-	var tracing gin.HandlerFunc
+	if s.conf.Sentry.UseSentry() {
+		bffTags = map[string]string{"service": "bff"}
+		tags = sentry.UseTags(bffTags)
+	}
+
 	if s.conf.Sentry.UsePerformanceTracking() {
-		tracing = sentry.TrackPerformance()
+		tracing = sentry.TrackPerformance(bffTags)
```
pdeziel, line 248: We don't really need the `bff` service tag since it's in a separate project, but I added it for consistency and it acts as a placeholder for any tags we would want to define here.

pkg/gds/interceptor.go
```diff
@@ -42,7 +59,7 @@ func (s *Service) serverInterceptor(ctx context.Context, in interface{}, info *g
 	// Call the handler to finalize the request and get the response.
 	var span *sentry.Span
 	if s.conf.Sentry.UsePerformanceTracking() {
-		span = sentry.StartSpan(ctx, info.FullMethod)
+		span = sentry.StartSpan(ctx, "grpc handler", sentry.TransactionName(info.FullMethod))
```
pdeziel, line 62: The way I'm interpreting this from the documentation is that a `transaction` represents the full request in our case, and the `operation` is a specific part of the request handler. So if we had the need to more finely track request performance we would add more spans with the same transaction name, but for now the span is just the entire handler.

bbengfort: That makes sense to me - for something like the BFF I could see breaking it down into the overall request and the request to mainnet vs testnet; but for most RPCs it's just the handler. 

One minor thing; would we just name the operation `grpc` just so it's a bit more digestible when we see it many times in Sentry? 

```suggestion
		span = sentry.StartSpan(ctx, "grpc", sentry.TransactionName(info.FullMethod))
```

pkg/utils/sentry/trace.go
```diff
@@ -28,15 +29,38 @@ func Init(conf Config) (err error) {
...
 }
 
-// Middleware that tracks request performance with Sentry.
-func TrackPerformance() gin.HandlerFunc {
+// Gin middleware that tracks HTTP request performance with Sentry.
+func TrackPerformance(tags map[string]string) gin.HandlerFunc {
 	return func(c *gin.Context) {
-		span := sentry.StartSpan(c.Request.Context(), c.Request.URL.Path)
+		request := fmt.Sprintf("%s %s", c.Request.Method, c.Request.URL.Path)
+		span := sentry.StartSpan(c.Request.Context(), "http handler", sentry.TransactionName(request))
```
bbengfort, line 36: Similarly - to distinguish between the API types name the operations `grpc` and `rest`?

```suggestion
		span := sentry.StartSpan(c.Request.Context(), "rest", sentry.TransactionName(request))
```

bbengfort (approved):
I think this is going to make it a lot easier to be able to understand and diagnose what's going on in Sentry, thanks for adding the tags!

PR #584, opened 2022-06-15
sc-5942 Add certificate model

pdeziel:
### Scope of changes

The point of this PR is to nail down the certificate model structure to enable future certificate management functionality.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Does the certificate model make sense? Do we need to include any more metadata fields based on what we plan to accomplish for the certificates management effort?

### Author checklist

### Reviewer(s) checklist


proto/gds/models/v1/models.proto
```diff
@@ -3,8 +3,34 @@ syntax="proto3";
...
+
+    // Certificate details
+    trisa.gds.models.v1beta1.Certificate details = 5;
+}
+
+enum CertificateState {
+    ISSUED = 0;
+    EXPIRED = 1;
+    REVOKED = 2;
+}
```
bbengfort, line 32: I go back and forth on whether the zero-valued enum (which will be the defacto-default when initializing the Certificate struct) should be a "good" value (e.g. ISSUED probably is the default) or if should be a true zero-value, e.g. UNKNOWN to force the user to specify the status rather than make mistakes with the zero value. Thoughts?

```suggestion
enum CertificateState {
    UNKNOWN = 0;
    ISSUED  = 1;
    EXPIRED = 2;
    REVOKED = 3;
}
```

bbengfort (approved):
Nice! This will certainly help us get started storing certificate records for the BFF!

proto/gds/models/v1/models.proto
pdeziel: Right, my thought was that GDS only needs to create these certificate records after they have already been issued, and there's no alternative initial state. However I do see the other side; it's a bit less clear from the code what state the certificate is in when it's created. I'm leaning towards keeping `ISSUED` as the initial state for right now but perhaps we should define a best practice for our future projects.

PR #587, opened 2022-06-16
Fetch VASP data on overview

pdeziel:
### Scope of changes

This modifies the GDS members `Summary` endpoint to return VASP members info for a requested VASP, and also changes the `Overview` endpoint in the BFF to retrieve the VASP ID from the auth claims in order to populate VASP details for the response.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Members `Summary` endpoint should now return VASP details if requested. We still need to write tests for the `Overview` endpoint in a future PR.

### Author checklist

### Reviewer(s) checklist


pkg/bff/members.go
```diff
@@ -78,16 +84,27 @@ func (s *Server) GetSummaries(ctx context.Context) (testnet *members.SummaryRepl
...
+	var claims *auth.Claims
+	if claims, err = auth.GetClaims(c); err != nil {
+		log.Error().Err(err).Msg("unable to retrieve bff claims from context")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse(err))
+		return
+	}
+
+	// Extract the VASP IDs from the claims
+	testnetID := claims.VASP["testnet"]
+	mainnetID := claims.VASP["mainnet"]
```
pdeziel, line 99: Is the expected behavior for the claims that this field will be empty if the registration hasn't been submitted for testnet or mainnet?

pkg/bff/members.go
```diff
@@ -122,12 +140,50 @@ func (s *Server) Overview(c *gin.Context) {
...
+
+			out.TestNet.MemberDetails = api.MemberDetails{
+				ID:          testnet.Vasp.Id,
+				Status:      testnet.Vasp.Status.String(),
+				CountryCode: testnet.Vasp.Country,
+			}
+		} else {
+			out.TestNet.MemberDetails = api.MemberDetails{
+				Status: pb.VerificationState_NO_VERIFICATION.String(),
+			}
```
pdeziel, line 160: I guess the frontend will have to check for this status to determine existence, should we make it more explicit?

pkg/bff/members.go
```diff
@@ -122,12 +140,50 @@ func (s *Server) Overview(c *gin.Context) {
...
+			if mainnet.Vasp == nil {
+				log.Error().Msg("could not retrieve mainnet VASP details")
+				c.JSON(http.StatusInternalServerError, api.ErrorResponse(fmt.Errorf("could not retrieve mainnet VASP details")))
+				return
+			}
+
+			out.MainNet.MemberDetails = api.MemberDetails{
+				ID:          mainnet.Vasp.Id,
+				Status:      mainnet.Vasp.Status.String(),
+				CountryCode: mainnet.Vasp.Country,
```
pdeziel, line 180: We could return more of the members information we get from `Summary`, but this is just what was originally defined for the overview endpoint.

pkg/bff/api/v1/api.go
```diff
@@ -46,21 +46,22 @@ type StatusReply struct {
...
+	Vasps              int           `json:"vasps"`
+	CertificatesIssued int           `json:"certificates_issued"`
+	NewMembers         int           `json:"new_members"`
+	MemberDetails      MemberDetails `json:"member_details"`
 }
 
-// VaspDetails contains VASP-specific information.
-type VaspDetails struct {
+// MemberDetails contains VASP-specific information.
+type MemberDetails struct {
```
bbengfort, line 64: I do like this name a little better - makes the code more readable.

pkg/bff/members.go
```diff
@@ -44,13 +46,18 @@ func ConnectMembers(conf config.DirectoryConfig) (_ members.TRISAMembersClient,
...
-		return client.Summary(ctx, &members.SummaryRequest{})
+func (s *Server) GetSummaries(ctx context.Context, testnetID, mainnetID string) (testnet *members.SummaryReply, mainnet *members.SummaryReply, err error) {
+	// Create the RPCs with the VASP ID parameters
+	makeRPC := func(req *members.SummaryRequest) MembersRPC {
+		return func(ctx context.Context, client members.TRISAMembersClient, network string) (rep proto.Message, err error) {
+			return client.Summary(ctx, req)
+		}
 	}
+	testnetRPC := makeRPC(&members.SummaryRequest{Vasp: testnetID})
+	mainnetRPC := makeRPC(&members.SummaryRequest{Vasp: mainnetID})
```
bbengfort, line 57: Nice factory for the closure!

pkg/bff/members.go
```diff
@@ -44,13 +46,18 @@ func ConnectMembers(conf config.DirectoryConfig) (_ members.TRISAMembersClient,
...
 
 	// Perform the parallel requests
-	results, errs := s.ParallelMembersRequests(ctx, rpc, false)
+	results, errs := s.ParallelMembersRequests(ctx, testnetRPC, mainnetRPC, false)
```
bbengfort, line 60: Originally I had envisioned the `rpc` function switching on the `network` inside of the closure; e.g. something more like:

```go
rpc := func(ctx context.Context, client members.TRISAMembersClient, network string) (rep proto.Message, err error) {
    rep := &members.SummaryRequest{}
    switch network {
    case testnet:
        rep.Vasp: testnetID
    case mainnet:
        rep.Vasp: mainnetID
    default:
        return nil, fmt.Errorf("unknown network %q", network)
    }
    return client.Summary(ctx, req) 
}
```

If you think the factory to create closures is the better way; perhaps we should change all of our `Parallel` helper functions to that style. What do you think?

pkg/bff/members.go
bbengfort: Yes, that is the expected behavior - and there can be many reasons why the field would be empty; for example, the user may have to log out and log back in in order for the claims to be correctly populated.

pkg/bff/members.go
```diff
@@ -110,7 +128,7 @@ func (s *Server) Overview(c *gin.Context) {
 
 	// Get the summaries for both testnet and mainnet
 	var testnet, mainnet *members.SummaryReply
-	testnet, mainnet, err = s.GetSummaries(context.Background())
+	testnet, mainnet, err = s.GetSummaries(context.Background(), testnetID, mainnetID)
```
bbengfort, line 131: Instead of `context.Background()` I think we should put in the context from `c` so that a timeout on the client request also cancels the two RPCs to the GDS (I think it's `c.Request.Context()`). Sorry I didn't catch this on the last PR.

pkg/bff/members.go
bbengfort: I think it might be better to just leave the `MemberDetails` as `nil` and then we'll have the front-end check for `null` before populating the component. @masskoder thoughts?

pkg/bff/members.go
bbengfort: Also, I might change these to `testsum` and `mainsum` so you don't override the `testnet` and `mainnet` constants.

pkg/bff/members.go
bbengfort: Any reason not to use the string constants for these?

pkg/gds/members.go
```diff
@@ -270,11 +249,21 @@ func (s *Members) Summary(ctx context.Context, in *api.SummaryRequest) (out *api
 		since = time.Now().Add(-time.Hour * 24 * 30)
 	}
 
-	// TODO: Add paramater to fetch details for a specific VASP
-
 	// Create response
 	out = &api.SummaryReply{}
 
+	if in.Vasp != "" {
```
bbengfort, line 255: Nice - so supplying the `Vasp` is completely optional and if it's omitted from the request then the `Vasp` in the response is just `nil` - nice one! I haven't looked at the tests yet, but we'll want an explicit test for this since the `BFF` depends on this behavior (to prevent regressions)

proto/gds/members/v1alpha1/members.proto
```diff
@@ -62,6 +62,9 @@ message VASPMember {
 message SummaryRequest {
     // The start date for determining how many members are new - optional
     string since = 1;
+
+    // The VASP ID to retrieve details for - optional
+    string vasp = 2;
```
bbengfort, line 67: Should we change this to `member_id` to match your other changes? If not, can we at least make it `vasp_id`?

```suggestion
    // Include your VASP ID for details about your VASP record included with the summary - optional
    string vasp_id = 2;
```

proto/gds/members/v1alpha1/members.proto
```diff
@@ -70,4 +73,7 @@ message SummaryReply {
     int32 vasps = 1;
     int32 certificates_issued = 2;
     int32 new_members = 3;
+
+    // Details for the requested VASP
+    VASPMember vasp = 4;
```
bbengfort, line 78: Same as above, should we change this to `member_info` or something similar? Does that make it more clear or does it just make "member" a confusing synonym with `vasp`?

bbengfort (approved):
I've left a couple of comments for discussion, some of which we may create new stories for, but which will be important for us to figure out so we can get the rest of the BFF endpoints done. 

We should create a story for documenting the Members summary endpoint in the TRISA docs. 

Also, another point of concern: there is nothing preventing the user from adding any VASP ID to the summary request and getting their member info out (in fact, the BFF relies on this). None of this information is sensitive but we should note that behavior; the members API is authenticated but is available publicly (e.g. to callers outside of the BFF). 

We do need to create a Members Detail endpoint as well so that the VASP details including the IVMS 101 legal person can be fetched by VASPs securely. I thought there was a story for this but couldn't find it. Before we close this PR can we locate that story or create a new one?

pkg/bff/members.go
pdeziel: Good catch

pkg/bff/members.go
pdeziel: I see, I completely forgot about the network parameter but I think that's less clunky than the factory method for now.

bbengfort: Ok, so in the future, we'll stick with the `network` switch rather than a factory method and passing two RPCs into the `Parallel`?

proto/gds/members/v1alpha1/members.proto
pdeziel: I agree that's a better name since we're in the members definitions. I think we've been using them interchangeably in a few places but here it does make a lot of sense to distinguish from the `vasps` parameter.

pkg/gds/members.go
pdeziel: Thanks, I'll add a check to the test to assert that specifically.

PR #590, opened 2022-06-17
Add certificate store

pdeziel:
### Scope of changes

This adds a storage interface for the certificate model and updates the certificate request handler to write the model to the database (namespace: `certs`) after certificates are downloaded.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

This PR was a lot of copy + paste, we do have pretty good test coverage for the storage interfaces and database interactions but it would be good for a second pair of eyes to check to make sure I didn't miss a piece.

### Author checklist

### Reviewer(s) checklist


pkg/gds/models/v1/models.go
```diff
@@ -225,6 +225,35 @@ func AppendCertID(vasp *pb.VASP, certID string) (err error) {
...
+
+	// Certificate request must be not nil.
+	if certRequest == nil {
+		return nil, errors.New("must supply a certificate request for certificate creation")
+	}
+
+	// VASP must contain an identity certificate.
+	if vasp.IdentityCertificate == nil {
+		return nil, errors.New("VASP must contain an identity certificate")
+	}
```
bbengfort, line 244: Not all certificates will be identity certificates - we plan on also issuing sealing certificates that are used for secure envelope PKI. As such I don't think we should rely on the `vasp.IdentityCertificate` for parsing the protocol buffers but should use the same method in the certman code to extract the protocol buffers from the downloaded certificate.

pkg/gds/models/v1/models.go
```diff
@@ -225,6 +225,35 @@ func AppendCertID(vasp *pb.VASP, certID string) (err error) {
...
+
+	cert = &Certificate{
+		Id:      uuid.New().String(),
```
bbengfort, line 247: The way that we identify the certificate in Sectigo is via the capital hex-encoded serial number; may I propose that we use that serial number here instead of a random UUID? Or if you think we should be more specific, could we include a Serial Number field at the top level that ensures the certificate serial number is rendered in such a way that it can be looked up using the Sectio UI?

proto/gds/models/v1/models.proto
```diff
@@ -19,7 +19,7 @@ message Certificate {
     string vasp = 3;
 
     // Current status of the certificate
-    string status = 4;
+    CertificateState status = 4;
```
bbengfort, line 22: Oops - sorry I missed that on the last code review! Good catch!

pkg/gds/store/store.go
```diff
@@ -95,15 +96,24 @@ type DirectoryStore interface {
 	DeleteVASP(id string) error
 }
 
-// CertificateStore describes how the service interacts with Certificate requests.
-type CertificateStore interface {
-	ListCertReqs() iterator.CertificateIterator
+// CertificateRequestStore describes how the service interacts with Certificate requests.
+type CertificateRequestStore interface {
```
bbengfort, line 100: I suppose I thought this name change would reach out and touch a lot of code but I guess the DB just implements all these interfaces so it's really just the inside of the `store` package that needs renaming - big win!

bbengfort (approved):
Nice! Thank you so much for doing the legwork to add this new model. I appreciate how thorough you were including both `leveldb` and `trtl` store methods and you also took care to update the `wire` package -- that gives me a lot of confidence that there aren't any missing places in the code where we'll need to add `CertificateStore` helpers to!

pkg/gds/models/v1/models.go
pdeziel: I pushed some changes to use the serial number instead of a random ID, can you check to make sure I interpreted your suggestions correctly?

PR #597, opened 2022-06-22
GDS BFF Network Announcement Endpoint

bbengfort:
### Scope of changes

This goal of this PR was to add a `GET /v1/announcements` endpoint to return network announcements for the dashboard. The design requested that we store these announcements in the BFF database. I extended the work for this to include a `POST /v1/announcements` endpoint to allow TRISA admins to create network announcements to display on the dashboard. This further required me to create the `create:announcements` permission in Auth0 and the "TRISA Admins` role so that only admins can create announcements. 

As I got into the code I realized that we aren't fully using the database module quite yet, so I had to put in some effort to exploring how we utilize the database. 

I came up with the idea of "Collections" which are like the GDS `Store` object that allow you to interact with a specific namespace and load and save specific models. 

The "model" for this task is the `Announcement` struct which is defined in `api/v1` -- this is not really the right place to define models and if the reviewer thinks I should create a protocol buffer for this I'm happy to. The shortcut allowed us to return exactly what the endpoint expects, but may cause problems in the future. Since I didn't use a protocol buffer, I decided to serialize the Announcement object using gzip compressed json. 

The `Announcements` collection can be fetched off of the db and operated on using chained methods, e.g.  `db.Announcements().Post()` or `db.Announcements().Recent()`. There is a `sync.Once` that ensures that the collection is only instantiated once, but is not instantiated if the code is never used. The collection object is thread-safe. 

I created a database test suite modeled off of our GDS test suite - running a `trtl` server with a leveldb in a temp dir and connecting to it via bufconn. I implemented tests to extend the coverage of the database. 

Now for the last problem: my idea for database storage was to create time ordered keys using ksuids. This causes two problems:

1. In order to effectively query this data structure we need to seek in reverse order since the keys are stored in ascending time order … but `trtl` doesn't offer this functionality. 
2. ksuids seem to be concurrent within 1 second granularity, so to load 10 ordered announcements into the database we need to sleep 1 second between each … causing a 10 second long test. Not ideal. 

Right now the entire set of announcements is loaded into memory, but this is _far_ from ideal. We need to figure out if we should use a different data structure, implement reverse cursor in trtl, or what. Much discussion is required. 

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

- [ ] Is announcement spelled correctly? I did do a spell check … but I got it wrong every time I typed it 
- [ ] What do we do about the recent query in trtl? Add trtl reverse cursor, some other data structure?
- [ ] Do we need to go to a protobuf or is the API "model" shortcut ok
- [ ] What do we think about cursors for the BFF database interaction?
- [ ] How will admins create announcements, should we create a CLI util for it? Something in the Admin UI? 

### Author checklist

### DIFF
--- a/pkg/bff/announcements.go
+++ b/pkg/bff/announcements.go
@@ -11,13 +11,13 @@ import (
 )
 
 const (
-	maxAnnouncements    = 10
-	announcementsWindow = -1 * 31 * 24 * time.Hour
+	maxAnnouncements = 10
+	subMonths        = -1
 )
 
 func (s *Server) Announcements(c *gin.Context) {
-	// Only fetch the previous 10 announcements from the last 30 days
-	nbf := time.Now().Add(announcementsWindow)
+	// Only fetch the previous 10 announcements from the last month
+	nbf := time.Now().AddDate(0, subMonths, 0)
 	out, err := s.db.Announcements().Recent(c.Request.Context(), maxAnnouncements, nbf)
 	if err != nil {
 		log.Error().Err(err).Msg("could not fetch recent announcements")
--- a/pkg/bff/announcements.go
+++ b/pkg/bff/announcements.go
@@ -60,7 +60,7 @@ func (s *Server) MakeAnnouncement(c *gin.Context) {
 	}
 
 	if claims.Email == "" {
-		log.Warn().Msg("missing email on claims cannot set author of network announcement")
+		log.Warn().Msg("missing email on claims, cannot set author of network announcement")
 		c.JSON(http.StatusInternalServerError, api.ErrorResponse("user claims are not correctly configured"))
 		return
 	}
--- a/pkg/bff/announcements.go
+++ b/pkg/bff/announcements.go
@@ -1,13 +1,81 @@
 package bff
 
 import (
+	"net/http"
+	"time"
+
 	"github.com/gin-gonic/gin"
+	"github.com/rs/zerolog/log"
+	"github.com/trisacrypto/directory/pkg/bff/api/v1"
+	"github.com/trisacrypto/directory/pkg/bff/auth"
+)
+
+const (
+	maxAnnouncements    = 10
+	announcementsWindow = -1 * 31 * 24 * time.Hour
 )
 
 func (s *Server) Announcements(c *gin.Context) {
+	// Only fetch the previous 10 announcements from the last 30 days
+	nbf := time.Now().Add(announcementsWindow)
+	out, err := s.db.Announcements().Recent(c.Request.Context(), maxAnnouncements, nbf)
+	if err != nil {
+		log.Error().Err(err).Msg("could not fetch recent announcements")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("unable to fetch recent announcements"))
+		return
+	}
 
+	// If the database is empty, set the last updated timestamp to now.
+	if len(out.Announcements) == 0 && out.LastUpdated == "" {
+		out.LastUpdated = time.Now().Format(time.RFC3339)
+	}
+
+	// Return the results
+	c.JSON(http.StatusOK, out)
 }
 
 func (s *Server) MakeAnnouncement(c *gin.Context) {
+	var (
+		id     string
+		err    error
+		claims *auth.Claims
+		post   *api.Announcement
+	)
 
+	if err = c.BindJSON(&post); err != nil {
+		log.Warn().Err(err).Msg("could not parse announcement post data")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("could not parse announcement JSON data"))
+		return
+	}
+
+	if post.PostDate != "" || post.Author != "" {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("cannot set the post_date or author fields on the post"))
+		return
+	}
+
+	if claims, err = auth.GetClaims(c); err != nil {
+		log.Error().Err(err).Msg("could not fetch claims from request")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not save announcement"))
+		return
+	}
+
+	if claims.Email == "" {
+		log.Warn().Msg("missing email on claims cannot set author of network announcement")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("user claims are not correctly configured"))
+		return
+	}
+
+	// Set the post date and the author
+	post.PostDate = time.Now().Format("2006-01-02")
+	post.Author = claims.Email
+
+	if id, err = s.db.Announcements().Post(c.Request.Context(), post); err != nil {
+		log.Error().Err(err).Msg("could not put announcement to trtl database")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not save announcement"))
+		return
+	}
+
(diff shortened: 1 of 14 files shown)


pkg/bff/announcements.go
```diff
@@ -0,0 +1,81 @@
...
+	if id, err = s.db.Announcements().Post(c.Request.Context(), post); err != nil {
+		log.Error().Err(err).Msg("could not put announcement to trtl database")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not save announcement"))
+		return
+	}
+
+	// Return a 204 No Content to indicate the post happened successfully
+	log.Info().Str("id", id).Str("title", post.Title).Str("author", post.Author).Msg("network announcement added")
+	c.JSON(http.StatusNoContent, nil)
+}
```
bbengfort, line 81: These endpoints are currently untested - it's now 1 AM and I think we need to make another story for testing this since we need to figure out how we want to mock the database. Most of the primary functionality is tested in the `db` module and I did manually test it locally so it should work to unblock @masskoder.

pkg/bff/api/v1/api.go
```diff
@@ -69,6 +71,22 @@ type MemberDetails struct {
...
+
+// Announcement represents a single network announcementthat can be posted to the
+// endpoint or returned in the announcements reply.
+type Announcement struct {
+	Title    string `json:"title"`
+	Body     string `json:"body"`
+	PostDate string `json:"post_date,omitempty"` // Ignored on POST only available on GET
+	Author   string `json:"author,omitempty"`    // Ignored on POST only available on GET
+}
+
```
bbengfort, line 89: This is serving as the database model as a shortcut that couples the db to the RPC handler.

pkg/bff/db/announcements.go
```diff
@@ -0,0 +1,215 @@
+package db
```
bbengfort, line 1: This is most of the code in the PR -- database interactions with a new "collection" approach to the key-value store as a document database.

pkg/bff/db/db_test.go
```diff
@@ -0,0 +1,106 @@
...
+	suite.Suite
+	db   *db.DB
+	trtl *trtl.Server
+	conn *bufconn.GRPCListener
+}
+
+func TestDatabaseSuite(t *testing.T) {
+	s := new(dbTestSuite)
+	suite.Run(t, s)
+}
```
bbengfort, line 32: Test suite style matches our mechanism of testing GDS with Trtl.

pkg/bff/db/announcements_test.go
```diff
@@ -0,0 +1,150 @@
...
+		require.NoError(err, "could not post announcement %d", i)
+		require.NotEmpty(id, "expected an ordered kid to be assigned")
+		ids = append(ids, id)
+
+		// Ensure there is at least 1 second between each post or the ksuid will
+		// identify the posts as concurrent and produce an incorrect ordering.
+		// Normally posts won't be published within the same second, but this is a bad
+		// situation for tests because it means that this test will take a min of 10
+		// seconds to run, which adds up in CI and beyond ...
+		time.Sleep(1 * time.Second)
```
bbengfort, line 51: Le sigh

pkg/bff/db/announcements.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	defer cursor.CloseSend()
+
+	// Consume the cursor
+	// TODO: this filtering method is NOT ideal since it scans in ascending time order
+	// rather than reverse time order (trtl doesn't support reverse ordering). This
+	// means as the number of announcements in the collection grows, this method will
+	// get slower and slower. Worse, we're loading everything in memory so we might
+	// overload the pod. We need to either figure out a different key-value storage
+	// and ordering mechanism or implement reverse seek in trtl.
```
pdeziel, line 89: Agreed this solution feels a little clunky and relies on some implicit behavior with the ksuids and trtl. Maybe we need something like an "ordered" namespace defined in trtl? Or if we don't care about keeping the older announcements around, make the records ephemeral somehow so that we only store the last month of entries?

pkg/bff/db/announcements.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// Namespace implements the collection interface
+func (a *Announcements) Namespace() string {
+	return a.namespace
+}
+
+// Encode an announcement for storage in the database.
+// HACK: the api definition is probably not the right place to store a database model,
+// should we just create protocol buffer model definitions instead of using this?
```
pdeziel, line 177: Here's a (maybe terrible) idea, if we are hard limiting the number of announcements can we just maintain the entire ordered list in a single object in the database? It opposes the way we are managing the other records but if we keep the size restricted then it simplifies the handling.

pkg/bff/db/announcements_test.go
pdeziel: That is annoying, the only way I can think of to avoid this is to not use the ksuids for ordering.

pdeziel (commented):
Wow, there are definitely some technical challenges here. I don't know what the specific requirements are for the announcements, but one way I can think of to reduce the complexity is to reframe the announcements as ephemeral objects, so we're only keeping a maximum number of them around and discarding the oldest ones. I suppose another solution would be to have a side database that's more tuned towards managing these types of objects, which might be better if we think we'll have a need for bff-related structures that are better managed separate from the VASP and certificate data.

pkg/bff/announcements.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+const (
+	maxAnnouncements    = 10
+	announcementsWindow = -1 * 31 * 24 * time.Hour
+)
+
+func (s *Server) Announcements(c *gin.Context) {
+	// Only fetch the previous 10 announcements from the last 30 days
+	nbf := time.Now().Add(announcementsWindow)
```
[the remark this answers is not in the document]
bbengfort: Ah didn't know about that function - that's definitely more readable!

pkg/bff/db/announcements.go
bbengfort: Technically the namespace is ordered, it's just ordered ascending and we need it ordered descending. If there was a ksuid that was naturally descending that would be ideal, but that's pretty difficult to do. You can think of it as us having an array of posts that are sorted in time order and we want to access the last n posts. 

In terms of ephemerality, we don't really want to lose announcements - it's good to have a record of them. 

One thing I thought of was to have per month collections. e.g. the key would be 2022-06 and then all the announcements for that month would be there. It would reduce the speed of writes (it would always be 2-3 gets and a put) and reading the entire collection of announcements would take longer, but it might solve the problem because we wouldn't have to seek.

pkg/bff/db/announcements.go
bbengfort: That's a very similar idea to storying a list of announcements per month - and I think there is definitely something to it!

pkg/bff/db/announcements_test.go
bbengfort: I know right? So annoying - and it feels like something that should be configurable. I found lots of git issues complaining about it and related issues.

pdeziel (approved):
:shipit:

PR #611, opened 2022-06-29
bff admin client

pdeziel:
### Scope of changes

This adds a GDS Admin API client to the bff, allowing the bff to make authenticated requests to the Admin API such as listing certificates. In order to do this, the client manages its own access tokens in memory which are signed by token keys in the configuration. The idea is that we would use the same token keys 

SC-6644

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

I manually tested this locally with docker compose by adding a command to the bff cli which calls an authenticated admin endpoint, but I didn't commit that code to this PR. I was planning on verifying this further in the PR for completing the bff certificates endpoint since the admin list certificates is merged now.

### Author checklist

### Reviewer(s) checklist


pkg/bff/client.go
```diff
@@ -13,20 +14,36 @@ import (
 // GlobalDirectoryClient is a unified interface which can be implemented to create a
 // unified client to multiple GDS services.
 type GlobalDirectoryClient interface {
+	admin.DirectoryAdministrationClient
```
pdeziel, line 17: I liked the idea of including the admin client in the unified interface as well, although it did cause some method collisions. However since it's HTTP, I can also see it being separated from the other two. What do you think?

pkg/bff/config/config.go
```diff
@@ -76,8 +86,8 @@ type DatabaseConfig struct {
 }
 
 type MTLSConfig struct {
-	CertPath string `split_words:"true" required:"true"`
-	PoolPath string `split_words:"true" required:"true"`
+	CertPath string `split_words:"true"`
+	PoolPath string `split_words:"true"`
```
bbengfort, line 90: Maybe we should move the `Insecure` bool here since the certs aren't required anymore?

pkg/bff/client.go
```diff
@@ -71,6 +88,100 @@ func (c *GDSClient) ConnectMembers(conf config.MembersConfig, opts ...grpc.DialO
...
+
+func (c *GDSClient) Review(ctx context.Context, in *admin.ReviewRequest) (out *admin.ReviewReply, err error) {
+	return c.admin.Review(ctx, in)
+}
+
+func (c *GDSClient) Resend(ctx context.Context, in *admin.ResendRequest) (out *admin.ResendReply, err error) {
+	return c.admin.Resend(ctx, in)
+}
+
+// GDS methods
```
bbengfort, line 184: Because the `APIv2` struct is exported, I think these methods can be added by direct embedding rather than call throughs.

containers/docker-compose.yaml
```diff
@@ -138,12 +140,22 @@ services:
...
+      - GDS_BFF_TESTNET_MEMBERS_INSECURE=true
+      - GDS_BFF_TESTNET_MEMBERS_ENDPOINT=gds:4435
+      - GDS_BFF_MAINNET_DIRECTORY_INSECURE=true
+      - GDS_BFF_MAINNET_DIRECTORY_ENDPOINT=gds-mainnet:8433
+      - GDS_BFF_MAINNET_DIRECTORY_TIMEOUT=5s
+      - GDS_BFF_MAINNET_ADMIN_ENDPOINT=http://gds:8434
+      - GDS_BFF_MAINNET_ADMIN_AUDIENCE=http://localhost:8433
+      - GDS_BFF_MAINNET_ADMIN_TOKEN_KEYS=1y9jUjVqRqRJaiKsNqrZmQkhTqe:/data/creds/current_token_key.pem,1y9jTDgrljWl0iVUOQBcCev7rBG:/data/creds/rotated_token_key.pem
+      - GDS_BFF_MAINNET_MEMBERS_INSECURE=true
+      - GDS_BFF_MAINNET_MEMBERS_ENDPOINT=gds-mainnet:8435
```
bbengfort, line 158: Thank you for updating these! @masskoder this should fix the issue you were having.

cmd/gds/main.go
```diff
@@ -983,7 +983,7 @@ func adminStatus(c *cli.Context) (err error) {
 	defer cancel()
 
 	var rep *admin.StatusReply
-	if rep, err = adminClient.Status(ctx); err != nil {
+	if rep, err = adminClient.AdminStatus(ctx); err != nil {
```
bbengfort, line 986: Does this mean we have some general name spacing issues? Is it just related to the embedding or something else?

pkg/bff/credentials.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// Load signs new access and refresh tokens using the token manager.
+func (c *Credentials) Load(api admin.DirectoryAdministrationClient) (err error) {
+	var accessToken, refreshToken *jwt.Token
+
+	claims := map[string]interface{}{
+		"hd":      "rotational.io",
+		"email":   "bff@rotational.io",
+		"name":    "GDS BFF",
```
bbengfort, line 47: Could we move these to constants at the top of the file please?

pkg/bff/credentials.go
```diff
@@ -0,0 +1,150 @@
...
+// tokens to the GDSClient.
+type Credentials struct {
+	tm      *tokens.TokenManager
+	access  string
+	refresh string
+}
+
+// NewCredentials creates a new Credentials object which can generate new access and
+// refresh tokens for authentication based on the provided token keys and audience.
+func NewCredentials(conf config.AdminConfig) (creds *Credentials, err error) {
```
bbengfort, line 25: Could we move this to a `bff/admin` sub package and change this to `New` -- I'm worried about overloading the main `bff` package and this is nicely self-contained. We could then create a story for follow on tests in a later sprint.

pkg/bff/credentials.go
```diff
@@ -0,0 +1,150 @@
...
+	var rep *admin.AuthReply
+	if rep, err = api.Reauthenticate(ctx, &admin.AuthRequest{Credential: c.refresh}); err != nil {
+		return "", "", err
+	}
+
+	c.access = rep.AccessToken
+	c.refresh = rep.RefreshToken
+
+	return c.access, c.refresh, nil
+}
```
bbengfort, line 143: This is very cool - but probably not necessary; it breaks the paradigm that this is a self-contained credentialer (e.g. it doesn't actually login with the admin) and simply creating a new local access token is the equivalent to refresh without a network call.

pkg/bff/credentials.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/trisacrypto/directory/pkg/gds/admin/v2"
+	"github.com/trisacrypto/directory/pkg/gds/tokens"
+)
+
+// Credentials implements the admin.Credentials interface to provide access and refresh
+// tokens to the GDSClient.
+type Credentials struct {
+	tm      *tokens.TokenManager
+	access  string
+	refresh string
```
bbengfort, line 20: Given that the credentials have the token manager, I don't think a refresh token is necessary. Refresh tokens reduce the chance that you have to resubmit your username/password if you're already authenticated. But they just hand out a new access token which this struct could just as easily generate itself.

bbengfort (commented):
Had a couple of comments @pdeziel -- great start!

pkg/bff/config/config.go
pdeziel: Sure, I can do that.

cmd/gds/main.go
pdeziel: It's embedding related, in the unified `GlobalDirectoryClient` interface we can't have methods with the same name.

pkg/bff/client.go
pdeziel: I think I understand what you mean by embedding the `APIv2` struct into the `GDSClient`, but the fields in `APIv2` are not exported so it doesn't look like I can set them?

cmd/gds/main.go
bbengfort: Ok, since we're going to make the admin client separate from the GDS client, could we please change this back?

pkg/bff/credentials.go
```diff
@@ -0,0 +1,150 @@
...
+
+	jwt "github.com/golang-jwt/jwt/v4"
+	"github.com/trisacrypto/directory/pkg/bff/config"
+	"github.com/trisacrypto/directory/pkg/gds/admin/v2"
+	"github.com/trisacrypto/directory/pkg/gds/tokens"
+)
+
+// Credentials implements the admin.Credentials interface to provide access and refresh
+// tokens to the GDSClient.
+type Credentials struct {
```
bbengfort, line 17: One thing that I didn't understand off the bat was that this implements the `admin.Credentials` interface and that's what ensures that the client can renew their credentials via the [`checkAuthentication`](https://github.com/trisacrypto/directory/blob/main/pkg/gds/admin/v2/client.go#L767) method. Could you please document this in the doc string and also add a:

```
// Ensure Credentials implements the admin.Credentials interface
var _ admin.Credentials = &Credentials{}
```

pkg/bff/credentials.go
bbengfort, line 25: Could we also change this so that it returns an `admin.Credentials`? This is just another helpful indication that this thing is implementing the credentials interface.

bbengfort (approved):

PR #613, opened 2022-06-30
BFF Lookup Endpoint Refactor

bbengfort:
### Scope of changes

Modifies the LookupReply to an object `{"testnet": {}, "mainnet": {}}` rather than an array of results `{"results": []}` 

I also added a `containers/local.sh` script to make it easier to run docker compose. Usage is as follows:

Run `docker system prune --all`:

```
$ ./containers/local.sh clean
```

Build the containers:

```
$ ./containers/local.sh build
```

Run the containers:

```
$ ./containers/local.sh up
```

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [x] other (refactoring)

### Acceptance criteria

@masskoder will this enable you to do what you need to do?

@pdeziel can you think of any other places this change might have an effect?

I did try to run docker compose locally to get you a snapshot of what the data might look like, but I also need #611 to be merged to run docker compose. 

### Author checklist

- [ ]  ~I have updated the dependencies list~
- [ ]  ~I have recompiled and included new protocol buffers to reflect changes I made~

### Reviewer(s) checklist

- [ ] This will allow the front-end to fix SC-6848
- [ ] No other changes in the code are necessary

### DIFF
--- a/containers/local.sh
+++ b/containers/local.sh
@@ -1,14 +1,53 @@
 #!/bin/bash
+# A helper for common docker compose operations to run GDS locally.
 
 # Print usage and exit
-help() {
-    echo "Usage $DIR/local.sh [clean|build|up]"
-    exit 2
+show_help() {
+cat << EOF
+Usage: ${0##*/} [-h] [-p PROFILE] [clean|build|up]
+A helper for common docker compose operations to run GDS
+services locally. Flags are as follows (getopt required):
+
+    -h          display this help and exit
+    -p PROFILE  specify the docker compose profile to use
+
+There are two ways to use this script. Run docker compose:
+
+    ${0##*/} up
+
+Build the images cleaning the docker cache first:
+
+    ${0##*/} clean
+    ${0##*/} build
+
+You can also specify the profile to run only some services.
+EOF
 }
+
+# Parse command line options with getopt
+PROFILE="all"
+OPTIND=1
+
+while getopts hp: opt; do
+    case $opt in
+        h)
+            show_help
+            exit 0
+            ;;
+        p)  PROFILE=$OPTARG
+            ;;
+        *)
+            show_help >&2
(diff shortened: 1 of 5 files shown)


pkg/bff/api/v1/api.go
```diff
@@ -118,7 +118,8 @@ type LookupParams struct {
 // also contain an "error" field - the BFF will handle this field by logging the error
 // but will exclude it from any results returned.
 type LookupReply struct {
-	Results []map[string]interface{} `json:"results"`
+	TestNet map[string]interface{} `json:"testnet"`
+	MainNet map[string]interface{} `json:"mainnet"`
 }
```
pdeziel, line 123: Cool, this makes sense to me since it matches what I've been doing for the other BFF endpoints.

containers/local.sh
```diff
@@ -0,0 +1,37 @@
...
+        exit 0
+    elif [[ $1 == "up" ]]; then
+        echo "starting docker server"
+    else
+        help
+    fi
+fi
+
+# By default just bring docker compose up
+docker compose -p gds -f $DIR/docker-compose.yaml --profile=all up
```
pdeziel, line 37: If it's not too much work, can we get a profile parameter here?

pdeziel (approved):
Looks good! I think this matches better with what we are returning for the `overview` endpoint. I will try to make those changes to #611 this morning so we have the ability to verify these endpoints in docker compose.

containers/local.sh
bbengfort: Yep, will do!

bbengfort: Ok, so now you can do `./containers/local.sh -p api up` and use the `-p` flag to specify the profile.

pdeziel: Awesome!

PR #620, opened 2022-07-01
Save Registration Form

bbengfort:
### Scope of changes

Endpoint 1: `POST /v1/register`: submit form data to be saved on the server instead of local storage. This will mostly be the same as the current form, except it will be expanded to include both testnet/mainnet endpoint/common name (e.g. it will look like the form data, so that it won't have to be edited into a single directory form).

- This endpoint will return 204: No Content if successful.
- This endpoint will perform only limited data validation: it will accept the required keys and parse required types (e.g. bools, numbers) but will not perform any other validation (required field, mismatched data types, etc).
- Requires user with `update:vasp` permission and `orgid` in claims

Endpoint 2: `POST /v1/testnet` or `POST /v1/mainnet`: will cause the form to be submitted to the directory service. No data is required in this POST because the data will be fetched from the saved form -- this means that the form MUST be saved before this endpoint is called.

- This endpoint will return 200: Ok with the response from the directory (e.g. the PKCS12 password) if successful.
- This endpoint will perform some basic validation on required fields before submitting to the directory service. The directory service may also return invalid argument.
- Requires user with `update:vasp` permission and `orgid` in claims

Endpoint 3: `GET /v1/register`: will return the form data.

- Requires user with `read` permission and `orgid` in claims
- Expectation: is called once when the register page loads.

This PR also updates the BFF client with credentials and client options which will hopefully make testing a lot easier. 

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist

- [ ] @masskoder check to ensure that the data exchange will work for the front end and there are no accidental discrepancies
- [ ] @pdeziel please review the `TestLoadRegisterForm` tests as a template for SC-6363; what else do we need to do?

### TODO

Still TODO, but please start the review since these are just tests. If we don't get to them in this PR, I'll create a new story for them.

- [x] Finish `SaveRegisterForm` tests 
- [x] Finish `SubmitRegistration` tests


pkg/bff/api/v1/credentials.go
```diff
@@ -0,0 +1,120 @@
...
+
+// Credentials provides a basic interface for loading an access token from Auth0 into
+// the BFF API Client. Credentials can be loaded from disk, generated, or directly
+// authenticated with Auth0 via a machine-to-machine token.
+//
+// NOTE: this is a fundamentally different mechanism than the GDS Admin API credentials
+// because the Admin API generates its own tokens, and Auth0 manages the BFF credentials.
+type Credentials interface {
+	AccessToken() (string, error)
+}
```
bbengfort, line 17: @pdeziel this is one way I'm hoping to unblock SC-6366; by allowing us to directly use the BFF client for testing and injecting the access token either as a string or loaded from fixtures.

pkg/bff/api/v1/client_test.go
```diff
@@ -172,36 +170,143 @@ func TestRegister(t *testing.T) {
...
+
+	// Execute the Login request
+	err = bff.Login(context.TODO())
+	require.NoError(t, err, "could not login using the bff client")
+
+	// Check to ensure double cookies are set. This doesn't test our code, but ensures
+	// that tests that depend on double cookies will work in the future.
+	u, err := url.Parse(ts.URL)
+	require.NoError(t, err, "could not parse test server url")
+	require.Len(t, client.Jar.Cookies(u), 2, "expected two cookies set in the cookie jar")
```
bbengfort, line 237: @pdeziel verified that `Login` works to set double cookies as long as we use `httptest.NewTLSServer` … the question is how to do this in the bff tests …

pkg/bff/auth/auth.go
```diff
@@ -89,14 +90,23 @@ func NewClaims() validator.CustomClaims {
...
 
+	// If we're in testing mode and no other options have been provided, connect to the
+	// default authtest server to validate local, test credentials
+	if conf.Testing && len(options) == 0 {
+		var ts *authtest.Server
+		if ts, err = authtest.Serve(); err != nil {
+			return nil, err
+		}
+		options = append(options, WithHTTPClient(ts.Client()))
+	}
```
bbengfort, line 108: The idea is if the configuration is in testing mode and no other client is passed to this middleware, then the singleton server in `authtest` will be used. This will only work for testing the `bff` package, we can't use `authtest` elsewhere with he current singleton implementation.

pkg/bff/auth/authtest/server.go
```diff
@@ -34,6 +44,38 @@ const (
...
+// no longer be created a second time because of the use of sync.Once.
+func Close() {
+	srvClose.Do(func() {
+		if srv != nil {
+			srv.Close()
+		}
+		srv = nil
+		srvErr = errors.New("the authtest server has been closed")
+	})
+}
```
bbengfort, line 77: Singleton module structure prevents us from having to figure out how to inject the authtest server client into the middleware.

pkg/bff/db/models/v1/json.go
```diff
@@ -0,0 +1,30 @@
...
+
+// MarshalJSON uses protojson with default marshaling options.
+func (r *RegistrationForm) MarshalJSON() ([]byte, error) {
+	return pbencoder.Marshal(r)
+}
+
+// UnmarshalJSON uses protojson with default unmarshaling options.
+func (r *RegistrationForm) UnmarshalJSON(data []byte) error {
+	return pbdecoder.Unmarshal(data, r)
+}
```
bbengfort, line 30: Ensures the "rewiring" and "unwiring" of json to protocol buffers and back works as expected.

pkg/bff/gds_test.go
```diff
@@ -62,70 +65,85 @@ func (s *bffTestSuite) TestLookup() {
 	require.NotEmpty(rep.TestNet, "expected testnet result from server")
 }
 
-func (s *bffTestSuite) TestRegister() {
+func (s *bffTestSuite) TestLoadRegisterForm() {
```
bbengfort, line 68: Hopefully this is a good template for doing tests against our API endpoints with the BFF client.

pkg/bff/api/v1/client.go
```diff
@@ -298,3 +381,49 @@ func (s *APIv1) Do(req *http.Request, data interface{}, checkStatus bool) (rep *
...
+	} else {
+		cookies = c.client.Jar.Cookies(u)
+		for _, cookie := range cookies {
+			cookie.MaxAge = -1
+		}
+	}
+
+	c.client.Jar.SetCookies(u, cookies)
+	return nil
+}
```
bbengfort, line 429: @pdeziel to use the login endpoint we'd have to use a TLS server - but that would require a bit of refactoring of the tests, so I just added this to the client to set CSRF protection manually.

pkg/bff/gds_test.go
```diff
@@ -62,70 +65,173 @@ func (s *bffTestSuite) TestLookup() {
...
 
+	// Create initial claims fixture
+	claims := &authtest.Claims{
+		Email:       "leopold.wentzel@gmail.com",
+		Permissions: []string{"read:nothing"},
+	}
+
+	// Endpoint must be authenticated
+	_, err := s.client.LoadRegistrationForm(context.TODO())
+	require.EqualError(err, "[401] this endpoint requires authentication", "expected error when user is not authenticated")
```
pdeziel, line 79: Would be nice if we could separate the status check from the error message check, or have a helper to construct and check the expected response, but I think this is fine for right now.

pdeziel (commented):
I've completed a first pass of this and had some comments and a potential bug discovery but overall looks good so far. Thanks for figuring out how to integrate the auth structures into the mock and adding the helpers for the tests!

pkg/bff/api/v1/options.go
```diff
@@ -0,0 +1,20 @@
...
+		return nil
+	}
+}
+
+func WithClient(client *http.Client) ClientOption {
+	return func(c *APIv1) error {
+		c.client = client
+		return nil
+	}
+}
```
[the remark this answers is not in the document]
bbengfort: That's true - we've definitely shied away from mocking our APIs; preferring to use `httptest.Server` and bufconn; feels like we're circling around a formal pattern, but I don't have a name for it quite yet. 

I think this style could be a good idea for the admin client; perhaps we should unify this into a `utils` package so both the admin and BFF clients have similar behavior?

pkg/bff/db/models/v1/organization.go
```diff
@@ -31,3 +31,31 @@ func ParseOrgID(orgID interface{}) (uuid.UUID, error) {
...
+			return false
+		}
+	case "mainnet":
+		if r.Mainnet == nil {
+			return false
+		}
+	default:
+		if r.Testnet == nil || r.Mainnet == nil {
+			return false
+		}
```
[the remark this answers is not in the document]
bbengfort: Happy too - I've added it to #623 to help prevent merge conflicts.

pkg/bff/gds.go
```diff
@@ -124,93 +123,206 @@ func (s *Server) Lookup(c *gin.Context) {
...
 	}
 
-	// Prevent panics by requiring data that will be unwired
-	switch {
-	case in.BusinessCategory == "":
-		c.JSON(http.StatusBadRequest, api.ErrorResponse("business category is required"))
+	// Ensure the registered_directory is one we understand
+	params.Directory = strings.ToLower(params.Directory)
+	if params.Directory != trisatest && params.Directory != vaspdirectory {
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("unknown registered directory"))
```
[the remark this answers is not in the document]
bbengfort: Great catch thank you! I've updated this in #623 to help prevent merge conflicts.

pkg/bff/gds_test.go
bbengfort: I agree - though @DanielSollis did make a good case that testing the string is technically testing both since the status is in the string. Still, it feels more explicit to have an int check and then a message check because then the message check can be optional (e.g. if it's a custom message based on the input).

pdeziel (approved):
:shipit:

PR #632, opened 2022-07-12
Add Auth0 AppMetdata Struct

bbengfort:
### Scope of changes

This PR should fix the panic caused when logging in a user that has a `nil` `app_metadata` and ensure that the record is correctly populated. It introduces a struct `AppMetadata` that will allow us to serialize the `app_metadata` JSON in a structured fashion, ensuring keys and parsing are correct.

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

This PR could really use an intense sanity check; particularly second eyes on the user-login logic. Happy to discuss if that would make things easier. I still need to test that this fixes the panic.

### Author checklist

### Reviewer(s) checklist

- [ ] Login algorithm makes sense
- [ ] It makes sense how we might extend and add to the `app_metadata`
- [ ] The claims using the same struct makes sense 
- [ ] Did I miss any test cases?


pkg/bff/users.go
```diff
@@ -78,39 +85,29 @@ func (s *Server) Login(c *gin.Context) {
...
-			if err = s.auth0.User.Update(*user.ID, user); err != nil {
-				log.Error().Err(err).Msg("could not update user app_metadata")
-				c.JSON(http.StatusInternalServerError, "could not complete user login")
-				return
-			}
-		}
+	if err = s.SaveAuth0AppMetadata(*user.ID, *appdata); err != nil {
+		log.Error().Err(err).Str("user_id", *user.ID).Msg("could not save user app_metadata")
+		c.JSON(http.StatusInternalServerError, "could not complete user login")
+		return
```
pdeziel, line 110: Since we are now syncing the user app metadata at the end, this protects us from partial updates?

pdeziel (approved):
This makes sense to me, I like the fact that we have a variety of test cases for marshaling and unmarshaling the auth0 response. I guess we will have to remember to update those if we add fields to auth0?

pkg/bff/users.go
bbengfort: There should only be two possible cases:

1. The user is logging in for the first time and an organization is created for them 
2. The user needs the VASP ids updated from their organization

In case #1 - the following data is saved:

```json
{
  "orgid": "uuid",
  "vasps": {
    "testnet": "",
    "mainnet": "",
  }
}
```

Case #1 should only happen once. 

Because I moved the syncing to the end, case #2 is going to happen on every single login since I removed the `MapEqual` check. I debated whether or not we should do this -- it is safer, and logins are infrequent. However, we will eventually need some change detection to alert the front-end that the user needs to login again; so this is likely temporary.

pkg/bff/admin.go
```diff
@@ -83,8 +83,8 @@ func (s *Server) Certificates(c *gin.Context) {
 	// Note that if testnet or mainnet are absent from the VASPs map, the ID will
 	// default to an empty string, and GetCertificates will return nil for that network
 	// instead of an error.
-	testnetID := claims.VASPs[testnet]
-	mainnetID := claims.VASPs[mainnet]
+	testnetID := claims.VASPs.TestNet
+	mainnetID := claims.VASPs.MainNet
```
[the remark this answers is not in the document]
bbengfort: Done!

bbengfort:
> This makes sense to me, I like the fact that we have a variety of test cases for marshaling and unmarshaling the auth0 response. I guess we will have to remember to update those if we add fields to auth0?

Agreed, if we add more fields we should add more cases; though I do think I went a little overboard with all the different combinations -- we don't have to test every single combination of empty and filled field!

PR #644, opened 2022-07-19
certificate migration

pdeziel:
### Scope of changes

SC-6543 This branch contains a new trtl cli command to perform the certificate record migration and should not be merged into main.

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [x] database migration

### Acceptance criteria

We should make sure the logic makes sense and it works locally before doing the migration.

### Author checklist

### Reviewer(s) checklist

- [ ] Does the migration logic make sense?
- [ ] Do we need any additional debug info to give us more confidence?


cmd/trtl/main.go
```diff
@@ -530,6 +546,195 @@ func initPeersClient(c *cli.Context) (err error) {
...
+		}
+
+		// Get all the certificate requests for this vasp
+		var ids []string
+		if ids, err = models.GetCertReqIDs(vasp); err != nil {
+			migrateErr = err
+			break
+		}
+
+		// Get the latest completed certificate request
```
pdeziel, line 597: I guess since we overwrite the identity certificate, we won't be able to capture the old certificates from the older certificate requests.

cmd/trtl/main.go
```diff
@@ -530,6 +546,195 @@ func initPeersClient(c *cli.Context) (err error) {
...
+	req := &pb.CursorRequest{
+		Namespace: wire.NamespaceVASPs,
+	}
+	var stream pb.Trtl_CursorClient
+	if stream, err = dbClient.Cursor(ctx, req); err != nil {
+		return cli.Exit(err, 1)
+	}
+
+	var migrated, skipped int
+	var migrateErr error
```
bbengfort, line 573: I think it might be better to use a `multierror` here - it feels like there are a couple of places where we might want to see multiple nested errors. It will also help protect us from the break logic.

cmd/trtl/main.go
bbengfort: Yeh, that's unfortunate -- there are only a handful through, so I could manually create those entries if we do this migration soon - mind creating a follow on story for that?

cmd/trtl/main.go
```diff
@@ -530,6 +546,195 @@ func initPeersClient(c *cli.Context) (err error) {
...
+
+		// Get the latest completed certificate request
+		var certreq *models.CertificateRequest
+		for _, id := range ids {
+			var rep *pb.GetReply
+			if rep, err = dbClient.Get(ctx, &pb.GetRequest{
+				Namespace: wire.NamespaceCertReqs,
+				Key:       []byte(id),
+			}); err != nil {
+				fmt.Printf("WARNING: could not get certreq %s for vasp %s (%s): %s\n", id, vasp.Id, vasp.CommonName, err)
```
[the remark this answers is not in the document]
bbengfort: Yeh, that's not the best - perhaps we could add a flag to clean these old certificate requests up?

cmd/trtl/main.go
```diff
@@ -530,6 +546,195 @@ func initPeersClient(c *cli.Context) (err error) {
...
+	}
+	var stream pb.Trtl_CursorClient
+	if stream, err = dbClient.Cursor(ctx, req); err != nil {
+		return cli.Exit(err, 1)
+	}
+
+	var migrated, skipped int
+	var migrateErr error
+
+	for {
```
bbengfort, line 575: I'd recommend putting loop labels on the outer and inner nested loops and then break explicitly.

cmd/trtl/main.go
```diff
@@ -530,6 +546,195 @@ func initPeersClient(c *cli.Context) (err error) {
...
+
+			current := &models.CertificateRequest{}
+			if err = proto.Unmarshal(rep.Value, current); err != nil {
+				migrateErr = err
+				break
+			}
+
+			if current.Status == models.CertificateRequestState_COMPLETED {
+				certreq = current
+			}
```
bbengfort, line 620: so I'm not sure this logic is 100% correct - a vasp might have multiple completed certificate requests; this does not imply that it is the latest request without timestamp checks. 

Multiple completed certreqs will often happen on reissuance. 

Also - do we want to create a certificate for all completed cert reqs and then mark the ones that need to be updated with certificate info? I know that might be a manual process.

cmd/trtl/main.go
```diff
@@ -530,6 +546,195 @@ func initPeersClient(c *cli.Context) (err error) {
...
+		}
+
+		if certreq == nil {
+			fmt.Printf("skipping vasp %s (%s) with no certificates\n", vasp.Id, vasp.CommonName)
+			if report != nil {
+				report.WriteString(fmt.Sprintf("skipping vasp %s (%s) with no certificates\n", vasp.Id, vasp.CommonName))
+			}
+			skipped++
+			continue
+		}
```
bbengfort, line 634: might also need a `vasp.IdentityCertificate == nil` check here.

bbengfort (commented):
Looks like a good start - we'll definitely need to try it locally on the live data/staging data to make sure it works as we expect! 

I've marked this PR as comment rather than a :shipit: since we're not planning on merging this PR.

cmd/trtl/main.go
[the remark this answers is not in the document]
pdeziel: Sure but it will take some effort since we don't have a convenient models helper for that currently.

cmd/trtl/main.go
pdeziel: So should we check the `Created` timestamp to determine the newest completed certificate to put the identity certificate in and create temporary records for the completed ones? It's tricky because we don't know the serial number to create the ID with but maybe we can populate those with a random UUID and when we do the manual cleanup, we already know which ones need to be corrected.

cmd/trtl/main.go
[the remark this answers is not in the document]
bbengfort: Understood - it's probably worth it to remove dangling references.

cmd/trtl/main.go
bbengfort: The Created timestamp methodology works for me. 

Instead of creating temporary records, would it be easier to mark them in the report (e.g. via certificate request id), and then we can create the records manually later?

kbelita (commented):
lgtm 👍

cmd/trtl/main.go
```diff
@@ -530,6 +583,607 @@ func initPeersClient(c *cli.Context) (err error) {
...
+		// Get the certificate request from the batch ID
+		var batchId int64
+		if batchId, err = strconv.ParseInt(record[rows["Batch Id"]], 10, 64); err != nil {
+			results = multierror.Append(results, fmt.Errorf("record %d has invalid batch id: %s", i, err))
+			continue recordLoop
+		}
+		var certReq *models.CertificateRequest
+		if certReq, ok = certReqs[batchId]; !ok {
+			cert.Status = models.CertificateState_REVOKED
+			results = multierror.Append(results, fmt.Errorf("WARNING: could not find certificate request with batch id %d", batchId))
```
pdeziel, line 1073: Not sure what's going on here, I was hoping to use the batch ID to lookup the corresponding certificate request but it seems that for some of the certificates (a few alice and bob ones, as well as a traveler staging cert) to be restored, there is no matching certificate request for the batch ID. Have they been deleted? If so should we assume they are revoked?

bbengfort:
@pdeziel should we close this PR now that the migration is complete?

PR #653, opened 2022-07-22
certmanager check vasp state

pdeziel:
### Scope of changes

This adds some checks to the certmanager routines to make sure that VASPs are in the correct states for certificate issuance. This helps ensure that VASPs are not issued certificates if they have errored or been rejected.

SC-6508

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Thought I should open this PR for review, although probably needs a few tests to verify.

### Author checklist

### Reviewer(s) checklist


pkg/gds/certs.go
```diff
@@ -85,7 +85,25 @@ func (s *Service) HandleCertifcateRequests(certDir string) (err error) {
...
-				if err := s.submitCertificateRequest(req); err != nil {
+				// Get the VASP from the certificate request
+				var vasp *pb.VASP
+				if vasp, err = s.db.RetrieveVASP(req.Vasp); err != nil {
+					logctx.Error().Str("vasp_id", req.Vasp).Msg("could not retrieve vasp for certificate request submission")
+					return
+				}
+
+				// Verify that the VASP has not errored or been rejected
+				if vasp.VerificationStatus != pb.VerificationState_REVIEWED {
```
bbengfort, line 96: I think this needs to be `vasp.VerificationStatus < pb.VerificationState_REVIEWED  || vasp.VerificationStatus > pb.VerificationState_VERIFIED`  -- because when we reissue certificates the VASP will be in the verified state. The `>` and `<` operators will also include `pb.VerificationState_ ISSUING_CERTIFICATE `  but that seems ok to me.

bbengfort (approved):
@pdeziel looks good; I just had one comment on the submit certificate state check. Nice work on passing the `vasp` down to the sub-methods rather than doing two database queries! 

I agree that some tests for this are needed; do you want to create a story for tests this sprint or move this story into this sprint and add testing to it?

bbengfort:
@pdeziel I'm looking at the Go Test CI and it looks like this is not a transient failure of the backup manager, but rather a data race has been introduced in the CertMan code:

https://github.com/trisacrypto/directory/runs/7499533774?check_suite_focus=true#step:9:584

I believe the issue is that you're reading the `req` outside of the go routine but then updating it in the go routine. To fix the issue, I think you're going to have to pass the `req` into the go routine along with the `logctx` otherwise the loop will change the `req` and it will not be correctly updated. 

Tough bug, I'm really glad for that race detector because I wouldn't have seen that!

pdeziel:
> @pdeziel I'm looking at the Go Test CI and it looks like this is not a transient failure of the backup manager, but rather a data race has been introduced in the CertMan code:
> 
> https://github.com/trisacrypto/directory/runs/7499533774?check_suite_focus=true#step:9:584
> 
> I believe the issue is that you're reading the `req` outside of the go routine but then updating it in the go routine. To fix the issue, I think you're going to have to pass the `req` into the go routine along with the `logctx` otherwise the loop will change the `req` and it will not be correctly updated.
> 
> Tough bug, I'm really glad for that race detector because I wouldn't have seen that!

@bbengfort I think we noticed this at the same time, I thought it was due to using the same `err` inside and outside of the go routine, so I've created a local variable for the go routine and it seems to have fixed the issue.

pkg/gds/certs.go
```diff
@@ -85,7 +85,28 @@ func (s *Service) HandleCertifcateRequests(certDir string) (err error) {
...
+				if vasp.VerificationStatus < pb.VerificationState_REVIEWED || vasp.VerificationStatus > pb.VerificationState_VERIFIED {
+					logctx.Info().Str("verification_status", vasp.VerificationStatus.String()).Msg("vasp is not ready for certificate request submission")
+					if err = models.UpdateCertificateRequestStatus(req, models.CertificateRequestState_CR_REJECTED, "certificate request rejected", "automated"); err != nil {
+						logctx.Error().Err(err).Msg("could not update certificate request status")
+						return
+					}
+					if err = s.db.UpdateCertReq(req); err != nil {
+						logctx.Error().Err(err).Msg("could not save updated certificate request")
+						return
+					}
```
bbengfort, line 108: @pdeziel I still think the `req` is a problem; since the for loop is going to modify `req` on line 75 and so the go routine might have the wrong pointer when it's saving the certificate request; I think the reason we haven't been caught by this bug in the past is that we only are usually doing one cert at a time -- but now that we're doing bulk re-issuance; I'm guessing it's going to crop up.

pdeziel: Gotcha, I've updated the PR with that fix.

bbengfort: Thanks!

PR #657, opened 2022-07-25
release trtl iterators

pdeziel:
### Scope of changes

This fixes a bug in the trtl store tests where we were failing to close iterators properly after using them, which manifested in race conditions during test suite teardown which were difficult to track down.

SC-6488

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

The race conditions in the tests should be addressed, unfortunately I was never able to reproduce them locally but so far they haven't shown up in CI yet 🤞 .

### Author checklist

### Reviewer(s) checklist

- [ ] For every spot in the tests where we are using the raw trtl streaming iterator, we are calling `Release()` to close the iterator properly on the client side.


pkg/gds/store/trtl/trtl_test.go
```diff
@@ -223,6 +224,7 @@ func (s *trtlStoreTestSuite) TestDirectoryStore() {
 	require.NoError(err)
```
bbengfort, line 224: On [Line 202](https://github.com/trisacrypto/directory/blob/e0be55fbb8477f63a905d6aa34079fd9f170dbd3/pkg/gds/store/trtl/trtl_test.go#L202) above and [Line 245](https://github.com/trisacrypto/directory/blob/e0be55fbb8477f63a905d6aa34079fd9f170dbd3/pkg/gds/store/trtl/trtl_test.go#L245) below, does calling `All()` release the iterator?

bbengfort (approved):
Thank you for figuring this out - it's definitely been annoying to have CI failing on us and this was definitely a hidden issue; nice one!

pkg/gds/store/trtl/trtl_test.go
pdeziel: Yes, `ListVASPs()` creates the iterator, and `All()` releases the iterator when it returns.

bbengfort: Cool - then this looks good to go to me! Hope it sorts out CI!

PR #658, opened 2022-07-25
remove certreq references on rejected registration

pdeziel:
### Scope of changes

This removes the certreq references on the VASP extra when a certificate request is deleted, which happens when a VASP registration is rejected on `Review`. This also happens on `DeleteVASP`, but since the VASP object is being deleted anyway there's no point doing it on that endpoint.

SC-7517

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

This change is not visible to the user, but I've updated the `TestReviewReject` test to account for this fix.

### Author checklist

### Reviewer(s) checklist

- [ ] The logging/error logic makes sense in `rejectRegistration`.
- [ ] There are no other places where we need to account for deleted certificate requests.

### DIFF
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -2092,8 +2092,17 @@ func (s *Admin) rejectRegistration(vasp *pb.VASP, reason string, claims *tokens.
 			continue
 		}
 
+		// Delete the certificate request
 		if err = s.db.DeleteCertReq(careq.Id); err != nil {
 			log.Error().Err(err).Str("id", careq.Id).Msg("could not delete certificate request")
+			continue
+		}
+
+		// Delete the VASP reference to the certificate request
+		if err = models.DeleteCertReqID(vasp, careq.Id); err != nil {
+			log.Error().Err(err).Str("vasp", vasp.Id).Str("certreq", careq.Id).Msg("could not delete certificate request ID from VASP")
+		} else if err = s.db.UpdateVASP(vasp); err != nil {
+			log.Error().Err(err).Str("vasp", vasp.Id).Msg("could not update VASP with deleted certificate request ID")
 		}
 		ncertreqs++
 	}
--- a/pkg/gds/admin_test.go
+++ b/pkg/gds/admin_test.go
@@ -1682,6 +1682,11 @@ func (s *gdsTestSuite) TestReviewReject() {
 	require.Equal(pb.VerificationState_REJECTED, log[3].CurrentState)
 	require.Equal(request.claims.Email, log[3].Source)
 
+	// Certificate request should be deleted from the VASP extra
+	ids, err := models.GetCertReqIDs(v)
+	require.NoError(err)
+	require.Len(ids, 0)
+
 	// Certificate request should be deleted
 	_, err = s.svc.GetStore().RetrieveCertReq(xrayID)
 	require.Error(err)
--- a/pkg/gds/models/v1/models.go
+++ b/pkg/gds/models/v1/models.go
@@ -175,6 +175,39 @@ func AppendCertReqID(vasp *pb.VASP, certreqID string) (err error) {
 	return nil
 }
 
+// DeleteCertReqID removes the certificate request ID from the VASP if it exists.
+func DeleteCertReqID(vasp *pb.VASP, certreqID string) (err error) {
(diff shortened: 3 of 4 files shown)


pkg/gds/admin.go
```diff
@@ -2092,8 +2092,17 @@ func (s *Admin) rejectRegistration(vasp *pb.VASP, reason string, claims *tokens.
...
 			log.Error().Err(err).Str("id", careq.Id).Msg("could not delete certificate request")
+			continue
+		}
+
+		// Delete the VASP reference to the certificate request
+		if err = models.DeleteCertReqID(vasp, careq.Id); err != nil {
+			log.Error().Err(err).Str("vasp", vasp.Id).Str("certreq", careq.Id).Msg("could not delete certificate request ID from VASP")
+		} else if err = s.db.UpdateVASP(vasp); err != nil {
+			log.Error().Err(err).Str("vasp", vasp.Id).Msg("could not update VASP with deleted certificate request ID")
 		}
```
pdeziel, line 2106: Added the continue here so the cert request count is accurate and the reference on the VASP is in sync with the certificate request if deleting the certificate request fails.

pkg/gds/admin.go
```diff
@@ -2092,8 +2092,17 @@ func (s *Admin) rejectRegistration(vasp *pb.VASP, reason string, claims *tokens.
 			continue
 		}
 
+		// Delete the certificate request
 		if err = s.db.DeleteCertReq(careq.Id); err != nil {
 			log.Error().Err(err).Str("id", careq.Id).Msg("could not delete certificate request")
+			continue
+		}
```
bbengfort, line 2099: I think we should take a minute to challenge our design and our assumptions. Is deleting the certificate request the right thing to do or should we mark it as `CR_REJECTED`? 

On the one hand, deleting the certificate request means there are fewer certificate requests for CertMan to iterate over (though we could fix that by tombstoning or archiving into a different namespace). On the other hand, if we want to "unreject" this VASP, we'll have to create a new certificate request.

To me, deleting the certificate request makes sense, but I did just want to take a minute to review our choices. 

Any thoughts one way or the other?

pkg/gds/admin.go
bbengfort: Good idea - well thought out!

pkg/gds/admin_test.go
```diff
@@ -1682,6 +1682,11 @@ func (s *gdsTestSuite) TestReviewReject() {
 	require.Equal(pb.VerificationState_REJECTED, log[3].CurrentState)
 	require.Equal(request.claims.Email, log[3].Source)
 
+	// Certificate request should be deleted from the VASP extra
+	ids, err := models.GetCertReqIDs(v)
+	require.NoError(err)
+	require.Len(ids, 0)
+
```
bbengfort, line 1689: Good additional test!

pkg/gds/models/v1/models.go
```diff
@@ -175,6 +175,39 @@ func AppendCertReqID(vasp *pb.VASP, certreqID string) (err error) {
...
+	}
+
+	// Search the slice for the certificate request ID
+	for i, containsID := range extra.CertificateRequests {
+		if certreqID == containsID {
+			// Remove the certificate request ID from the array
+			extra.CertificateRequests = append(extra.CertificateRequests[:i], extra.CertificateRequests[i+1:]...)
+			break
+		}
+	}
```
bbengfort, line 202: If the list of CertificateRequests on VASPs gets longer, we may want to consider refactoring to an ordered list of UUIDs so that we can use `sort.Search` for better retrieval. Just something to keep in mind for the future, no action required now.

bbengfort (approved):
Thank you for fixing this bug @pdeziel!

pkg/gds/admin.go
pdeziel: In general I think I agree with deleting the certificate requests. In my mind they are more of a method of interacting with the certificate provider and once they have served their purpose there's not much use having them around besides debug purposes.

Perhaps to allow a path to "unrejection", we could mark as `CR_REJECTED` here and have the cert manager be responsible for deleting old certificate requests.

PR #670, opened 2022-07-27
BFF: Add form state to registration form

pdeziel:
### Scope of changes

This adds a form state to the registration form so that it is loaded and stored along with the form on the bff register endpoints.

SC-7605

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

We need to verify that this structure is sufficient for storing and retrieving the stepper state so that the frontend can rely on it for providing a consistent form status view across different devices.

### Author checklist

### Reviewer(s) checklist

- [ ] The form state fields make sense and are sufficient for storing the form progress state.
- [ ] Does this design assume too much about the frontend? What happens if the stepper layout changes on the frontend and there's an old form in the database?
- [ ] Are we okay with all users in the organization being "forced" to the same step in the form?


bbengfort (approved):
Good solution; I like the generic sequence of steps that doesn't directly couple the form state to the form itself.

PR #679, opened 2022-08-01
Certman refactor

pdeziel:
### Scope of changes

There are two main parts to this PR.

1. Refactor certman into its own package, allowing it to be tested completely independently of the GDS service.
2. Refactor the fixture code into its own package, this helps with the independent testing of certman and also cleans up a lot of the code in `service_test.go`.

SC-6621

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [x] refactor

### Acceptance criteria

Test should still be passing.

### Author checklist

### Reviewer(s) checklist

- [ ] Are we okay with the various arguments that have to be passed into the `New` method for certman?
- [ ] The fixtures refactoring makes sense and it's clear how to use the package for writing new tests.


pkg/gds/admin_test.go
```diff
@@ -410,7 +411,7 @@ func (s *gdsTestSuite) TestSummary() {
 		VASPsCount:           14,
 		PendingRegistrations: 6,
 		ContactsCount:        39,
-		VerifiedContacts:     26,
+		VerifiedContacts:     24,
```
rebeccabilbro, line 414: I think this is the only remaining mystery?

rebeccabilbro (approved):
Thank you @pdeziel for the live walkthrough and for creating some follow-on stories based on the conversation!

pkg/gds/admin_test.go
pdeziel: I think I've cracked this, there are some tests which modify the fixtures in the reference map. To get around this we usually defer `LoadReferenceFixtures` to reload the references for the following tests. This was missing in `RetrieveVASP` because `CompareFixture` nils out some of the VASP fields before comparison. We had also previously put in some deep copy code into `CompareFixture` to avoid modifying the values in the reference fixtures. However, I think the more user friendly option is just to call `LoadReferenceFixtures` in the 3-4 tests that need to modify the reference objects so I've removed the deep copy code as it was a bit hard to follow.

kbelita (commented):
thank you for the walkthrough @pdeziel!

kbelita (approved):

PR #681, opened 2022-08-02
Add started timestamp to organization form

pdeziel:
### Scope of changes

This adds a "started" timestamp to the organization form. The point of this is to be able to determine from the backend if the user has started filling out a form (in other words, they have saved it at least once from the UI). Once use case of this is the `Attention` endpoint which can check this field to determine if it should return the "start your registration" message.

SC-7760

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

When a user saves an organization form, the `Attention` endpoint will no longer return the "start registration" message.

### Author checklist

### Reviewer(s) checklist

- [ ] The assumption is that a user has not started the form until they hit "save" and the form successfully goes to the backend. Is this the right semantic?
- [ ] When we save the form in the backend, what should we do if the form is empty? Currently we interpret this as a "delete" request and the started timestamp does not get populated.


pkg/bff/gds.go
```diff
@@ -250,6 +250,12 @@ func (s *Server) SaveRegisterForm(c *gin.Context) {
 		return
 	}
 
+	// Mark the form as started
+	// NOTE: If an empty form was passed in, the form will not be marked as started.
+	if form.State != nil && form.State.Started == "" {
+		form.State.Started = time.Now().Format(time.RFC3339)
+	}
+
```
pdeziel, line 258: If an empty form is passed in we treat it as a delete since it completely overwrites the existing value in the database. So, I think the right thing to do is not mark the form as started in that case, which is what this code is doing.

rebeccabilbro (approved):
LGTM @pdeziel -- I know this will be useful to the front end team!

PR #688, opened 2022-08-04
Add registration status endpoint

pdeziel:
### Scope of changes

This adds an endpoint for the frontend which returns the registration status for an organization in the form of submitted timestamps. At this point it's just a data pass-through to fields on the organization object to allow the frontend to determine whether or not to show the registration buttons to the user.

SC-7903

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Endpoint should return submitted timestamps for testnet and mainnnet.

### Author checklist

### Reviewer(s) checklist

- [ ] Do we need any other metadata in the returned payload?


pkg/bff/admin.go
```diff
@@ -354,3 +354,27 @@ func (s *Server) Attention(c *gin.Context) {
...
+
+	// Build the response
+	// TODO: We should be querying the VASP record instead to allow for re-registration
+	out := &api.RegistrationStatus{}
+	if org.Testnet != nil && org.Testnet.Submitted != "" {
+		out.TestNetSubmitted = org.Testnet.Submitted
+	}
+	if org.Mainnet != nil && org.Mainnet.Submitted != "" {
+		out.MainNetSubmitted = org.Mainnet.Submitted
+	}
```
pdeziel, line 378: Currently it doesn't look like the frontend allows a path to re-registration, because the bff doesn't know which state the VASP is in. In order to make this possible, I think we need to check the status of the VASP in GDS rather than using the submitted field on the organization, since those fields don't get reset. In order to do that, we really need to have a bff-specific API for the bff to communicate with the Admin API, which is a larger effort.

rebeccabilbro: Hmm, that's a good point -- does that mean re-registration currently means wiping the state stored by the front end and starting over?

rebeccabilbro (approved):
LGTM @pdeziel -- I think we are OK to merge this (I know there's a lot going on with the upcoming deployment, but I'm assuming this is needed for the release?)

pkg/bff/admin.go
pdeziel: Yes I think we would have to update the organization record in the database. This PR just allows the frontend to know if the form has already been submitted before allowing the user to submit a registration (they currently get an error when clicking the submit button).

PR #692, opened 2022-08-05
function to check if an email has been sent to the contact type sc-7480 

kbelita:
### Scope of changes

Created a function to check if an email has been sent to the contact in some time window) (e.g. the past 30 days) and added  it to pkg/gds/emails. 

Function checks by contact type and not by VASP. This is to make address cases where different contacts have different verifications status.

[sc-7480](https://app.shortcut.com/rotational/story/7480/sent-email-log)

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/gds/emails/sentlogs.go
```diff
@@ -0,0 +1,37 @@
+// This contains helper functions to help confirm if emails were sent.
+package emails
```
pdeziel, line 2: This actually looks like a good fit for the `models` package, may I suggest moving this into the `pkg/gds/models/v1/contacts.go` since it's a direct check against the `Contacts` object? Also, can we pass in the Contacts directly into the method instead of the VASP?

pkg/gds/emails/sentlogs.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"time"
+
+	"github.com/trisacrypto/directory/pkg/gds/models/v1"
+	pb "github.com/trisacrypto/trisa/pkg/trisa/gds/models/v1beta1"
+)
+
+// VerifyEmailSentVASPContact checks if an email has been sent to the VASP contacts in some time window (e.g. the past 30 days).
+// Returns false for either email has not been sent or an error occurred.
+func VerifyEmailSentVASPContact(vasp *pb.VASP, kind string, timeWindowDays int) (status bool, err error) {
```
pdeziel, line 13: Instead of a bool, can we return a count of the sent emails and also require a reason string to be passed in? I realize this is a bit of a refactor but @DanielSollis will need to check how many emails are sent since both the 30 day and 7 days reminders are classified as the same reason. We can rename this method to `GetSentEmailCount` or something.

pkg/gds/emails/sentlogs.go
```diff
@@ -0,0 +1,37 @@
...
+
+	contact := models.ContactFromType(vasp.Contacts, kind)
+	emailLog, err := models.GetEmailLog(contact)
+	if err != nil {
+		status = false
```
pdeziel, line 18: Note: Go assigns default values to declared variables (false for bool), so it's not necessary to set status to false since it's declared in the returned params. This also works for ints (initialized to 0) and slices (initialized to nil).

pdeziel (changes requested):
Good start! I have a few suggestions below to make this method more useful based on our discussion this afternoon.

pkg/gds/emails/sentlogs.go
```diff
@@ -0,0 +1,37 @@
...
+		return status, err
+	} else {
+		// get latest timestamp
+		strTimestamp := emailLog[len(emailLog)-1].Timestamp
+
+		//format emailTimestamp
+		timestamp, _ := time.Parse(time.RFC3339, strTimestamp)
```
pdeziel, line 25: Also, can we handle the error case here?

pkg/gds/models/v1/contacts.go
```diff
@@ -295,3 +296,27 @@ func AppendEmailLog(contact *pb.Contact, reason string, subject string) (err err
 	}
 	return nil
 }
+
+func GetSentEmailCount(contact *pb.Contact, reason string, timeWindowDays int) (sent int, err error) {
```
kbelita, line 300: @pdeziel ready for your review!

pdeziel (approved):
:shipit:

PR #742, opened 2022-08-25
Return registration form on PUT

pdeziel:
### Scope of changes

To avoid the frontend having to call GET register immediately after PUT register to retrieve a saved registration form, this adds the registration form to the response body on PUT. With this change, the frontend should capture the returned form since the `SaveRegistrationForm` handler on the BFF may modify the state in the form "stepper".

SC-8348

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

When the frontend calls the register PUT endpoint the BFF should now return the saved registration form in the response body with a 200 status code. If the frontend puts an empty registration form, the BFF should return 204 No Content instead.

### Author checklist

### Reviewer(s) checklist


rebeccabilbro (approved):
Nice work @pdeziel! Hopefully this allows the front end to reliably check to see if unsubmitted registrations already have some saved data

PR #745, opened 2022-08-26
Sync with auth0 after submitting registration

pdeziel:
### Scope of changes

This adds a syncing step with auth0 on the submit registration endpoint. This will allow the frontend to refresh the access token after the user submits a registration, which will mean that the claims are updated with the VASP IDs without the user having to login again.

In order to do this, I expanded the authtest mock server a bit so it maintains some user app metadata state that tests can assert against.

SC-8518

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

When the user submits a registration, the claims in auth0 should contain the VASP ID for the corresponding registration without the user having to login to the frontend again.

### Author checklist

### Reviewer(s) checklist


pkg/bff/gds.go
```diff
@@ -404,5 +425,12 @@ func (s *Server) SubmitRegistration(c *gin.Context) {
 		return
 	}
 
+	// Commit the user metadata updates to auth0
+	if err = s.SaveAuth0AppMetadata(*user.ID, *appdata); err != nil {
+		log.Error().Err(err).Str("user_id", *user.ID).Msg("could not save user app metadata")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not save user app metadata"))
```
rebeccabilbro, line 431: If this fails, maybe it's okay to just fall back to the current state, which is that the user will have to log out and log back in again (rather than erroring with a 500 here). No matter what, the GDS registration has already been submitted at this point and we don't want them to misunderstand and believe they need to resubmit. We should keep the error log message though so we can introspect!

rebeccabilbro (approved):
@pdeziel thanks for finding a quick solution! I had one note about a 500 error that you should take a look at, and there might be some cleanup from a merge conflict related to some of the other PRs that we merged in yesterday. This should be a quick one for you to wrap up when you're back in the office tomorrow.

Are there any other followon testing stories that we should create that can use the new authtest functionality introduced by this PR?

pkg/bff/gds.go
pdeziel: That's a good point, I'll keep the error to make sure we catch that!

PR #757, opened 2022-09-01
add workflow action for gcr-cleaner dry-run sc-8003

kbelita:
### Scope of changes
Add workflow action for gcr-cleaner dry-run. This current PR is a for dry-run only. It will not delete anything and outputs what would be deleted.
sc-8003

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [x] other (yaml)

### Acceptance criteria
--
### Author checklist

- [X]  I have moved the associated Shortcut story to "Ready for Review"

### Reviewer(s) checklist


.github/workflows/gcr-cleaner.yaml
```diff
@@ -0,0 +1,32 @@
+# .github/workflows/gcr-cleaner.yaml
+name: 'gcr-cleaner'
```
bbengfort, line 2:
```suggestion
name: GCR Cleanup
```

To match our other GitHub Actions names (Containers, Staging Containers, etc.) could we rename the action?

.github/workflows/gcr-cleaner.yaml
```diff
@@ -0,0 +1,32 @@
...
+    branches:
+      - main
+    tags:
+      - 'v*'
+jobs:
+  gcr-cleaner:
+    runs-on: 'ubuntu-latest'
+    steps:
+      # configure based on registry
+      - uses: 'docker/login-action@v2'
```
bbengfort, line 14: We're using `docker/login-action@v1` in our container builds:

https://github.com/trisacrypto/directory/blob/main/.github/workflows/containers.yaml#L55

Do you think we should upgrade our other workflows to use `v2`?

.github/workflows/gcr-cleaner.yaml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
```
bbengfort, line 17: The secret we already have for DockerHub login is `DOCKERHUB_ACCESS_TOKEN` and we've created a GitHub-specific access token so that we can easily revoke it as necessary. Did you add a `DOCKERHUB_PASSWORD` to the Actions secrets? If so could we delete it and use the access token instead? That will make managing and revoking things much easier. 

https://github.com/trisacrypto/directory/blob/main/.github/workflows/containers.yaml#L58

.github/workflows/gcr-cleaner.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+      # customized based on the gcr-cleaner flags
+      # -repo - list of repo to clean (required)
+      # -grace - relative duration in which to ignore references, refs newer than duration will not be deleted
+      # -keep - minimum number of images to keep, will not consider images inside the grace duration
+      # -dry_run - if set to true, will not delete anything and outputs what would have been deleted
+      # -recursive - if set to true, will recursively search all child repositories
```
bbengfort, line 24: Excellent comment! So if I'm reading this correctly we'll always keep 30 days worth of images as well as the 10 most recent images that are older than 30 days?

bbengfort (commented):
I had a couple of minor comments in the review but my biggest question is if the authentication is correct? To login to GCR (which is Google, not Docker) we use the following in our container build workflows:

https://github.com/trisacrypto/directory/blob/main/.github/workflows/containers.yaml#L60-L66

```
      - name: Login to GCR
        uses: docker/login-action@v1
        with:
          registry: gcr.io
          username: _json_key
          password: ${{ secrets.GCR_SERVICE_ACCOUNT }}
```

Wouldn't we have to use this login instead of the DockerHub login for this to work? 

Can we use this tool to clean up our DockerHub repository as well?

.github/workflows/gcr-cleaner.yaml
kbelita: perhaps we can update all the workflows together at another time. I'll create a story.

.github/workflows/gcr-cleaner.yaml
kbelita: will update this. did not actually add the `DOCKERHUB_PASSWORD`

.github/workflows/gcr-cleaner.yaml
kbelita: Correct. Here is the information from the [gcr-cleaner `README.md`](https://raw.githubusercontent.com/GoogleCloudPlatform/gcr-cleaner/main/README.md) as an FYI.

> - `grace` - Relative duration in which to ignore references. This value is
>   specified as a time duration value like "5s" or "3h". If set, refs newer than
>   the duration will not be deleted. If unspecified, the default is no grace
>   period (all untagged image refs are deleted).
> 
> - `keep` - If an integer is provided, it will always keep that minimum number of
>   images. Note that it will not consider images inside the `grace` duration. GCR
>   Cleaner attempts to keep the most recently created images, but there are some
>   caveats. Some community tooling sets container creation time to a date back in
>   1980, which breaks the default sorting algorithm. As such, GCR Cleaner uses
>   the following sorting algorithm for container images:
> 
>     - If either of the containers were created before Docker even existed, it
>       sorts by the date the container was uploaded to the registry.
> 
>     - If two containers were created at the same timestamp, it sorts by the date
>       the container was uploaded to the registry.
> 
>     - In all other situations, it sorts by the timestamp the container was
>       created.
> 
>   This algorithm exists to preserve ordering for containers that are moved
>   between registries.

kbelita:
Regarding the comment on [authentication](https://github.com/trisacrypto/directory/pull/757#pullrequestreview-1093961306) - I updated file to include GCR authentication and kept the Docker Hub authentication because we can use this same cleaner for Docker Hub. I just need the Docker Hub repository information so I can add it to this workflow.

bbengfort (approved):
@kbelita the DockerHub repositories are under the `trisa/` org: https://hub.docker.com/orgs/trisa/repositories

Would it be possible to have `-repo=trisa/` `-recursive=True` to capture them all?

PR #785, opened 2022-09-15
Fix inconsistent overview response

pdeziel:
### Scope of changes

This fixes a bug that existed in a code path that was untested in the bff overview handler. We were constructing a member details response for mainnet but not testnet which made the the response a bit inconsistent for the frontend. This removes some logic to make the response more consistent and adds a test case to make sure the code path is covered.

SC-9299

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

In the case where the user is not associated with a mainnet VASP, the overview endpoint should now return an empty string for the `status` field in the `MemberDetails` response for mainnet instead of `NO_VERIFICATION`.

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/pkg/bff/gds_test.go
+++ b/pkg/bff/gds_test.go
@@ -383,7 +383,6 @@ func (s *bffTestSuite) TestSubmitRegistration() {
 	// User metadata should be updated with the directory IDs
 	appdata := &auth.AppMetadata{}
 	require.NoError(appdata.Load(s.auth.GetUserAppMetadata()))
-	fmt.Println(appdata)
 	require.Equal("6041571e-09b4-47e7-870a-723f8032cd6c", appdata.VASPs.TestNet, "incorrect testnet directory id in user metadata")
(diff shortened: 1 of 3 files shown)


pkg/bff/gds_test.go
```diff
@@ -383,7 +383,6 @@ func (s *bffTestSuite) TestSubmitRegistration() {
 	// User metadata should be updated with the directory IDs
 	appdata := &auth.AppMetadata{}
 	require.NoError(appdata.Load(s.auth.GetUserAppMetadata()))
-	fmt.Println(appdata)
```
rebeccabilbro, line 386: 😅

rebeccabilbro (approved):
Thank you @pdeziel for making the behavior more symmetric between MainNet and TestNet. I'm struggling to recall why it was that we had different behavior for the two in the first place, but these changes certainly make things seem more sensible!

PR #808, opened 2022-09-23
Replace BFF trtl client with store client

pdeziel:
### Scope of changes

This replaces the BFF trtl client with the store client interface. Previously the BFF was using namespace "collections" and directly invoking the trtl client to do PUTs/GETs to the trtl database. This has been removed in favor of a struct which embeds a store client for doing CRUD operations against the store. This will enable the BFF to be able to interact with some of the directory service objects as well.

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [x] refactoring

### Acceptance criteria

Hopefully this hasn't broken anything, I think the only external changes we will have to make is updating the names of a few environment variables.

### Author checklist

### Reviewer(s) checklist


.env.template
```diff
@@ -180,9 +180,9 @@ GDS_BFF_MAINNET_ADMIN_TOKEN_KEYS=
 
 GDS_BFF_DATABASE_URL=trtl://localhost:4436/
 GDS_BFF_DATABASE_REINDEX_ON_BOOT=false
-GDS_BFF_DATABASE_MTLS_INSECURE=true
-GDS_BFF_DATABASE_MTLS_CERT_PATH=
-GDS_BFF_DATABASE_MTLS_POOL_PATH=
+GDS_BFF_DATABASE_INSECURE=true
+GDS_BFF_DATABASE_CERT_PATH=
+GDS_BFF_DATABASE_POOL_PATH=
```
pdeziel, line 185: We will have to remember to update the staging and production deployments.

pkg/store/config/config.go
```diff
@@ -0,0 +1,24 @@
...
+
+func (c StoreConfig) Validate() error {
+	// If the insecure flag isn't set then we must have certs when connecting to trtl.
+	if strings.HasPrefix(c.URL, "trtl://") && !c.Insecure {
+		if c.CertPath == "" || c.PoolPath == "" {
+			return errors.New("invalid configuration: connecting to trtl over mTLS requires certs and cert pool")
+		}
+	}
+	return nil
+}
```
pdeziel, line 24: Moved this configuration to a separate package to avoid the import cycle, so now this can be imported by both GDS and the BFF.

pkg/store/store.go
```diff
@@ -122,6 +122,7 @@ type CertificateStore interface {
 type AnnouncementStore interface {
 	RetrieveAnnouncementMonth(date string) (*bff.AnnouncementMonth, error)
 	UpdateAnnouncementMonth(m *bff.AnnouncementMonth) error
+	DeleteAnnouncementMonth(date string) error
```
pdeziel, line 125: We are not deleting announcement months in production code currently but this makes the tests a bit easier.

pkg/bff/db/organizations.go
```diff
@@ -1,118 +1,36 @@
...
-	}
-
-	if err = o.db.Put(ctx, uu[:], value, o.namespace); err != nil {
-		return nil, err
-	}
-
-	return org, nil
+// Create an organization and get back the organization stub with an assigned ID.
+func (store *DB) CreateOrganization() (*models.Organization, error) {
+	return store.db.CreateOrganization()
```
bbengfort, line 10: Is it necessary to have these indirects or the `db` module at all? Shouldn't we just delete the db module and use the store directly from the BFF?

pkg/store/config/config.go
bbengfort: Makes sense - I'll have to remember that this will change the configuration during the deployment.

.env.template
bbengfort, line 185: These will also have to be changed in `docker-compose.yaml`

bbengfort (approved):
Massive PR! Thank you so much for working so hard on refactoring this. It seems like we can take things a step farther and remove the `bff/db` package entirely; do you agree? A couple of minor comments but otherwise I think we're good to go.

pkg/bff/db/organizations.go
pdeziel: I agree that the db module is a bit unnecessary at this point, and we can just move the high level announcement methods to the BFF server struct instead.

PR #813, opened 2022-09-27
Delete collaborator

pdeziel:
### Scope of changes

This adds the BFF endpoint for deleting a collaborator from a user's organization.

SC-4306

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/bff/collaborators.go
```diff
@@ -93,3 +93,49 @@ func (s *Server) ReplaceCollaborator(c *gin.Context) {
...
+	// Make sure we can delete the collaborator by key
+	key := collaborator.Key()
+	if key == "" {
+		log.Warn().Err(err).Msg("missing key on collaborator")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse("invalid collaborator in request"))
+		return
+	}
+
+	// Delete the collaborator from the organization
+	delete(org.Collaborators, key)
```
pdeziel, line 131: The current code does not return an error if the collaborator doesn't exist, but I suppose we could check the map and return a 404 to ensure that the backend and frontend are on the same page?

bbengfort: @masskoder any thoughts about the status code you'd like returned?

bbengfort (approved):
Glad that putting these endpoints together is becoming more routine; everything looks great!

pkg/bff/collaborators.go
masskoder: @pdeziel @bbengfort  it would be great if we could have 200 (if okay ). 204 , if the collaborators ID is not found or invalid.

pdeziel: @masskoder 2XX codes are meant to be used for successful statuses, in my view an invalid collaborator key/ID is definitely an error. If we are changing the PUT and DELETE endpoints to accept IDs (e.g., `DELETE /v1/collaborators/:collabid`) perhaps the endpoint should return 400 for an invalid collabid and a 404 if the collaborator is missing. Does that make sense?

pkg/bff/api/v1/api.go
```diff
@@ -64,6 +65,11 @@ type StatusReply struct {
 // BFF v1 API Requests and Responses
 //===========================================================================
 
+// UpdateRolesParams contains a list of new roles for a collaborator.
+type UpdateRolesParams struct {
+	Roles []string `json:"roles"`
```
bbengfort, line 70: I think this is good future-proofing - but right now a TRISA user should only have one role; do you think we should re-architect the roles to make them compositional rather than hierarchical? E.g. right now an organizational leader has all the permissions of an organizational collaborator and a TRISA admin has all the permissions of an organizational leader.

pkg/bff/collaborators.go
```diff
@@ -34,13 +39,33 @@ func (s *Server) AddCollaborator(c *gin.Context) {
 		return
 	}
 
-	// Add the collaborator to the organization
-	if err = org.AddCollaborator(collaborator); err != nil {
-		log.Error().Err(err).Msg("could not add new collaborator to organization")
+	// Make sure the collaborator is valid for storage
+	if err = collaborator.Validate(); err != nil {
```
bbengfort, line 43: Nice - good use of a validator here!

pkg/bff/collaborators.go
```diff
@@ -34,13 +39,33 @@ func (s *Server) AddCollaborator(c *gin.Context) {
...
 		return
 	}
 
+	// Don't overwrite an existing collaborator
+	id := collaborator.Key()
+	if _, ok := org.Collaborators[id]; ok {
+		log.Warn().Str("collabID", id).Msg("collaborator already exists")
+		c.JSON(http.StatusConflict, api.ErrorResponse("collaborator already exists"))
+		return
+	}
```
bbengfort, line 55: Now I'm really glad that we're using the md5 hash! It creates an effective uniqueness check that I think will protect us in the long run.

pkg/bff/collaborators.go
```diff
@@ -34,13 +39,33 @@ func (s *Server) AddCollaborator(c *gin.Context) {
...
+
+	// Add the collaborator to the organization
+	if org.Collaborators == nil {
+		org.Collaborators = make(map[string]*models.Collaborator)
+	}
+	org.Collaborators[id] = collaborator
+
+	// TODO: We can search the email address in Auth0 to see if the user already exists
 	// TODO: Send invite/verification email to the collaborator
```
bbengfort, line 69: Do we have stories for these TODOs?

pkg/bff/collaborators.go
```diff
@@ -69,27 +94,198 @@ func (s *Server) ReplaceCollaborator(c *gin.Context) {
...
+	// Remove the user's existing roles so we can add the new ones
+	// TODO: There might be a more atomic way to do this in the management API
+	var userRoles *management.RoleList
+	if userRoles, err = s.auth0.User.Roles(collaborator.UserId); err != nil {
+		log.Error().Err(err).Str("collabID", collabID).Str("auth0_id", collaborator.UserId).Msg("could not fetch user roles from Auth0")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not fetch user roles from Auth0"))
+		return
+	}
+
+	if err = s.auth0.User.RemoveRoles(collaborator.UserId, userRoles.Roles); err != nil {
```
bbengfort, line 149: Very thorough, thank you for ensuring the old roles are removed!

pkg/bff/collaborators.go
```diff
@@ -69,27 +94,198 @@ func (s *Server) ReplaceCollaborator(c *gin.Context) {
...
+			return
+		}
+
+		// Update the app metadata with the removed organization
+		appdata.OrgID = ""
+		if err = s.SaveAuth0AppMetadata(*user.ID, *appdata); err != nil {
+			log.Error().Err(err).Str("collabID", collabID).Str("auth0_id", collaborator.UserId).Msg("could not save user app metadata")
+			c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not save user app metadata"))
+			return
+		}
```
bbengfort, line 244: We should check to see what happens when the user tries to log in without an OrgID -- I think they'll get an error right?

bbengfort (approved):
Looks like a great start to collaborator management! Thank you for pushing so hard on this! I guess the question now is how to ensure that Auth0 is in a valid state for users - e.g. if we delete a users organization ID - what happens when that user logs in again? Can we guarantee that an organization always has a leader, etc. But these are all next steps that we can move forward on now that you've laid the groundwork in this PR!

pkg/bff/api/v1/api.go
pdeziel: Interesting, I didn't realize that they were hierarchical and that would definitely simplify things, but I'm also fine with keeping the current implementation in case we decide to go that route (I don't think we need to make changes to the current roles). One thing I haven't figured out yet is how to handle the multi-org case. For example, if a user is an organization leader in an organization are they also a leader in every other organization they are a part of?

pkg/bff/collaborators.go
pdeziel: I'll create them before closing this story.

pkg/bff/collaborators.go
pdeziel: Hmm, I guess what would happen is that they would just get assigned a new OrgID. In other words the Login method doesn't currently distinguish between a new user and a user that has been removed from an org.

bbengfort: Oh good point - so probably not an issue!

pkg/bff/api/v1/api.go
bbengfort: Oh, that's a good point … shoot. I think they might be the way things are currently set up.

PR #816, opened 2022-09-28
Replace bff admin client with store client

pdeziel:
### Scope of changes

Previously the BFF was using its own authenticated client to connect to the Admin API in order to retrieve data from GDS. However this relied on the BFF having to create its own tokens for authenticating with the Admin API. This PR tries a different approach in the direction we've been taking with the recent BFF PRs, which allows the BFF to connect to the trtl database directly in order to retrieve the GDS records.

This involved a bit of refactoring, the Admin client was completely removed in favor of the generic store client. The BFF test suite has also been changed up a bit (it now requires running some trtl bufconn servers since we are directly interacting with the GDS records rather than the client mock).

SC-8655

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [x] refactoring

### Acceptance criteria

This should not change the overall behavior of the BFF except that the configuration will have to be updated again.

### Author checklist

### Reviewer(s) checklist


.env.template
```diff
@@ -174,9 +175,10 @@ GDS_BFF_MAINNET_MEMBERS_TIMEOUT=5s
 GDS_BFF_MAINNET_MEMBERS_MTLS_CERT_PATH=
 GDS_BFF_MAINNET_MEMBERS_MTLS_POOL_PATH=
 
-GDS_BFF_MAINNET_ADMIN_ENDPOINT=http://localhost:5434
-GDS_BFF_MAINNET_ADMIN_AUDIENCE=http://localhost:4433
-GDS_BFF_MAINNET_ADMIN_TOKEN_KEYS=
+GDS_BFF_MAINNET_DATABASE_URL=trtl://localhost:4436
+GDS_BFF_MAINNET_DATABASE_INSECURE=true
+GDS_BFF_MAINNET_DATABASE_CERT_PATH=
+GDS_BFF_MAINNET_DATABASE_POOL_PATH=
```
pdeziel, line 181: Note: Updated environment variables

pkg/bff/gds.go
```diff
@@ -445,6 +448,423 @@ func (s *Server) SubmitRegistration(c *gin.Context) {
 	c.JSON(http.StatusOK, out)
 }
 
+// GetCertificates makes parallel calls to the databases to get the certificate
+// information for both testnet and mainnet. If testnetID or mainnetID are empty
+// strings, this will simply return a nil response for the corresponding network so
+// the caller can distinguish between a non registration and an error.
+func (s *Server) GetCertificates(ctx context.Context, testnetID, mainnetID string) (testnetCerts, mainnetCerts []*models.Certificate, testnetErr, mainnetErr error) {
```
pdeziel, line 455: Moving the code from the former `admin.go` to here, since the methods don't have anything to do with the Admin API anymore and I couldn't think of a different name to organize them under.

bbengfort (approved):
Certainly a lot of refactoring to get to this point. I'm also going to have to remember that the BFF will now need to connect to trtl in a different namespace -- which is going to cause some devops headaches, but they are solvable. 

Let's take great care moving forward about what we allow the BFF to do in the database. We want to make sure that both the Admin API and the BFF are in sync with what changes they make to the database and we may want to extend our integration testing in the future to protect us from big regressions.

.env.template
bbengfort: Were these updated in docker-compose.yaml?

PR #864, opened 2022-11-01
List collaborators

pdeziel:
### Scope of changes

This completes the collaborators CRUD API by adding a list collaborators endpoint.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

The new endpoint returns the list of collaborator objects in the user's organization.

### Author checklist

### Reviewer(s) checklist

- [ ] Should collaborators be ordered by email address or some other means when they are returned to the frontend?


pkg/bff/collaborators.go
```diff
@@ -78,6 +79,40 @@ func (s *Server) AddCollaborator(c *gin.Context) {
...
+	}
+
+	for _, collab := range org.Collaborators {
+		if err = s.LoadCollaboratorDetails(collab); err != nil {
+			log.Error().Err(err).Str("collabID", collab.Key()).Msg("could not load collaborator details from Auth0")
+		}
+		out.Collaborators = append(out.Collaborators, collab)
+	}
+
+	// Sort by email address so we return consistent responses
```
pdeziel, line 108: I'm assuming this makes the frontend work easier, however it might not be required if it's going to be re-sorted anyway. One reason I added this was to make the output for this endpoint consistently ordered since iterating over the map is non-deterministic.

pkg/bff/collaborators.go
```diff
@@ -78,6 +79,40 @@ func (s *Server) AddCollaborator(c *gin.Context) {
...
+	}
+
+	// Build the response from the internal map
+	out := &api.ListCollaboratorsReply{
+		Collaborators: make([]*models.Collaborator, 0),
+	}
+
+	for _, collab := range org.Collaborators {
+		if err = s.LoadCollaboratorDetails(collab); err != nil {
+			log.Error().Err(err).Str("collabID", collab.Key()).Msg("could not load collaborator details from Auth0")
```
pdeziel, line 103: Should we return an error here? I'm leaning towards no, because at the worst case we just have stale data.

pkg/bff/api/v1/api.go
```diff
@@ -22,6 +22,7 @@ type BFFClient interface {
 
 	// Authenticated Endpoints
 	AddCollaborator(context.Context, *models.Collaborator) (*models.Collaborator, error)
+	ListCollaborators(context.Context) (*ListCollaboratorsReply, error)
```
bbengfort, line 25: I'm assuming no pagination is necessary for this?

pkg/bff/collaborators.go
```diff
@@ -251,6 +286,11 @@ func (s *Server) DeleteCollaborator(c *gin.Context) {
 // Auth0. The collaborator must have an user ID on it and the data in Auth0 will
 // overwrite the data on the collaborator record.
 func (s *Server) LoadCollaboratorDetails(collab *models.Collaborator) (err error) {
+	// If the user is not verified in Auth0 then we can't retrieve the details
+	if collab.VerifiedAt == "" {
+		return nil
```
bbengfort, line 291: If this returns `nil,` will an empty collaborator be added to the collaborator's list or is there some data on the collaborator even if it can't be retrieved?

pkg/bff/collaborators.go
```diff
@@ -78,6 +79,40 @@ func (s *Server) AddCollaborator(c *gin.Context) {
...
+		return
+	}
+
+	// Build the response from the internal map
+	out := &api.ListCollaboratorsReply{
+		Collaborators: make([]*models.Collaborator, 0),
+	}
+
+	for _, collab := range org.Collaborators {
+		if err = s.LoadCollaboratorDetails(collab); err != nil {
```
bbengfort, line 102: Does this make an auth0 request for every collaborator?

bbengfort (approved):
Looks good so far - just a question about the load collaborator detail request to auth0. I'm wondering if we should cache or parallelize those requests.

pkg/bff/api/v1/api.go
pdeziel: I suppose it depends on the number of collaborators we are expecting organizations to have. Probably not more than a dozen right?

pkg/bff/collaborators.go
bbengfort: Ok, so it will have data - it just might not be up to date; that works. 

Likely we'll want to implement some sort of cache as well to prevent large numbers of auth0 API requests. Would you mind creating a story for that?

pkg/bff/api/v1/api.go
bbengfort: Probably not - but perhaps we should set a fixed limit just in case (and for security). Could you create a story to limit the number of collaborators to 50? If a 51st collaborator is added, we should return an error instead.

pkg/bff/collaborators.go
```diff
@@ -78,6 +79,45 @@ func (s *Server) AddCollaborator(c *gin.Context) {
...
+		out.Collaborators = InsortCollaborator(out.Collaborators, collab, func(a, b *models.Collaborator) bool {
+			return a.Email < b.Email
+		})
+	}
+
+	// Collaborators exist on the organization record so we must persist the updated
+	// organization record to the database
+	if err = s.db.UpdateOrganization(org); err != nil {
+		log.Error().Err(err).Msg("could not save organization with updated collaborators")
+	}
```
pdeziel, line 116: Glad I took a second look at this!

pkg/bff/collaborators.go
bbengfort: If there are no collaborators on the organization, then the organization is marked for deletion; if the collaborators are being listed, it implies there is at least one -- the user who is listing the collaborators, right?

PR #874, opened 2022-11-07
Add orgID to login endpoint for multiorg support

pdeziel:
### Scope of changes

This mostly modifies the login endpoint to support multi-org login. There is a new orgID parameter which is optional on the request but allows the frontend to specify which organization to login to. Because of this, there is some additional logic implemented in the login endpoint to handle the new user, returning user, and invited user cases.

SC-10906

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

I was able to write some unit tests so most of the code paths are at least exercised but we should look closely at the implemented logic to make sure it matches what we intended.

### Author checklist

### Reviewer(s) checklist

- [ ] This PR is mostly about verifying that the login logic is correct and matches what we intended.


cmd/bff/main.go
```diff
@@ -216,7 +216,7 @@ func announce(c *cli.Context) (err error) {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
-	if err = client.Login(ctx); err != nil {
+	if err = client.Login(ctx, &api.LoginParams{}); err != nil {
```
bbengfort, line 219: Is it ok to pass in `nil` instead of an empty struct?

pkg/bff/auth/auth.go
```diff
@@ -249,6 +250,20 @@ func UserInfo(conf config.AuthConfig) (_ gin.HandlerFunc, err error) {
 			return
 		}
 
+		// User must have complete identity data for downstream processing
+		if user.ID == nil || *user.ID == "" || user.Email == nil || *user.Email == "" {
```
bbengfort, line 254: Nice short circuit conditional.

pkg/bff/auth/metadata.go
```diff
@@ -1,19 +1,47 @@
...
+)
 
 // AppMetadata makes it easier to serialize and deserialize JSON from the auth0
 // app_metadata assigned to the user by the BFF (and ensures the data is structured).
 type AppMetadata struct {
-	OrgID string `json:"orgid"`
-	VASPs VASPs  `json:"vasps"`
+	OrgID         string              `json:"orgid"`
+	VASPs         VASPs               `json:"vasps"`
+	Organizations map[string]struct{} `json:"organizations"`
```
bbengfort, line 14: Keep in mind that this will be serialized to JSON - are you trying to ensure that `Organizations` is a `set` or just trying to implement forward compatibility if we want to store other information with the Organization? If the answer is `set` then I think it might be better to use `[]string` here and enforce the set some other way in code.

pkg/bff/auth/metadata.go
```diff
@@ -1,19 +1,47 @@
...
 
+// TODO: Hash-based method might be more maintainable, but this avoids error handling for now
+func (meta *AppMetadata) Equals(other *AppMetadata) bool {
+	if meta.OrgID != other.OrgID {
+		return false
+	}
+
+	if meta.VASPs != other.VASPs {
+		return false
+	}
```
bbengfort, line 30: Note that struct equality only works the way you expect here if the `VASP` struct only has comparable fields:

https://golangbyexample.com/struct-equality-golang/

If we add say a `[]string` to `VASP` then this check will no longer work.

pkg/bff/auth/metadata.go
```diff
@@ -1,19 +1,47 @@
...
+
+	if len(meta.Organizations) != len(other.Organizations) {
+		return false
+	}
+
+	for i, org := range meta.Organizations {
+		if value, ok := other.Organizations[i]; !ok || value != org {
+			return false
+		}
+	}
```
bbengfort, line 40: Not to be a buzzkill, but I make this mistake a lot too with map/set equality. This only checks to make sure that all the key/value pairs in `meta.Organizations` are in and equal to the key/value pairs in `other.Organizations`. If there are additional key/value pairs in `other.Organizations` then this method will return true. 

e.g. the following organizations will be equal with this logic

```
meta.Organizations := map[string]struct{}{
    "apple": {},
    "banana": {},
    "orange": {},
}

other.Organizations := map[string]struct{}{
    "apple": {},
    "banana": {},
    "orange": {},
    "cantaloupe": {},
    "frostbite": {},
}
```

pkg/bff/auth/metadata.go
```diff
@@ -43,3 +71,45 @@ func (meta *AppMetadata) Dump() (appdata map[string]interface{}, err error) {
 
 	return appdata, nil
 }
+
+// ClearOrganization removes all organization-related data from the app metadata.
```
bbengfort, line 75: Good helper!

pkg/bff/auth/metadata.go
```diff
@@ -43,3 +71,45 @@ func (meta *AppMetadata) Dump() (appdata map[string]interface{}, err error) {
...
+		meta.VASPs.TestNet = org.Testnet.Id
+	} else {
+		meta.VASPs.TestNet = ""
+	}
+
+	if org.Mainnet != nil && org.Mainnet.Id != "" {
+		meta.VASPs.MainNet = org.Mainnet.Id
+	} else {
+		meta.VASPs.MainNet = ""
+	}
```
bbengfort, line 97:
```suggestion
	meta.ClearOrganization()
	meta.OrgID = org.Id

	if org.Testnet != nil {
		meta.VASPs.TestNet = org.Testnet.Id
	}

	if org.Mainnet != nil {
		meta.VASPs.MainNet = org.Mainnet.Id
	}
```

pkg/bff/collaborators.go
```diff
@@ -154,44 +154,19 @@ func (s *Server) UpdateCollaboratorRoles(c *gin.Context) {
 	}
 
 	// Collaborator needs to be a verified user in Auth0
-	if collaborator.VerifiedAt == "" || collaborator.UserId == "" {
+	if !collaborator.Verified || collaborator.UserId == "" {
```
bbengfort, line 157: That's a nicer check.

pkg/bff/organization.go
```diff
@@ -56,3 +48,19 @@ func (s *Server) OrganizationFromClaims(c *gin.Context) (org *models.Organizatio
 
 	return org, nil
 }
+
+func (s *Server) OrganizationFromID(id string) (org *models.Organization, err error) {
+	// Organizations are stored by UUID in the database
+	var uuid uuid.UUID
+	if uuid, err = models.ParseOrgID(id); err != nil {
+		log.Error().Err(err).Str("org_id", id).Msg("could not parse organization ID")
```
bbengfort, line 56: Let's try to push logging to the as close to the endpoint as possible; e.g. most logging should be in the handler; that will make it easier to trace where our logs are.

pkg/bff/users.go
```diff
@@ -27,24 +28,38 @@ const (
...
+		log.Error().Err(err).Msg("could not bind request")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse(err))
+		return
+	}
+
 	// Fetch the user from the context
 	if user, err = auth.GetUserInfo(c); err != nil {
 		log.Error().Err(err).Msg("login handler requires user info; expected middleware to return 401")
-		c.JSON(http.StatusInternalServerError, "could not identify user to login")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not identify user to login"))
```
bbengfort, line 62: 🙏🏽

pkg/bff/users.go
```diff
@@ -53,121 +68,213 @@ func (s *Server) Login(c *gin.Context) {
 	appdata := &auth.AppMetadata{}
 	if err = appdata.Load(user.AppMetadata); err != nil {
 		log.Error().Err(err).Msg("could not parse user app metadata")
-		c.JSON(http.StatusInternalServerError, "could not parse user app metadata")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not parse user app metadata"))
+		return
+	}
+
+	// Fetch the current user role.
+	if roles, err = s.auth0.User.Roles(*user.ID); err != nil {
```
bbengfort, line 76: I'm going to need a little more time with this logic; so I'm going to stop the review here so you can see and resolve the other comments. I'll try to get to this part of the review soon!

bbengfort (commented):
I've reviewed everything but the Login logic - I'm going to need a bit more time with it to understand what is going on but I wanted to make sure you had my review comments from the other section. I'll finish the rest of the review soon!

cmd/bff/main.go
pdeziel: Yes it does work.

pkg/bff/auth/metadata.go
pdeziel: We might want to have additional info in there at some point, however I think I agree that the `[]string` is the simplest for now.

pkg/bff/auth/metadata.go
pdeziel: Right, but I think that's a compile error so at least it won't catch us by surprise.

pkg/bff/auth/metadata.go
pdeziel: I had copilot generate this and I definitely thought it was wrong too, however the length check before the check you mentioned should cover that case. I'm going to change the organization map to a slice per your suggestion anyway though.

pkg/bff/auth/metadata.go
pdeziel: Nice, that makes it a lot cleaner.

pkg/bff/auth/metadata.go
bbengfort: Ah you're right about the length check - sorry about that!

pkg/bff/users.go
```diff
@@ -54,121 +68,213 @@ func (s *Server) Login(c *gin.Context) {
...
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not complete user login"))
 		return
 	}
 
-	// Retrieve the user's organization from the database
 	var org *models.Organization
-	if appdata.OrgID == "" {
-		// Create the organization
+	if params.OrgID == "" && appdata.OrgID == "" {
+		// This is a new user so create a new organization for them
```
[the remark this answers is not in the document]
bbengfort: Nevermind it looks like it does happen later!

pkg/bff/users.go
```diff
@@ -54,121 +68,213 @@ func (s *Server) Login(c *gin.Context) {
...
-		}
-		if org.Mainnet != nil && org.Mainnet.Id != "" {
-			appdata.VASPs.MainNet = org.Mainnet.Id
+		// Retrieve the user's collaborator record from the organization
+		var collaborator *models.Collaborator
+		if collaborator = org.GetCollaborator(*user.Email); collaborator == nil {
+			log.Debug().Str("email", *user.Email).Str("org_id", org.Id).Msg("could not find user in organization")
+			c.JSON(http.StatusUnauthorized, api.ErrorResponse("user is not authorized to access this organization"))
+			return
 		}
```
bbengfort, line 142: This section is critical for security; can we add some comments to it to ensure future maintainers guard against this case; also comments around the test for this case?

pkg/bff/users.go
```diff
@@ -54,121 +68,213 @@ func (s *Server) Login(c *gin.Context) {
...
+			}
 
-		// Update the organization in the database
-		if err = s.db.UpdateOrganization(org); err != nil {
-			log.Error().Err(err).Msg("could not update the organization with the new collaborator")
-			c.JSON(http.StatusInternalServerError, "could not complete user login")
-			return
-		}
+			// Remove the collaborator record from the previous organization
+			// TODO: This might require a user confirmation prompt
```
bbengfort, line 166: We will probably want a story for this - I think the idea is that this will be an additional endpoint to serve as confirmation.

bbengfort (approved):
Thank you for tackling this very tricky story; I know the Login logic is pretty tough to grok now, but your test coverage is excellent and that makes a big difference. As we continue to work on related stories, we can look for ways to break down the massive Login method into functional steps and look into optimizations to ensure that this endpoint doesn't take forever - but I'm glad we can move forward with it now!

pkg/bff/users.go
```diff
@@ -7,17 +7,17 @@ import (
...
 	"github.com/rs/zerolog/log"
 	"github.com/trisacrypto/directory/pkg/bff/api/v1"
 	"github.com/trisacrypto/directory/pkg/bff/auth"
 	"github.com/trisacrypto/directory/pkg/bff/models/v1"
 )
 
 const (
-	// TODO: do not hard code this value but make it a configuration
-	CollaboratorRole   = "Organization Collaborator"
+	// TODO: Need to make sure these roles are in sync with the roles in Auth0
```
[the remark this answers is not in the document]
bbengfort: Thank you!

PR #890, opened 2022-11-14
BFF Create and list organizations

pdeziel:
### Scope of changes

This adds two endpoints for creating and listing organizations, fulfilling sc-11106 and sc-11108. 

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

With these endpoints the frontend should be able to create organization records in the database and list the organizations they are a part of.

### Author checklist

- [ x]  I have recompiled and included new protocol buffers to reflect changes I made

### Reviewer(s) checklist

- [ ] How do we determine uniqueness for organizations? Do we need a domain or name index?
- [ ] Can only TSP users create organizations? Should we default the user to the created organization?


pkg/bff/organization.go
```diff
@@ -13,6 +15,145 @@ import (
...
+	// Don't allow the user to create duplicate organizations
+	// TODO: Should we do a universal check against the database using an index?
+	for _, id := range appdata.GetOrganizations() {
+		if org, err := s.OrganizationFromID(id); err != nil {
+			log.Error().Err(err).Str("org_id", id).Msg("could not retrieve organization from database")
+		} else if org.Domain == domain {
+			c.JSON(http.StatusConflict, api.ErrorResponse("organization with domain name already exists"))
+			return
+		}
+	}
```
pdeziel, line 72: Should we do a universal check against the entire database to determine uniqueness? That's going to require a bit more work as we currently don't have a way of querying all the organizations from trtl and we might need an index to speed up the search.

pkg/bff/organization.go
```diff
@@ -13,6 +15,145 @@ import (
...
+	org.Name = params.Name
+	org.Domain = domain
+	if err = s.db.UpdateOrganization(org); err != nil {
+		log.Error().Err(err).Msg("could not update organization in database")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not create organization"))
+		return
+	}
+
+	// Assign the user to the organization
+	appdata.AddOrganization(org.Id)
```
pdeziel, line 92: Should we also assign the user to the new organization so after the frontend does the refresh they are directed to the dashboard of the new organization?

pkg/bff/collaborators.go
```diff
@@ -223,7 +223,7 @@ func (s *Server) DeleteCollaborator(c *gin.Context) {
 
 	// If the collabroator is already verified in Auth0, then remove them from the
 	// organization
-	if !collaborator.Verified && collaborator.UserId != "" {
+	if collaborator.Verified && collaborator.UserId != "" {
```
bbengfort, line 226: Good catch!

pkg/bff/users.go
```diff
@@ -234,7 +234,7 @@ func (s *Server) Login(c *gin.Context) {
 	}
 
 	// If the user app metadata has changed, set the refresh flag in the response
-	if !oldAppdata.Equals(appdata) {
+	if appdata.Equals(oldAppdata) {
```
bbengfort, line 237: If `appdata` equals `oldAppdata` doesn't that imply that it hasn't changed? I guess I'm not understanding this change if you wouldn't mind help clarifying -- and if the logic has changed maybe the comment also needs an update?

pkg/bff/organization.go
```diff
@@ -13,6 +15,145 @@ import (
...
+		log.Error().Err(err).Msg("could not parse user app metadata")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not parse user app metadata"))
+		return
+	}
+
+	// Unmarshal the params from the POST request
+	params := &api.OrganizationParams{}
+	if err := c.ShouldBind(params); err != nil {
+		log.Warn().Err(err).Msg("could not bind request")
+		c.JSON(http.StatusBadRequest, &api.Reply{Error: err.Error()})
```
bbengfort, line 46:
```suggestion
		c.JSON(http.StatusBadRequest, api.ErrorResponse(err))
```

You can use `api.ErrorResponse` with `errors` as well if you'd like - it's a bit simpler.

pkg/bff/organization.go
```diff
@@ -13,6 +15,145 @@ import (
...
+		return
+	}
+
+	// Name is a required paramater
```
bbengfort, line 50:
```suggestion
	// Name is a required parameter
```

pkg/bff/organization.go
bbengfort, line 72: It might be helpful to put the search as a `Contains` method on `appdata` -- it seems like something we will want to do frequently and it will make it easier to optimize in the future (e.g. binary search with an ordered list of organizations).

pkg/bff/organization.go
```diff
@@ -13,6 +15,145 @@ import (
...
+	}
+
+	// Update the organization with the request params
+	org.Name = params.Name
+	org.Domain = domain
+	if err = s.db.UpdateOrganization(org); err != nil {
+		log.Error().Err(err).Msg("could not update organization in database")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not create organization"))
+		return
+	}
```
bbengfort, line 89: Are two DB queries necessary? What is `CreateOrganization` doing that requires us to update the org and then save it again? E.g. can we get name and domain into the organization at create time?

pkg/bff/organization.go
```diff
@@ -13,6 +15,145 @@ import (
...
+			})
+		}
+	}
+
+	c.JSON(http.StatusOK, out)
+}
+
+// Normalize a domain name for matching purposes.
+func normalizeDomain(domain string) string {
+	return strings.ToLower(strings.TrimRight(strings.TrimSpace(domain), "."))
```
bbengfort, line 154: I'm thinking that we'll want to strip off any `http://` that's part of the domain as well, so we could perhaps optimistically parse the URL (but if it doesn't parse, then ignore it).

```suggestion
	if u, err := url.Parse(domain); err != nil {
		domain = u.Host()
	}    
	return strings.ToLower(strings.TrimRight(strings.TrimSpace(domain), "."))
```

We may also want to have a regular expression to validate the domain - something pretty permissive, but to prevent things like just `"foo"` or an email address. E.g. `^(((?!\-))(xn\-\-)?[a-z0-9\-_]{0,61}[a-z0-9]{1,1}\.)*(xn\-\-)?([a-z0-9\-]{1,61}|[a-z0-9\-]{1,30})\.[a-z]{2,}$` from https://stackoverflow.com/questions/10306690/what-is-a-regular-expression-which-will-match-a-valid-domain-name-without-a-subd -- though some tests would be good to make sure it does what we want.

bbengfort (commented):
@pdeziel looks good so far! I had a number of comments and questions but we should be able to address them pretty quickly in a second pass!

pkg/bff/users.go
pdeziel: Whoops, I think we got borked by merge conflict resolution here. Thanks for catching it!

pkg/bff/server.go
```diff
@@ -347,6 +347,11 @@ func (s *Server) setupRoutes() (err error) {
 		v1.GET("/users/roles", s.ListUserRoles)
 
 		// Authenticated routes
+		organizations := v1.Group("/organizations")
+		{
+			organizations.GET("", auth.Authorize("read:organizations"), userinfo, s.ListOrganizations)
+			organizations.POST("", auth.DoubleCookie(), auth.Authorize("create:organizations"), userinfo, s.CreateOrganization)
```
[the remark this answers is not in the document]
pdeziel: I was going to add the `read:organizations` and `create:organizations` permission to the TSP, unless we just want to collapse it into a single `switch:organizations` which is what you were thinking?

pkg/bff/organization.go
pdeziel: Thanks, I forgot about that one.

pkg/bff/organization.go
pdeziel: Agreed!

pkg/bff/organization.go
pdeziel: I think it's definitely possible to do that, we just have to add or edit a trtl method. This is the first time we've known anything about the organization before creating it since we were previously just creating them for new users.

pkg/bff/server.go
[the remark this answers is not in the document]
bbengfort: I'm ok with either - my original idea was that `switch:organizations` would be sufficient - but if you think we should make the permissions more granular, then I'm ok with it. Just let me know which way you're planning on going!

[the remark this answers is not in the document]
bbengfort: Oh and also please document what perms needed to be added for when we do the deployment.

pkg/bff/organization.go
bbengfort: Ok, I think that might be useful to do if that's ok with you.

pkg/bff/domain.go
```diff
@@ -0,0 +1,43 @@
...
+	if u, err := url.ParseRequestURI(domain); err == nil {
+		domain = u.Hostname()
+	}
+
+	// Convert to lowercase and remove trailing characters.
+	domain = strings.ToLower(strings.TrimSpace(domain))
+
+	// Convert international domain names to ASCII
+	var err error
+	if domain, err = idna.ToASCII(domain); err != nil {
```
bbengfort, line 28: Very nice! I'll have to remember this snippet to use in Ensign as well!

cc @daniellemaxwell

pkg/bff/domain_test.go
```diff
@@ -0,0 +1,73 @@
...
+		{"foo.bar.com", true, "valid domain with subdomain"},
+		{"foo-bar.com", true, "valid domain with hyphen"},
+		{"foo.bar.bar2.com", true, "valid domain with two subdomains"},
+		{"foo.bar.bar2.bar3.bar4.com", true, "valid domain with multiple subdomains"},
+		{"google.com.au", true, "valid domain with two trailing characters"},
+		{"xn--bcher-kva.ch", true, "valid domain with international characters"},
+		{"foo.xn--com", true, "valid domain with international characters and subdomain"},
+		{"foo.bar.xn--com", true, "valid domain with international characters and subdomain"},
+		{"xn-fsqu00a.xn-0zwm56d", true, "valid domain with international characters and numbers"},
+		{"trisa-a1234b1234c1234d1234e1234f123456.example.longdomains.com", true, "valid domain with long subdomain"},
```
bbengfort, line 51: Great set of test cases!

pkg/store/leveldb/leveldb.go
```diff
@@ -610,26 +610,27 @@ func (s *Store) DeleteAnnouncementMonth(date string) (err error) {
...
-func (s *Store) CreateOrganization() (o *bff.Organization, err error) {
-	// Create an empty organization
-	ts := time.Now().Format(time.RFC3339Nano)
-	uu := uuid.New()
-	o = &bff.Organization{
-		Id:       uu.String(),
-		Created:  ts,
-		Modified: ts,
+// unique ID if it doesn't exist and setting the created and modified timestamps.
+func (s *Store) CreateOrganization(o *bff.Organization) (id string, err error) {
```
bbengfort, line 614: Nicely done!

pkg/bff/users_test.go
```diff
@@ -107,23 +107,25 @@ func (s *bffTestSuite) TestReturningUserLogin() {
...
+		Testnet: &models.DirectoryRecord{
+			Id: metadata.VASPs.TestNet,
+		},
+		Mainnet: &models.DirectoryRecord{
+			Id: metadata.VASPs.MainNet,
+		},
 	}
-	require.NoError(s.DB().UpdateOrganization(org), "could not update organization")
+	_, err = s.DB().CreateOrganization(org)
+	require.NoError(err, "could not create organization")
```
bbengfort, line 120: Much cleaner!

pkg/bff/organization.go
bbengfort: Yes, I think that's a very good idea.

bbengfort (approved):
Looks great, thanks for making all of those changes @pdeziel!

PR #893, opened 2022-11-16
Add login timestamps to collaborator

pdeziel:
### Scope of changes

This adds two new metadata timestamps to the collaborator record. The purpose of this is to better inform the frontend about when collaborators have accepted invitations.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Timestamps make sense and I didn't miss any spots where they need to be updated.

### Author checklist

### Reviewer(s) checklist


proto/bff/models/v1/models.proto
```diff
@@ -52,6 +52,8 @@ message Collaborator {
     // Metadata as RFC3339Nano Timestamps
     string created_at = 10;
     string modified_at = 11;
+    string joined_at = 12;
+    string last_login = 13;
```
pdeziel, line 56: The last login timestamp will be returned on `ListOrganizations` requests so that the frontend can order the organizations for the user by most recently used.

DanielSollis: So the frontend will order the collaborators multiple organizations by the last login of the collaborator?

pdeziel: The problem is that is has to be per-user and the `Organization` record has multiple collaborators on it (in the map). So what I was thinking was that on `ListOrganizations` as we are iterating through the organizations we would just lookup the collaborator from the map by ID (which is a hashed version of the user's email address) and add the timestamp to the reply for that organization.

DanielSollis (approved):
Looks good, glad I could take a look at an area of GDS i'm not as familiar with.

PR #895, opened 2022-11-17
Auth0 user invite tickets

pdeziel:
### Scope of changes

This adds some functionality to the BFF to generate user invite tickets. There are two types of tickets the BFF generates:

1. Password reset - this should be generated when we need to create a user account for a collaborator on their behalf (after they are added as a collaborator), and allows them to change their password from the generated one.
2. Verification email - this is currently used to ensure the user is still verified in Auth0 and redirect them to the authentication callback after they click the link.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

With this change the user should receive a user invite email when they are added as a collaborator in the organization. If they click the link they should be able to change their password if they are a new user, or be automatically redirected to the configured GDS_BFF_AUTH0_REDIRECT_URL if they are an existing registered user.

### Author checklist

### Reviewer(s) checklist


pkg/bff/collaborators.go
```diff
@@ -300,6 +348,50 @@ func (s *Server) LoadCollaboratorDetails(collab *models.Collaborator) (err error
...
+// organization and returns the ticket URL. Depending on the current state of the user,
+// this will either be a password change ticket or an email verification ticket. The
+// caller is responsible for sending the user the ticket URL, which is usually done by
+// including it in the user invite email.
+func (s *Server) GetAuth0UserInviteURL(user *management.User, org *models.Organization) (inviteURL *url.URL, err error) {
+	// Ticket should include the redirect to the auth callback and should expire in 7 days
+	redirectURL := fmt.Sprintf("%s?orgid=%s", s.conf.Auth0.RedirectURL, org.Id)
+	expiration := int((time.Hour * 24 * 7).Seconds())
+	ticket := &management.Ticket{
+		UserID:    user.ID,
```
DanielSollis, line 361: Is it possible user could be Nil here? If so you could do a check first and wouldn't need to have a compound If on 366.

DanielSollis (commented):
This looks good, I just had one comment suggesting a possible change.

pkg/bff/collaborators.go
pdeziel: It is possible, although we've been relying on tests to assert that anything calling this method won't pass in nil values and at the worst case Auth0 gives us back a nil pointer and we go into panic recovery (although unlikely since the Auth0 SDK should have returned an error).

The reason for the nil check on 366 is because the Auth0 SDK declares the [management.User](https://pkg.go.dev/github.com/auth0/go-auth0@v0.7.0/management#User) fields as pointers. So it's not checking that the `user` pointer is nil but rather that the `EmailVerified` field is nil before we try to dereference it.

DanielSollis: Ok, that makes sense

DanielSollis (approved):

PR #916, opened 2022-11-28
Replace Auth0 invites with BFF invites

pdeziel:
### Scope of changes

This simplifies the user invite process for new collaborators by removing the Auth0 interactions and handling the workflow entirely within the BFF. The user invitation is now stored as state on the collaborator record instead of relying on the Auth0 password reset and verification flows. The invitation link in the invite email now directs the user to the login or register frontend page depending on if they are an existing Auth0 user.

This change involves deprecating some configuration values in favor of some new ones.

SC-11458

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Users invited by collaborators should now be provided with a direct link to the frontend rather than the Auth0 invitation.

### Author checklist

### Reviewer(s) checklist

- [ ] The two new configuration values make sense and are appropriately tested
- [ ] The collaborator invite logic makes sense


pkg/bff/collaborators.go
```diff
@@ -65,74 +64,60 @@ func (s *Server) AddCollaborator(c *gin.Context) {
...
+			log.Warn().Err(err).Str("email", collaborator.Email).Msg("collaborator already exists")
+			c.JSON(http.StatusConflict, api.ErrorResponse(err))
+		case errors.Is(err, models.ErrMaxCollaborators):
+			log.Warn().Err(err).Int("maximum", models.MaxCollaborators).Msg("maximum number of collaborators reached")
+			c.JSON(http.StatusForbidden, api.ErrorResponse(err))
+		default:
+			log.Error().Err(err).Msg("could not add collaborator")
+			c.JSON(http.StatusInternalServerError, api.ErrorResponse(err))
+		}
+		return
```
pdeziel, line 86: Reducing some repeated code here between `collaborators.go` and `models/v1/organization.go`. We still want to have specific HTTP responses returned, so I've added some defined errors here.

pkg/bff/collaborators.go
```diff
@@ -65,74 +64,60 @@ func (s *Server) AddCollaborator(c *gin.Context) {
...
-			Email:       &collaborator.Email,
-			Password:    &password,
-			Connection:  &s.conf.Auth0.ConnectionName,
-			VerifyEmail: &verifyEmail,
-		}
-		if err = s.auth0.User.Create(user); err != nil {
-			log.Error().Err(err).Str("email", collaborator.Email).Msg("error creating user in Auth0")
-			c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not add collaborator"))
-			return
+			Email: &collaborator.Email,
```
pdeziel, line 102: Previously we were creating the Auth0 user on the user's behalf (normally it would happen when they sign up on the frontend) if they didn't exist yet. However, this just complicated things because it involved maintaining state in both the BFF and Auth0 and it was difficult to tell which users had verified their email. To simplify this, the new code just creates the collaborator record in Auth0 and all users still have to go through the standard registration process to create their Auth0 account.

pkg/bff/config/config.go
```diff
@@ -33,6 +33,8 @@ type Config struct {
 	LogLevel     logger.LevelDecoder `split_words:"true" default:"info"`
 	ConsoleLog   bool                `split_words:"true" default:"false"`
 	AllowOrigins []string            `split_words:"true" default:"http://localhost,http://localhost:3000,http://localhost:3003"`
+	RegisterURL  string              `split_words:"true" required:"true"`
+	LoginURL     string              `split_words:"true" required:"true"`
```
pdeziel, line 37: These configuration values will have to be added when we do the next staging deployment.

pkg/bff/docs/docs.go
```diff
@@ -1333,6 +1339,9 @@ const docTemplate = `{
                 "email": {
                     "type": "string"
                 },
+                "expires_at": {
+                    "type": "string"
+                },
```
pdeziel, line 1344: The files in `docs` are autogenerated so treat them as `.pb` files for the purposes of review.

pkg/bff/users.go
```diff
@@ -170,8 +171,16 @@ func (s *Server) Login(c *gin.Context) {
...
+		// Verify that pending invitations have not expired
+		if err = collaborator.ValidateInvitation(); err != nil {
+			log.Debug().Err(err).Str("email", collaborator.Email).Str("org_id", org.Id).Msg("invalid user invitation")
+			c.JSON(http.StatusForbidden, api.ErrorResponse("user invitation has expired"))
+			return
+		}
+
 		// Other endpoints expect the user's verification status to be up to date
 		collaborator.Verified = *user.EmailVerified
+		collaborator.ExpiresAt = ""
```
pdeziel, line 183: This is the final part of the invite workflow when the user logs in and the `orgid` is included in the request. At this point, we need to verify that the organization they are trying to log into has their "invitation" record and that the invite has not expired. We will eventually need a way for users to resend invitations, although currently they can accomplish this by deleting and re-inviting the user.

.env.template
```diff
@@ -143,12 +143,11 @@ GDS_BFF_LOG_LEVEL=info
 GDS_BFF_CONSOLE_LOG=true
 GDS_BFF_ALLOW_ORIGINS="http://localhost:3000,http://localhost:3003"
 GDS_BFF_COOKIE_DOMAIN=localhost
+GDS_BFF_LOGIN_URL=
+GDS_BFF_REGISTER_URL=
 
 GDS_BFF_SERVE_DOCS=true
 
-GDS_BFF_AUTH0_DOMAIN=
```
rebeccabilbro, line 149: Is `GDS_BFF_AUTH0_DOMAIN` still needed? It looks like we still have it in the docker-compose file

pkg/bff/collaborators.go
```diff
@@ -65,74 +64,60 @@ func (s *Server) AddCollaborator(c *gin.Context) {
...
+		switch {
+		case errors.Is(err, models.ErrInvalidCollaborator):
+			log.Warn().Err(err).Msg("invalid collaborator")
+			c.JSON(http.StatusBadRequest, api.ErrorResponse(err))
+		case errors.Is(err, models.ErrCollaboratorExists):
+			log.Warn().Err(err).Str("email", collaborator.Email).Msg("collaborator already exists")
+			c.JSON(http.StatusConflict, api.ErrorResponse(err))
+		case errors.Is(err, models.ErrMaxCollaborators):
+			log.Warn().Err(err).Int("maximum", models.MaxCollaborators).Msg("maximum number of collaborators reached")
+			c.JSON(http.StatusForbidden, api.ErrorResponse(err))
```
rebeccabilbro, line 81: Will this error still happen? It looks like we have removed the `MaxCollaborators` const.

pkg/bff/config/config.go
rebeccabilbro: Should we add a little inline comment reminder that trailing slashes are not allowed for these two values?

pkg/bff/docs/swagger.yaml
```diff
@@ -816,6 +818,10 @@ paths:
           description: Unauthorized
           schema:
             $ref: '#/definitions/api.Reply'
+        "403":
+          description: User invitation has expired
+          schema:
+            $ref: '#/definitions/api.Reply'
```
rebeccabilbro, line 824: Thanks for keeping these docs up-to-date!

pkg/bff/collaborators.go
rebeccabilbro: Nevermind! Looks like it's just moved to `pkg/bff/models/v1/organization.go`

pkg/bff/collaborators.go
rebeccabilbro: Do we need an additional test to cover the case where an existing Auth0 user without an org now needs to log in and become associated with the org?

pkg/bff/models/v1/collaborator.go
```diff
@@ -39,3 +40,22 @@ func (collab *Collaborator) Validate() error {
 	// TODO: More comprehensive validation may be required
 	return nil
 }
+
+// Helper to determine if a collaborator invite is valid based on the expiration date.
+// If there is no expiration date, this method assumes that the collaborator invitation
+// is still valid.
```
rebeccabilbro, line 46: Trying to think through when this would be the case (that there is no expiration) -- is this an edge case for people who registered between the last release and the upcoming release?

rebeccabilbro (approved):
Looks good @pdeziel -- I made a few notes and had a few questions since I've been a bit out of the loop on the new collaborators feature!

.env.template
pdeziel: Whoops, thanks for catching this, it's a bit of a cognitive load to keep the configuration values in sync across the three different places!

pkg/bff/models/v1/collaborator.go
pdeziel: When a collaborator logs in successfully and "accepts" the invite we clear out the expiration date, which allows them to log in successive times after they have accepted the invite and joined the organization. This means that it's important we clear out the expiration date on login so that we know they are actually a returning user. There are actually a few other ways this could happen though:

1. The user logs in for the first time and a default organization is created for them and they are added as a collaborator - there is no expiration because there was no invite and they are considered an "Organization Leader".
2. A TSP user creates a new organization for a new VASP - similarly to 1 they are added as a collaborator in the new organization and we don't need to send them an invite.

pkg/bff/collaborators.go
pdeziel: I think this is covered in the "user invite" test - we have a case where the user is logging in with an `orgid` in the request but has no current `orgid` in the Auth0 metadata - in which case we have to check the database to see if they were invited. https://github.com/trisacrypto/directory/blob/cfd08c2be9d8d2f0002e17dc724e65bd6d4f4c35/pkg/bff/users_test.go#L196

PR #918, opened 2022-11-29
Return last login timestamp

pdeziel:
### Scope of changes

This PR adds a field to the `OrganizationReply` responses that are returned to the frontend. This timestamp represents when the user last logged into the organization. In a previous PR we added this timestamp to the collaborator record so that it gets updated when the frontend calls `Login` to complete the user login. The frontend needs this timestamp in the organization list reply so that we can order the available organizations in a useful way to the user (e.g. TSP users will likely have many organizations so the list could get quite long).

SC-11457

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

The `GET /v1/organizations` endpoint should now return the user's last login timestamp as part of each organization record response.

### Author checklist

### Reviewer(s) checklist

- [ ] The main question here is if we want to return an error if we can't find the user on the organization record or just log the error and skip listing that organization.


rebeccabilbro (approved):
lgtm!

PR #931, opened 2022-12-02
Auth0 user cache

pdeziel:
### Scope of changes

This adds an Auth0 user data cache. The main motivator of this is that we wanted to reduce the number of calls when the user loads the collaborators page. Currently a separate call has to be made for each collaborator to fetch user data such as their name, email address, etc. This PR introduces a TTL cache so that some user data can be saved in memory. After a configurable time, user data will expire and has to be refreshed from Auth0.

The goal is that we can find some configuration values that make sense, so that we can reduce the latency of the `ListCollaborators` call by retrieving user data from the cache while keeping the data relatively up to date.

SC-10884

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/bff/auth/cache/cache.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (c *TTLCache) SetStore(store KeyValueStore) {
+	c.store = store
+}
+
+// Get returns a value by key from the cache. If the key is not found or the value has
+// expired, we have to get a new value from the fetcher and handle evictions to abide
+// by the cache size limit.
+func (c *TTLCache) Get(key string) (value interface{}, err error) {
```
pdeziel, line 44: This is a critical part of the code and it's where we would see race conditions/deadlocks so it's important that it's vetted closely.

pkg/bff/auth/cache/cache.go
```diff
@@ -0,0 +1,88 @@
...
+// TTLCache is a time-to-live cache that relies on a key value store for persistence,
+// automatically refreshing entries when they expire and uses a FIFO eviction policy
+// to manage the cache size.
+type TTLCache struct {
+	sync.RWMutex
+	fetcher ResourceFetcher
+	store   KeyValueStore
+	conf    config.CacheConfig
+	keys    []string
+}
```
pdeziel, line 23: I was thinking we eventually might want to use this for other Auth0 requests as well or utilize a different type of cache storage (e.g. disk) so this implementation is pretty generic.

pkg/bff/config/config.go
```diff
@@ -43,6 +43,7 @@ type Config struct {
 	Database     config.StoreConfig
 	Email        EmailConfig
 	Sentry       sentry.Config
+	UserCache    CacheConfig `split_words:"true"`
```
pdeziel, line 46: For some reason the variables were not being picked up from the environment without this tag.

containers/docker-compose.yaml
```diff
@@ -178,6 +178,9 @@ services:
       - GDS_BFF_AUTH0_AUDIENCE=https://bff.vaspdirectory.dev
       - GDS_BFF_AUTH0_CLIENT_ID
       - GDS_BFF_AUTH0_CLIENT_SECRET
+      - GDS_BFF_USER_CACHE_ENABLED
```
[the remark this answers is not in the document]
pdeziel: I could go either way but there is some different frontend behavior with the cache on vs off - for instance if someone's role gets updated from the Auth0 dashboard the update might not be visible until the expiration happens. We may want the ability to test some of those things at least locally.

pkg/bff/config/config.go
pdeziel: Aha, thanks for clearing that up.

pkg/bff/auth/cache/ttl.go
```diff
@@ -0,0 +1,75 @@
+package cache
```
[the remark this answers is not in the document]
pdeziel: That seems better to me.

pkg/utils/cache/disabled.go
```diff
@@ -0,0 +1,17 @@
...
+// Disabled implements the Cache interface.
+var _ Cache = &Disabled{}
+
+func (d *Disabled) Get(key interface{}) (data interface{}, ok bool) {
+	return nil, false
+}
+
+func (d *Disabled) Add(key interface{}, data interface{}) {}
+
+func (d *Disabled) Remove(key interface{}) {}
```
bbengfort, line 17: Excellent! The compiler will optimize these with inlining - essentially removing the calls to Add and Remove entirely if it is disabled.

pkg/utils/cache/ttl.go
```diff
@@ -0,0 +1,68 @@
...
+	cache = &TTL{
+		conf: conf,
+	}
+
+	if cache.lru, err = lru.New(int(conf.Size)); err != nil {
+		return nil, err
+	}
+
+	return cache, nil
+}
```
bbengfort, line 27: Let's add the enabled/disabled behavior down to the `cache` module rather than have the external caller decide.

```suggestion
func New(conf config.CacheConfig) (Cache, error) {
	if !conf.Enabled {
		return &Disabled{}, nil
	}
	return NewTTL(conf)
}

func NewTTL(conf config.CacheConfig) (cache *TTL, err error) {
	cache = &TTL{
		conf: conf,
	}

	if cache.lru, err = lru.New(int(conf.Size)); err != nil {
		return nil, err
	}

	return cache, nil
}
```

pkg/bff/server.go
```diff
@@ -104,6 +105,16 @@ func New(conf config.Config) (s *Server, err error) {
...
 		}
+
+		// Initialize the user cache or use a no-op cache if disabled
+		if s.conf.UserCache.Enabled {
+			if s.users, err = cache.NewTTL(s.conf.UserCache); err != nil {
+				return nil, fmt.Errorf("could not initialize user cache: %s", err)
+			}
+		} else {
+			s.users = &cache.Disabled{}
+		}
```
bbengfort, line 116: See below

```suggestion
		// Initialize the user cache or use a no-op cache if disabled
		if s.users, err = cache.New(s.conf.UserCache); err != nil {
			return nil, fmt.Errorf("could not initialize user cache: %w", err)
		}
```

bbengfort (approved):
Thanks for making those changes!

PR #936, opened 2022-12-05
Add switch:organizations permission

pdeziel:
### Scope of changes

This adds a `switch:organizations` permission so that the Login logic relies solely on permissions and not roles. We encountered an issue recently where a user with the TRISA Administrator permission was allowed to create organizations but not join them. This is because we had some logic which was relying on the user role and not the underlying permission.

The bottom line is that the BFF shouldn't be making decisions based on user roles; although there are some places where we have to assign users a specific role (for instance so that they can add collaborators to their own organization). Instead, roles in Auth0 should just be a collection of permissions.

This PR also takes the opportunity to do some cleanup with the permissions to avoid typos.

SC-11712

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/bff/users.go
```diff
@@ -189,21 +194,23 @@ func (s *Server) Login(c *gin.Context) {
...
 		// Non-TSP users can only exist in one organization
 		if appdata.OrgID != "" && org.Id != appdata.OrgID {
-			// When switching to a new organization, make sure leader roles are not
-			// unintentionally preserved
-			if userRole == LeaderRole && len(org.Collaborators) > 1 {
-				userRole = CollaboratorRole
+			// Users should not be able to add collaborators if they are migrated to an
+			// existing organization.
+			if claims.HasPermission(auth.UpdateCollaborators) && len(org.Collaborators) > 1 {
+				userRole = auth.CollaboratorRole
```
pdeziel, line 213: Here is the main part of the PR where we check the user permission in the claims rather than checking the user role.

PR #941, opened 2022-12-06
Add recipient address to email log

pdeziel:
### Scope of changes

This adds the recipient email address to the contact email log, which will make it a bit easier to determine where emails were sent when looking at the unified email log.

This PR has a low complexity but a few tests and other files had to be changed as a result of this.

SC-11179

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/models/v1/models.go
```diff
@@ -176,24 +176,15 @@ func AppendAdminEmailLog(vasp *pb.VASP, reason string, subject string) (err erro
...
 }
 
-// VASPEmailEntry represents a log entry relevant to a single contact. It embeds an
-// email log entry and adds contact-specific information so that higher level methods
-// reading a unified contact log can distinguish between entries for different contacts.
-// TODO: This metadata should really be stored in the email log entry itself.
-type VASPEmailEntry struct {
-	*EmailLogEntry
-	ContactType string
-}
```
pdeziel, line 186: We were previously using this intermediate struct but now that we have the email address in the log this is no longer necessary.

rebeccabilbro (approved):
Thank you @pdeziel , this all makes sense to me -- from the perspective of the log it's much more important to know the contact's email address than what type of contact they are. When we need to sift through the log, it's nearly always because someone doesn't remember receiving an email that the system says it sent, or because the verification steps happened out of order due to human error, and certificates were sent to the first verified contact rather than the most appropriate (e.g. technical) point-of-contact.

PR #960, opened 2022-12-13
sc-8667: Cleanup duplicate code in models.go

DanielSollis:
### Scope of changes

There is some duplicated code between GetSentAdminEmailCount in models.go and GetSentEmailCount in contacts.go, this PR  refactors the duplicated code into a helper function.

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [x] other (Refactoring to remove duplicated code)

### Acceptance criteria

 - Run the directory tests

### Author checklist

### Reviewer(s) checklist


pkg/models/v1/contacts.go
```diff
@@ -311,19 +311,8 @@ func GetSentEmailCount(contact *pb.Contact, reason string, timeWindowDays int) (
...
-		}
-
-		matchedReason := reason == value.Reason
-		withinTimeWindow := timestamp.After(time.Now().AddDate(0, 0, -timeWindowDays))
-
-		if matchedReason && withinTimeWindow {
-			sent++
-		}
+	if sent, err = countSentEmails(emailLog, reason, timeWindowDays); err != nil {
+		return 0, err
```
pdeziel, line 315: This is great, I'm happy we're getting rid of the duplicated code. To take this a step farther, we could probably also remove the `GetSentEmailCount` and `GetSentAdminEmailCount` in favor of the new helper. This means that we have to fetch the email log wherever those were being called but I think it will reduce the bloat in the models package a bit more and prevent us from having to pass through the `reason` and `timeWindowDays`. What do you think?

pkg/models/v1/models.go
```diff
@@ -237,14 +237,22 @@ func GetVASPEmailLog(vasp *pb.VASP) (emails []*VASPEmailEntry, err error) {
...
 func GetSentAdminEmailCount(vasp *pb.VASP, reason string, timeWindowDays int) (sent int, err error) {
 	var adminEmailLog []*EmailLogEntry
 	if adminEmailLog, err = GetAdminEmailLog(vasp); err != nil {
 		return 0, err
 	}
 
-	for _, value := range adminEmailLog {
+	if sent, err = countSentEmails(adminEmailLog, reason, timeWindowDays); err != nil {
+		return 0, err
+	}
```
pdeziel, line 249: The caller has to check the error anyway so we can just return `return countSentEmails(adminEmailLog, reason, timeWindowDays)` here. Alternatively, since this is looking more like a pass-through method after your refactoring we could probably get rid of this method as described in my other comment.

pdeziel (approved):
Looks great @DanielSollis , I'm glad we're taking some time to clean up the models package a bit. I think we could take a step farther and reduce a lot more, let me know what you think of my suggestion below.

pdeziel (approved):
This looks a lot better - thanks for making the changes!

PR #970, opened 2022-12-26
Debug Sentry Transactions

bbengfort:
### Scope of changes

Updates the use of our interceptors for better Sentry handling.

Fixes SC-12088

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

This is mostly a refactoring of our interceptors so that we chain them together more like standard middleware. However, we want to ensure that no part of our previous middlewares is missed and that the behavior of this new middleware should be the same as the old. 

### Author checklist

### Reviewer(s) checklist


pkg/utils/sentry/config.go
```diff
@@ -13,7 +13,10 @@ type Config struct {
 	Environment      string  `split_words:"true"`
 	Release          string  `split_words:"true"`
 	TrackPerformance bool    `split_words:"true" default:"false"`
-	SampleRate       float64 `split_words:"true" default:"1.0"`
+	SampleRate       float64 `split_words:"true" default:"0.85"`
+	Service          string  `split_words:"true"`
+	ReportErrors     bool    `split_words:"true" default:"false"`
+	Repanic          bool    `split_words:"true" default:"false"`
```
bbengfort, line 19: This will require an update to the configuration documentation.

bbengfort: Story created.

pkg/gds/interceptor.go
```diff
@@ -3,141 +3,93 @@ package gds
...
-			log.WithLevel(zerolog.PanicLevel).Err(fmt.Errorf("unknown service type: %T", info.Server))
-			panicked = false
-			return nil, status.Error(codes.Unimplemented, "unknown service type for request")
-		}
-		sentry.CurrentHub().Scope().SetTag("service", service)
-	}
+// Prepares the interceptors (middleware) for the unary RPC endpoints of the server.
+// The first interceptor will be the outer-most handler and the last will be the
+// inner-most wrapper around the final handler. All unary interceptors returned by this
+// method should be changed using grpc.ChaingUnaryInterceptor().
```
DanielSollis, line 24: typos: 'chained' not 'changed', also 'grpc.ChainUnaryInterceptor()'

pkg/gds/interceptor.go
```diff
@@ -3,141 +3,93 @@ package gds
...
-	}()
-
-	// Check if we're in maintenance mode -- no streaming method should be available
-	if s.conf.Maintenance {
-		err = status.Error(codes.Unavailable, "the GDS service is currently in maintenance mode")
-		log.Trace().Err(err).Str("method", info.FullMethod).Msg("gds service unavailable during maintenance mode")
-
-		panicked = false
-		return err
+	// Return Unary interceptors
```
pdamodaran, line 64: Change to `Return Stream interceptors`

pkg/trtl/interceptor.go
```diff
@@ -1,216 +1,67 @@
...
+			interceptors.StreamRecovery(),
+			interceptors.StreamMaintenance(statusMethod),
 		}
-		return nil, fmt.Errorf("unexpected peer transport credentials type: %T", gp.AuthInfo)
 	}
 
-	chains = tlsAuth.State.VerifiedChains
-	if len(chains) == 0 {
-		return nil, errors.New("could not find certificate chain")
+	// Return Unary interceptors
```
pdamodaran, line 56: Change to `Return Stream interceptors`

pkg/utils/interceptors/logging.go
```diff
@@ -0,0 +1,119 @@
...
+func ParseMethod(method string) (string, string) {
+	method = strings.TrimPrefix(method, "/") // remove leading slash
+	if i := strings.Index(method, "/"); i >= 0 {
+		return method[:i], method[i+1:]
+	}
+	return "unknown", "unknown"
+}
+
+func StreamType(info *grpc.StreamServerInfo) string {
+	if !info.IsClientStream && !info.IsServerStream {
```
DanielSollis, line 109: Since this is only called in `StreamLogging`, which only accepts `grpc.ServerStreams`, do we really need to check here for unary RPC's?

pkg/utils/interceptors/mtls.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/rs/zerolog/log"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/peer"
+	"google.golang.org/grpc/status"
+)
+
+// UnaryMTLS adds authenticated peer info to the context and returns unauthenticated if
```
DanielSollis, line 19: might want to say `returns an 'unauthenticated' error` here

pkg/utils/interceptors/logging.go
bbengfort: I think we do - if we don't handle all cases of the different combination of bools then the default will be "bidirectional" and that's not the case when they're both false.

bbengfort:
@pdamodaran @DanielSollis I've made the changes you've requested and I believe I've fixed the tests as well.

PR #972, opened 2022-12-29
Sc-8652: Handle Expired Certificates in Certman

DanielSollis:
### Scope of changes

Updates the Certman reissuance loop to handle the case where the certificate has expired.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/gds/certman/certs.go
```diff
@@ -824,6 +835,10 @@ vaspsLoop:
 	}
 }
 
+func RetrieveCertID(identityCert *pb.Certificate) string {
+	return fmt.Sprintf("%X", identityCert.SerialNumber)
+}
```
pdeziel, line 840: I do like this helper, however I think we can make it a bit more useful by moving it to the `models` package (`pkg/models/v1`) so other packages can use it. Then, let's replace all instances where we are encoding the serial number with this helper method so that we can standardize that a bit more (basically just search the repo for `fmt.Sprintf("%X`).

Also since we're moving it to a different package can we change the name of the method to `GetCertID` or something along those lines to better match the naming conventions there?

pkg/gds/certman/certs.go
```diff
@@ -751,9 +752,19 @@ vaspsLoop:
...
+
+		// If a certificate has expired, update the certificate record.
+		case daysBeforeExpiration <= 0:
+			var cert *models.Certificate
+			if cert, err = c.db.RetrieveCert(RetrieveCertID(identityCert)); err != nil {
+				log.Error().Err(err).Str("vasp_id", vasp.Id).Msg("could not update expired certificate record status")
+				continue vaspsLoop
+			}
+			cert.Status = models.CertificateState_EXPIRED
+			c.db.UpdateCert(cert)
```
pdeziel, line 766: We need to handle any errors returned by `UpdateCert` here, my first instinct is that we don't need to bail out of the loop or anything but I think we should at least log the error. What do you think?

pkg/gds/certman/certs_test.go
```diff
@@ -276,9 +276,10 @@ func (s *certTestSuite) TestCertManagerSevenDayReissuanceReminder() {
 	require.NoError(err, "could not get hotel VASP")
 	hotelVASP = s.setupVASP(hotelVASP)
 
-	// Call the certman function at 6 days, which will send
-	// the seven day cert reissuance reminder to echoVASP.
+	// Call the certman function at 6 days and 1 day, which will send seven day
+	// cert reissuance reminders to charlieVASP and hotelVASP.
 	s.updateVaspIdentityCert(charlieVASP, 6)
+	s.updateVaspIdentityCert(hotelVASP, 1)
```
pdeziel, line 282: Was this test failing? I don't understand why this line is necessary now.

pkg/gds/certman/certs_test.go
```diff
@@ -312,6 +313,36 @@ func (s *certTestSuite) TestCertManagerSevenDayReissuanceReminder() {
 	emails.CheckEmails(s.T(), messages)
 }
 
+func (s *certTestSuite) TestCertManagerExpiration() {
+	s.setupCertManager(sectigo.ProfileCipherTraceEE, fixtures.Small)
+	defer s.teardownCertManager()
+	defer s.fixtures.LoadReferenceFixtures()
+	require := s.Require()
+
+	// setup the datastore to contain the modified charlieVASP
```
pdeziel, line 322:
```suggestion
	// Setup the datastore to contain the modified hotelVASP
```

pkg/gds/certman/certs_test.go
```diff
@@ -312,6 +313,36 @@ func (s *certTestSuite) TestCertManagerSevenDayReissuanceReminder() {
...
+	}
+	err = s.db.UpdateCert(cert)
+	require.NoError(err)
+
+	// Run the loop again to ensure that emails are not resent to contacts
+	s.certman.HandleCertificateReissuance()
+
+	cert, err = s.db.RetrieveCert(certID)
+	require.NoError(err)
+	require.Equal(cert.Status, models.CertificateState_EXPIRED)
```
pdeziel, line 340: So `hotelVASP` has an expired certificate by default?

pdeziel (commented):
Looking good, I did have a few suggestions before merging but they should be fairly straightforward.

pkg/gds/certman/certs_test.go
DanielSollis: Ya, hotelVASP's cert is expired so it was triggering the `<= 7 days before expiration` case by default, but now since we have a `<= 0 days before expiration` case it was triggering that and throwing off the expected emails and causing the test to fail, so I added this line to update it's cert expiration.

pkg/gds/certman/certs_test.go
DanielSollis: yep, see the above comment.

pkg/gds/certman/certs_test.go
DanielSollis: Good catch!

pkg/gds/certman/certs.go
DanielSollis: Fixed! I hadn't notice UpdateCert returning an error, my bad!

pkg/gds/certman/certs.go
DanielSollis: That's a good idea!

pkg/gds/certman/certs_test.go
pdeziel: That makes complete sense, thanks.

pkg/models/v1/models.go
```diff
@@ -504,6 +504,10 @@ func UpdateCertificateRequestStatus(request *CertificateRequest, state Certifica
 	return nil
 }
 
+func GetCertID(identityCert *pb.Certificate) string {
```
pdeziel, line 507: I think we should add a comment above this method to make it clear what it's doing, since it's not obvious from the name that it's returning the capital hex encoded string version of the serial number, and also why it's necessary (e.g. the certificate IDs are the string encoded version of the certificate serial number which match what is displayed by sectigo).

pdeziel (approved):
Thanks for making the changes, I had an additional suggestion to add a docstring but other than that feel free to merge when you're ready.

PR #976, opened 2023-01-05
DO NOT MERGE - Recipient migration

pdeziel:
### Scope of changes

This adds a trtl command to perform the recipient email address migration. It should retroactively add the contact's email address to the email log on each contact on each VASP if the recipient address is missing.

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/cmd/trtl/main.go
+++ b/cmd/trtl/main.go
@@ -16,14 +16,17 @@ import (
 	"github.com/syndtr/goleveldb/leveldb/opt"
 	"github.com/trisacrypto/directory/pkg"
 	profiles "github.com/trisacrypto/directory/pkg/gds/client"
+	"github.com/trisacrypto/directory/pkg/models/v1"
 	"github.com/trisacrypto/directory/pkg/trtl"
 	"github.com/trisacrypto/directory/pkg/trtl/config"
 	"github.com/trisacrypto/directory/pkg/trtl/pb/v1"
 	"github.com/trisacrypto/directory/pkg/trtl/peers/v1"
 	"github.com/trisacrypto/directory/pkg/utils/wire"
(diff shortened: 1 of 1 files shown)


cmd/trtl/main.go
```diff
@@ -558,6 +575,95 @@ func initPeersClient(c *cli.Context) (err error) {
...
+			}
+
+			if err = contact.Extra.UnmarshalTo(extra); err != nil {
+				return cli.Exit(err, 1)
+			}
+
+			// Iterate over the email log and populate the recipients
+			for _, email := range extra.EmailLog {
+				if email.Recipient == "" {
+					email.Recipient = contact.Email
```
pdeziel, line 632: Note that this might be inaccurate if the contact email was changed at some point (as older entries will contain the newer email address instead of the older address), we might have to fix that with another migration (or manually) if that ever happened.

rebeccabilbro: Interesting - this is certainly going to be true for several VASPs that have experienced staff turnover such as Anchorage, Apex Crypto, and probably a handful of others.  Ironically this would be a good use case for change data capture a la Ensign 😁  But I think we can live with slightly wrong data for some historical records, as the email log feature is primarily useful during the certificate issuance or reissuance process.

rebeccabilbro (approved):
This looks good to me @pdeziel 👍

cmd/trtl/main.go
pdeziel: Good to know, at least it will be accurate going forward.

PR #977, opened 2023-01-06
Sorted appdata organizations

pdeziel:
### Scope of changes

This modifies the `AddOrganization` and `RemoveOrganization` methods on the `AppMetadata` struct so that ordering is preserved in the list for the org IDs in a user's app metadata. This is primarily a performance improvement which is relevant to TSPs who exist in many organizations at once, since a binary search can be used (via the go `sort` package) rather than iterating through the entire list.

This change will require a migration to sort the organization IDs in Auth0 (currently they are unordered).

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [x] performance optimization

### Acceptance criteria

The existing methods should still operate as described in the docstring with the new sort code.

### Author checklist

### Reviewer(s) checklist

- [ ] The optimization makes sense
- [ ] There is enough test coverage

### DIFF
--- a/pkg/bff/auth/metadata.go
+++ b/pkg/bff/auth/metadata.go
@@ -2,6 +2,7 @@ package auth
 
 import (
 	"encoding/json"
+	"sort"
 
 	"github.com/trisacrypto/directory/pkg/bff/models/v1"
 )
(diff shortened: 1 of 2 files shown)


pkg/bff/auth/metadata.go
```diff
@@ -106,23 +107,41 @@ func (meta *AppMetadata) UpdateOrganization(org *models.Organization) {
...
+	copy(meta.Organizations[i:], meta.Organizations[i+1:])
+	meta.Organizations[len(meta.Organizations)-1] = ""
+	meta.Organizations = meta.Organizations[:len(meta.Organizations)-1]
+}
+
+// indexOf uses a binary search to return the index where the target string should be
+// inserted or found in the list. The list must already be sorted in ascending order,
+// otherwise this method will have undefined behavior.
+func indexOf(list []string, target string) int {
+	return sort.Search(len(list), func(i int) bool { return list[i] >= target })
```
pdeziel, line 146: This only works if the list is already sorted, but does reduce the lookup complexity from `O(n)` to `O(log n)` because we are able to use a binary search instead of a linear search.

pkg/bff/auth/metadata.go
```diff
@@ -106,23 +107,41 @@ func (meta *AppMetadata) UpdateOrganization(org *models.Organization) {
 // part of. This method is idempotent and will not add the organization ID if it
 // already exists.
 func (meta *AppMetadata) AddOrganization(orgID string) {
```
pdeziel, line 109: This method is called every time TSP users login to the UI (or switch organizations), so hopefully the login performance will not degrade as much with this change as TSPs continue to add organizations.

daniellemaxwell (approved):
The code comments and PR comments were very helpful. It was great to see how `sort` is used in Go, especially in comparison to JavaScript. Everything looks good!

PR #979, opened 2023-01-06
Add pagination to list orgs

pdeziel:
### Scope of changes

This adds pagination to the `ListOrganizations` BFF endpoint. It now accepts two query parameters to control the pagination: `page` and `page_size`, similar to how pagination is implemented for `ListVASPs` on the Admin API. This will allow the user UI frontend to load pages of organizations individually to avoid the list filling up the entire page.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Callers should be able to use the query params in the `ListOrganizations` request to control the paginated response.

### Author checklist

### Reviewer(s) checklist

- [ ] The pagination logic is correct
- [ ] There is enough test coverage

### DIFF
--- a/pkg/bff/docs/docs.go
+++ b/pkg/bff/docs/docs.go
@@ -486,11 +486,27 @@ const docTemplate = `{
                     "organizations"
                 ],
                 "summary": "List organizations [read:organizations]",
+                "parameters": [
+                    {
+                        "type": "integer",
+                        "default": 1,
+                        "description": "Page number",
+                        "name": "page",
+                        "in": "query"
+                    },
+                    {
+                        "type": "integer",
+                        "default": 8,
+                        "description": "Page size",
+                        "name": "page_size",
+                        "in": "query"
+                    }
+                ],
                 "responses": {
                     "200": {
                         "description": "OK",
                         "schema": {
-                            "type": "list"
+                            "$ref": "#/definitions/api.ListOrganizationsReply"
                         }
                     },
                     "401": {
@@ -1073,6 +1089,26 @@ const docTemplate = `{
                     "items": {
                         "$ref": "#/definitions/models.Collaborator"
                     }
+                }
+            }
+        },
+        "api.ListOrganizationsReply": {
+            "type": "object",
+            "properties": {
+                "count": {
+                    "type": "integer"
+                },
+                "organizations": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/api.OrganizationReply"
+                    }
+                },
+                "page": {
+                    "type": "integer"
+                },
+                "page_size": {
+                    "type": "integer"
                 }
             }
         },
--- a/pkg/bff/organization.go
+++ b/pkg/bff/organization.go
@@ -144,7 +144,9 @@ func (s *Server) CreateOrganization(c *gin.Context) {
 // @Description Return the list of organizations that the user is assigned to.
 // @Tags organizations
 // @Produce json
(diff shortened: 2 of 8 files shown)


pkg/bff/docs/docs.go
```diff
@@ -1076,6 +1092,26 @@ const docTemplate = `{
...
+                    }
+                },
+                "page": {
+                    "type": "integer"
+                },
+                "page_size": {
+                    "type": "integer"
+                }
+            }
+        },
```
pdeziel, line 1114: Autogenerated from the doc updates.

pkg/bff/organization.go
```diff
@@ -169,29 +171,64 @@ func (s *Server) ListOrganizations(c *gin.Context) {
...
+			out.Organizations = append(out.Organizations, &api.OrganizationReply{
 				ID:        org.Id,
 				Name:      org.ResolveName(),
 				Domain:    org.Domain,
 				CreatedAt: org.Created,
 				LastLogin: collaborator.LastLogin,
 			})
 		}
+
+		out.Count++
```
pdeziel, line 231: Note that `Count` refers to the total count of orgs across all pages.

pkg/bff/organization.go
```diff
@@ -169,29 +171,64 @@ func (s *Server) ListOrganizations(c *gin.Context) {
...
+	// Parse the params from the GET request
+	params := &api.ListOrganizationsParams{}
+	if err = c.ShouldBindQuery(params); err != nil {
+		log.Warn().Err(err).Msg("could not bind request with query params")
+		c.JSON(http.StatusBadRequest, api.ErrorResponse(err))
+		return
+	}
+
+	// Set pagination defaults
+	if params.Page <= 0 {
```
DanielSollis, line 183: should this return an error instead of switching to the default?

pkg/bff/organization.go
```diff
@@ -169,29 +171,64 @@ func (s *Server) ListOrganizations(c *gin.Context) {
...
+		// Make sure the user is a collaborator in the organization
+		// Note: The UserInfo middleware ensures that the user email address is
+		// present in the claims so we can safely dereference it here.
+		var collaborator *models.Collaborator
+		if collaborator = org.GetCollaborator(*user.Email); collaborator == nil {
+			log.Error().Str("org_id", id).Str("email", *user.Email).Msg("could not find user in organization collaborators")
+			continue
+		}
+
+		if out.Count >= minIndex && out.Count < maxIndex {
```
DanielSollis, line 220: Am I right in thinking that for each call to `ListOrganizations` you're iterating through all organizations? Wouldn't it be more efficient to only iterate through the number of organizations in `params.PageSize`?

DanielSollis (approved):

pkg/bff/organization.go
pdeziel: You're right the endpoint would be faster. We could design the pagination so that we only iterate up to the page limit as you suggest, the advantage of this approach is that returning the total count allows the client to compute how many pages there are for frontend purposes.

pkg/bff/organization.go
pdeziel: We could, in my opinion it increases the complexity of the design without providing much benefit. Recently I've been trying to follow the principle of minimizing the number of errors that we return to the client. In this case we can assume a default value so it's my preference to not generate an error out of nothing but I can also see the case where generating a 400 is perfectly valid.

PR #981, opened 2023-01-09
bffutil: Appdata sort orgs command

pdeziel:
### Scope of changes

This adds a command to `bffutil` to sort organizations on the user's appdata record in Auth0. We recently changed the way organizations are added and removed from user Auth0 records in sc-11172 so that the list is always sorted. However, we need to fix this retroactively for existing users in order for the implemented sorting logic to work.

This command can be run outside of the pod since it only needs to interact with Auth0 and it sorts every user's current org list and pushes the appdata with the sorted list to Auth0.

SC-12601

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [x] bffutil command

### Acceptance criteria

This command correctly makes the Auth0 updates and is idempotent (it can be run multiple times and produces the same result).

### Author checklist

### Reviewer(s) checklist

- [ ] Is it okay to require a valid configuration with all of the required BFF fields (even though the command doesn't touch the database)?
- [ ] The command output is understandable

### DIFF
--- a/cmd/bffutil/main.go
+++ b/cmd/bffutil/main.go
@@ -159,6 +159,19 @@ func main() {
 				},
 			},
 		},
+		{
+			Name:   "appdata:sortorgs",
+			Usage:  "sort the organization list on each user's app_metadata",
+			Action: sortAppdataOrgs,
+			Before: Before(loadConf, connectAuth0),
+			Flags: []cli.Flag{
+				&cli.BoolFlag{
+					Name:    "dry-run",
+					Aliases: []string{"d"},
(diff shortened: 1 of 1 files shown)


rebeccabilbro (approved):
Looks good to me @pdeziel! Do we need to create a follow on task to execute the resorting for existing records?

pdeziel:
> Looks good to me @pdeziel! Do we need to create a follow on task to execute the resorting for existing records?

We do, I was planning on doing that the next time we deploy to prod.

PR #983, opened 2023-01-11
BFF delete organization

pdeziel:
### Scope of changes

This adds a delete organization endpoint to the BFF API. The tricky part here is making sure users in that organization can still login to the frontend. For example, if a user's current organization is the one being deleted, we need to update their app metadata on Auth0 so that they default to a different organization (if available).

SC-11817

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

This endpoint should allow users with the `delete:organizations` permission to delete organizations that they are a collaborator on.

### Author checklist

### Reviewer(s) checklist

- [ ] The organization switching logic makes sense
- [ ] There is sufficient test coverage

### DIFF
--- a/pkg/bff/docs/swagger.yaml
+++ b/pkg/bff/docs/swagger.yaml
@@ -59,6 +59,19 @@ definitions:
           $ref: '#/definitions/models.Collaborator'
         type: array
     type: object
+  api.ListOrganizationsReply:
+    properties:
+      count:
+        type: integer
+      organizations:
+        items:
+          $ref: '#/definitions/api.OrganizationReply'
+        type: array
+      page:
+        type: integer
+      page_size:
+        type: integer
+    type: object
   api.LoginParams:
     properties:
       orgid:
@@ -596,13 +609,24 @@ paths:
   /organizations:
     get:
       description: Return the list of organizations that the user is assigned to.
+      parameters:
+      - default: 1
+        description: Page number
+        in: query
+        name: page
+        type: integer
+      - default: 8
+        description: Page size
+        in: query
+        name: page_size
+        type: integer
       produces:
       - application/json
       responses:
         "200":
           description: OK
           schema:
-            type: list
+            $ref: '#/definitions/api.ListOrganizationsReply'
         "401":
           description: Unauthorized
           schema:
@@ -654,8 +678,9 @@ paths:
       - organizations
   /organizations/{id}:
     delete:
-      description: Completely delete an organization, including the registration and
-        collaborators.
+      description: Permanently delete an organization, including the registration
+        and collaborators. This action is irreversible so the frontend should obtain
+        confirmation from the user before calling this endpoint.
       responses:
         "200":
           description: OK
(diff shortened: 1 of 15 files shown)


pkg/bff/organization.go
```diff
@@ -197,6 +197,97 @@ func (s *Server) ListOrganizations(c *gin.Context) {
...
+
+	// The user must be a collaborator in the organization to delete it
+	if org.GetCollaborator(*user.Email) == nil {
+		log.Error().Err(err).Str("org_id", orgID).Str("email", *user.Email).Msg("could not find user in organization collaborators")
+		c.JSON(http.StatusForbidden, api.ErrorResponse("user is not authorized to access this organization"))
+		return
+	}
+
+	// Delete the organization from the database
+	if err = s.db.DeleteOrganization(org.UUID()); err != nil {
```
pdeziel, line 245: Note that we will need at least a confirmation prompt on the frontend, since deleting the organization record will completely delete the registration from the database.

pkg/bff/organization.go
```diff
@@ -197,6 +197,97 @@ func (s *Server) ListOrganizations(c *gin.Context) {
...
+		log.Error().Err(err).Str("org_id", orgID).Msg("could not delete organization from database")
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not delete organization from database"))
+		return
+	}
+
+	// Remove the organization from all collaborators so that they don't get an error
+	// when they try to log in. If a user no longer has an organization, they will be
+	// assigned one automatically the next time they log in.
+	for _, collab := range org.GetCollaborators() {
```
pdeziel, line 254: This is pretty tricky because we're relying on state in Auth0 to decide which organization to attempt to log the user into. If the organization is deleted out from under them, we need to have a way in which they can still login. For most users, what this means is that they will be assigned a new default organization on login. For TSP users who can be assigned to multiple organizations at once, we need to do a bit of extra work to figure out which organization to default them to.

rebeccabilbro: If GDS went down in the middle of this operation, could we end up with some collaborators who still have metadata for deleted orgs? If so, are there any places where you think this might cause problems?

pkg/bff/docs/swagger.yaml
```diff
@@ -652,6 +652,34 @@ paths:
       summary: Create a new organization [create:organizations]
       tags:
       - organizations
+  /organizations/{id}:
+    delete:
+      description: Completely delete an organization, including the registration and
```
rebeccabilbro, line 657: Should we add a little note like "Warning! Deletes are not reversible! Be sure to prompt user for confirmation before executing."

rebeccabilbro (approved):
Looks good @pdeziel ! I had just a few questions (probably resulting from the fact that I am not very familiar with the collaborators feature).

pkg/bff/organization.go
```diff
@@ -197,6 +197,97 @@ func (s *Server) ListOrganizations(c *gin.Context) {
...
+	orgID := c.Param("orgID")
+
+	// Fetch the organization to be deleted
+	if org, err = s.OrganizationFromID(orgID); err != nil {
+		log.Error().Err(err).Str("org_id", orgID).Msg("could not retrieve organization from database")
+		c.JSON(http.StatusNotFound, api.ErrorResponse("organization not found"))
+		return
+	}
+
+	// The user must be a collaborator in the organization to delete it
```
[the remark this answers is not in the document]
pdeziel: The user will need the `delete:organizations` permission to get past the middleware to this endpoint. This is mainly a sanity check to make sure the user is actually in the organization (e.g. to protect against the frontend passing in the wrong orgID).

pkg/bff/organization.go
pdeziel: I think this would mostly cause problems when users try to login; in that case we might have to manually change things in Auth0 to make sure everything is in sync.

PR #987, opened 2023-02-28
VASP Country Code Fix

bbengfort:
### Scope of changes

Fixes SC-12213 - a small bug where the VASP country code was missing. 

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

This should be a quick and simple review!

### DIFF
--- a/pkg/gds/admin.go
+++ b/pkg/gds/admin.go
@@ -782,15 +782,16 @@ func (s *Admin) ListCountries(c *gin.Context) {
 
 		// Prevent nil panics -- if this happens then the VASP is in an invalid state.
 		if vasp.Entity == nil {
-			log.Warn().Str("vasp_id", vasp.Id).Msg("VASP entity is nil")
+			log.Warn().Str("vasp_id", vasp.Id).Msg("vasp entity is nil")
 			continue
 		}
 
 		// Counts are aggregated by country ISO code
 		var code string
 		if code = vasp.Entity.CountryOfRegistration; code == "" {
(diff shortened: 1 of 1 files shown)


pkg/gds/admin.go
```diff
@@ -780,10 +780,17 @@ func (s *Admin) ListCountries(c *gin.Context) {
...
+		if vasp.Entity == nil {
+			log.Warn().Str("vasp_id", vasp.Id).Msg("VASP entity is nil")
+			continue
+		}
+
 		// Counts are aggregated by country ISO code
 		var code string
 		if code = vasp.Entity.CountryOfRegistration; code == "" {
-			log.Error().Str("vasp_id", vasp.Id).Msg("VASP country code is empty")
+			// TODO: is there a valid reason why the country code might be empty?
```
pdeziel, line 792: This implies that an empty country code might be valid? https://github.com/trisacrypto/trisa/blob/cc5ca7a6bc1b72ec84ca374ef37769f0cf5a9fbb/pkg/ivms101/validate.go#L137-L146. Does the UI currently allow the field to be empty?

pdeziel (approved):
Looks good to me!

pkg/gds/admin.go
bbengfort: Ah good find - I don't know about the UI, but I think the validate code is enough to reduce this to a `Debug()` instead of a warn and remove the TODO (replacing with a note). Thanks, I'll make the change!

PR #1004, opened 2023-04-13
Sc-14487: Implement leveldb DeleteContact

DanielSollis:
### Scope of changes

Implements the DeleteContact method in store/leveldb

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pdeziel (commented):
Looks good, my comments here are pretty similar to the trtl PR so could you just make the same changes I requested there but in these methods?

pkg/models/v1/models.go
```diff
@@ -223,10 +223,10 @@ func GetVASPEmailLog(vasp *pb.VASP) (emails []*EmailLogEntry, err error) {
 }
 
 // Normalize the email and convert to bytes
-func ContactKey(email string) []byte {
+func NormalizeEmail(email string) string {
```
pdeziel, line 226: Is there a reason why this wouldn't be the same method as in the trtl store?

pdeziel (commented):
Looks good, I just had a comment about the helper method.

pkg/models/v1/models.go
DanielSollis: The leveldb store requires a string to add a byte prefix to using the makeKey method with the preContacts string as an argument, while the trtl requires a byte key, I pulled out the common code into NormalizeEmail but it turns out that the last step is unfortunately different for each store.

pdeziel: Gotcha, I definitely like having `NormalizeEmail` as a shared method. The only suggestion I would have is to move the call to `NormalizeEmail` to the `contactKey` method, which would simplify the code in the store methods.

DanielSollis: That's a really good idea! I just implemented that and it's much cleaner.

PR #1007, opened 2023-05-03
initial work

DanielSollis:
### Scope of changes
Creates a new Contacts package and tests containing helper functions for the models.contact message.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/gds/contacts/contacts.go
pdeziel, line 1: To be honest we might not need most of these helpers now (which is a good thing) - I think the reason for most of them is because we needed to have GDS-specific data on the TRISA protocol buffers without having to modify the originals in the TRISA repo. That involved a lot of marshaling/unmarshaling hence the helpers.

An exception is the `AppendEmailLog`, we may still want that one since it allows the caller to not worry about if the log is nil or not.

pkg/gds/contacts/contacts.go
```diff
@@ -0,0 +1,75 @@
...
+func GetEmailLog(contact *models.Contact) (_ []*models.EmailLogEntry) {
+	// If the extra data is nil, return nil (no email log).
+	if contact == nil {
+		return nil
+	}
+	return contact.EmailLog
+}
+
+// Create and add a new entry to the EmailLog on the extra data on the Contact record.
+func AppendEmailLog(contact *models.Contact, reason, subject string) (err error) {
```
pdeziel, line 55: Since this now only returns an error if the contact is nil, I'm wondering if we should just make this a method on `models.Contact`? The way I'm picturing us using this is:

```golang
// Retrieve contact from the database
var contact *models.Contact
if contact, err = db.RetrieveContact(email); err != nil {
    return err
}

...

// Append an entry to the log (in some method that sends an email)
contact.AppendEmailLog(reason, subject)
```

`contact` won't be nil because now we only get it from a database lookup, and we should be returning an error instead. Before we had to worry about if it was nil because it was a field on a VASP.

So as far as I can tell, we don't need to return an error from `AppendEmailLog` anymore, what do you think?

pdeziel (commented):
Definitely appreciate the tests! After looking at this, I realized that I don't think most of the helpers are especially helpful anymore (since we don't have the extra data anymore and the contact is stored by itself), which is great because I think we can simplify things a lot.

pdeziel (approved):
Looks good, were you going to add the `AppendEmailLog` helper to a different PR?

DanielSollis:
It's in this PR, I added it to models.pb.go

PR #1031, opened 2023-05-26
Update registration tests

pdeziel:
### Scope of changes

This adds some test coverage for update registration.

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [x] testing

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


pkg/bff/models/v1/registration_test.go
```diff
@@ -576,6 +576,125 @@ func TestValidateTRISA(t *testing.T) {
...
+	update := &RegistrationForm{}
+	err := loadJSONFixture("testdata/registration_form.json", update)
+	require.NoError(t, err, "error loading registration form fixture")
+
+	// An error should be returned if an unknown step is provided
+	err = form.Update(update, "invalid")
+	require.EqualError(t, err, "unknown step \"invalid\"", "error should be returned for unknown step")
+
+	// Test updating the basic details step
+	update.State.Current = 1
```
bbengfort, line 602: Is there an error if the state is not on the correct step? 

Also - I think we should be returning the step name in the state, but we haven't been doing that; is there a story to remediate that?

bbengfort (approved):

pkg/bff/models/v1/registration_test.go
pdeziel: We currently are not checking the state on the backend side, should we be?

bbengfort: No we shouldn't be - I was just wondering why the tests were setting current to the correct step. But if it's what the front-end is doing then it's probably good to have in our tests. 

As for the step names; the backend is currently the only thing that knows those names -- but we can leave state entirely to the front-end if that makes more sense.

pdeziel: I was just setting it to something to ensure that the state is being updated on the form, but we have been relying on the frontend to manage when the state is changed.

bbengfort: Ok, sounds good

PR #1080, opened 2023-07-11
Fix the switch request & export member list issue

masskoder:
### Scope of changes

This fix will allow users to switch between mainnet and testnet, and they can also export member lists.

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

 - after each network selects the request payload should update 
 
 
image

### Author checklist

### Reviewer(s) checklist


web/gds-user-ui/src/modules/dashboard/member/utils.ts
```diff
@@ -3,11 +3,26 @@ import { convertToCVS, downloadCSV } from 'utils/utils';
...
+  {
+    key: 'last_updated',
+    label: t`Last Updated`
+  },
+  {
+    key: 'network',
+    label: t`Network`
+  },
+  {
+    key: 'status',
```
daniellemaxwell, line 23: The BE returns a number for the verification state. In a separate PR, I plan on creating a util that will get the status based on what's provided from the BE. Here, we'll have to make sure that `VERIFIED` is displayed or users will be confused if they see a number. Here's a link to where the VerificationState is coming from: https://github.com/trisacrypto/trisa/blob/0b895ae982a8162b729f67e54b36e619ef39b8bf/pkg/trisa/gds/models/v1beta1/models.pb.go#L82C1-L82C1

web/gds-user-ui/src/modules/dashboard/member/memberType.ts
```diff
@@ -22,10 +22,10 @@ export enum DirectoryTypeEnum {
   MAINNET = 'mainnet'
 }
 export enum VaspDirectoryEnum {
-  TESTNET = 'testnet.net',
+  TESTNET = 'trisatest.net',
```
daniellemaxwell, line 25: Good catch!

daniellemaxwell (approved):
The changes for the switch component look good. When exporting mock data, the issue with data not appearing in the spreadsheet didn't appear to be resolved. However, using a key is a good idea. If you could take a look into that and also make sure that the correct status is returned, then this should be good to merge.

PR #1096, opened 2023-07-13
Adds temp switch for business category values

daniellemaxwell:
### Scope of changes

Displays the `Business Category` as a text value and not a number.

Fixes SC-18847

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

Screenshot 2023-07-13 at 2 58 48 PM

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/web/gds-user-ui/src/modules/dashboard/member/utils.ts
+++ b/web/gds-user-ui/src/modules/dashboard/member/utils.ts
@@ -39,25 +39,25 @@ export const getVaspNetwork = (dir: any) => {
   }
 };
 
-export const getVaspStatus = (status: number) => {
+export const getVaspStatus = (status: any) => {
   switch (status) {
-    case 1:
+    case 1 || 'NO_VERIFICATION':
       return t`NO VERIFICATION`;
-    case 2:
+    case 2 || 'SUBMITTED':
       return t`SUBMITTED`;
-    case 3:
+    case 3 || 'EMAIL_VERIFIED':
       return t`EMAIL VERIFIED`;
-    case 4:
+    case 4 || 'PENDING_REVIEW':
       return t`PENDING REVIEW`;
-    case 5:
+    case 5 || 'REVIEWED':
       return t`REVIEWED`;
-    case 6:
+    case 6 || 'VERIFIED':
       return t`VERIFIED`;
-    case 7:
+    case 7 || 'REJECTED':
       return t`REJECTED`;
-    case 8:
+    case 8 || 'APPEALED':
       return t`APPEALED`;
-    case 9:
+    case 9 || 'ERRORED':
       return t`ERRORED`;
     default:
       return t`NO VERIFICATION`;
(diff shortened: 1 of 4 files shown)


web/gds-user-ui/src/modules/dashboard/member/utils.ts
```diff
@@ -87,3 +87,18 @@ export async function copyToClipboard(data: any) {
       console.error('[copyToClipboard]', err);
   }
 }
+
+export const getBusinessCategory = (category: any) => {
```
daniellemaxwell, line 91: This is a temporary fix for now because the BE is returning a number instead of a string. Here's a link to how the Business Categories are listed on the BE: https://github.com/trisacrypto/trisa/blob/0b895ae982a8162b729f67e54b36e619ef39b8bf/pkg/trisa/gds/models/v1beta1/models.pb.go#L28C1-L28C1

Once a fix is made to return strings, this will be removed. Also, I added the strings following the discussion during QA.

web/gds-user-ui/src/modules/dashboard/member/utils.ts
```diff
@@ -39,25 +39,25 @@ export const getVaspNetwork = (dir: any) => {
   }
 };
 
-export const getVaspStatus = (status: number) => {
+export const getVaspStatus = (status: any) => {
   switch (status) {
-    case 1:
+    case 1 || 'NO_VERIFICATION':
```
daniellemaxwell, line 44: I added strings here to be consistent with `getBusinessCategory` below. Once the BE fix is made, this can be updated.

web/gds-user-ui/src/modules/dashboard/member/utils.ts
```diff
@@ -87,3 +87,18 @@ export async function copyToClipboard(data: any) {
...
+    case 2 || 'GOVERNMENT_ENTITY':
+      return t`Government Entity`;
+    case 3 || 'BUSINESS_ENTITY':
+      return t`Business Entity`;
+    case 4 || 'NON_COMMERCIAL_ENTITY':
+      return t`Non-Commercial Entity`;
+    default:
+      return t`Unknown Entity`;
+  }
+};
```
masskoder, line 104:
```suggestion
export const getBusinessCategory = (category: any) => {
  switch (category) {
    case 1:
    case 'PRIVATE_ORGANIZATION':
      return t`Private Organization`;
    case 2:
    case 'GOVERNMENT_ENTITY':
      return t`Government Entity`;
    case 3:
    case 'BUSINESS_ENTITY':
      return t`Business Entity`;
    case 4:
    case 'NON_COMMERCIAL_ENTITY':
      return t`Non-Commercial Entity`;
    default:
      return t`Unknown Entity`;
  }
};
```

This ensures that both the numeric value and the corresponding string representation will be handled correctly.

masskoder (approved):
Looks good, but wouldn't it be better to handle only the data that will be returned and wait for the backend changes before merging it? In this way, we can avoid creating separate stories for a future update to remove the numeric values.

PR #1100, opened 2023-08-17
VASP names endpoint

pdeziel:
### Scope of changes

This adds an endpoint to the BFF to support autocomplete on search - by returning the list of deduplicated VASPs in testnet and mainnet.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

The endpoint should return VASP names from both testnet and mainnet but not return any VASPs that are unverified.

### Author checklist

### Reviewer(s) checklist


pkg/bff/server.go
```diff
@@ -363,6 +363,7 @@ func (s *Server) setupRoutes() (err error) {
 
 		// GDS public routes (no authentication required)
 		v1.GET("/lookup", s.Lookup)
+		v1.GET("/vasps", s.VASPNames)
```
bbengfort, line 366: I wonder if we should add this at `/lookup/autocomplete` instead -- since that's more specific to what we're using this list for. It is possible that we may need a `/vasps` endpoint in the future for a read-only REST endpoint with List and Detail.

pkg/bff/api/v1/api.go
```diff
@@ -18,6 +18,7 @@ type BFFClient interface {
 	// Unauthenticated Endpoints
 	Status(context.Context, *StatusParams) (*StatusReply, error)
 	Lookup(context.Context, *LookupParams) (*LookupReply, error)
+	VASPNames(context.Context) ([]string, error)
```
bbengfort, line 21: I think we should call this `LookupAutocomplete` instead -- see below comment.

pkg/bff/gds.go
```diff
@@ -155,6 +155,87 @@ func (s *Server) Lookup(c *gin.Context) {
...
+			var vasp *pb.VASP
+			if vasp, err = iter.VASP(); err != nil {
+				sentry.Error(c).Err(err).Str("network", network).Msg("could not get VASP from ListVASP iterator")
+				return nil, err
+			}
+
+			// Filter any VASPs that are not verified
+			if vasp.VerificationStatus != pb.VerificationState_VERIFIED {
+				continue
+			}
```
bbengfort, line 187: Thank you for adding this!

pkg/bff/gds.go
```diff
@@ -155,6 +155,87 @@ func (s *Server) Lookup(c *gin.Context) {
...
+		iter := db.ListVASPs(ctx)
+		defer iter.Release()
+
+		var names []string
+		for iter.Next() {
+			var vasp *pb.VASP
+			if vasp, err = iter.VASP(); err != nil {
+				sentry.Error(c).Err(err).Str("network", network).Msg("could not get VASP from ListVASP iterator")
+				return nil, err
+			}
```
bbengfort, line 182: If we can't parse one VASP record it doesn't make sense to blow up the whole return; instead we'll get the Sentry error and try to fix the corrupted protobuf record specifically for that key:

```suggestion
			if vasp, err = iter.VASP(); err != nil {
				sentry.Error(c).Err(err).Str("network", network).Bytes("key", iter.Key()).Msg("could not get VASP from ListVASP iterator")
				continue
			}
```

pkg/bff/gds.go
```diff
@@ -155,6 +155,87 @@ func (s *Server) Lookup(c *gin.Context) {
...
+			// Filter any VASPs that are not verified
+			if vasp.VerificationStatus != pb.VerificationState_VERIFIED {
+				continue
+			}
+
+			var name string
+			if name, err = vasp.Name(); err != nil {
+				sentry.Warn(c).Err(err).Str("network", network).Msg("could not resolve VASP name from VASP")
+				continue
+			}
```
bbengfort, line 193:
```suggestion
			if name, err = vasp.Name(); err != nil || name == "" {
				sentry.Warn(c).Err(err).Str("network", network).Bytes("key", iter.Key()).Msg("could not resolve VASP name from VASP")
				continue
			}
```

pkg/bff/gds.go
```diff
@@ -155,6 +155,87 @@ func (s *Server) Lookup(c *gin.Context) {
...
+	// Execute the parallel GDS list request
+	results, errs := s.ParallelDBRequests(sentry.RequestContext(c), rpc, true)
+	if len(errs) > 0 {
+		c.JSON(http.StatusInternalServerError, api.ErrorResponse("could not retrieve VASP names"))
+		return
+	}
+
+	// Deduplicate the results for the response
+	out := make([]string, 0)
+	names := make(map[string]struct{})
```
bbengfort, line 215: Nice!

bbengfort (approved):
I had a couple of fixes and comments, but once those are done - this looks good to go!

pkg/bff/server.go
pdeziel: Sure, makes sense to me.

PR #1115, opened 2023-08-24
publish network activity from gds:

pdamodaran:
### Scope of changes

This PR resolves sc-20305 which requested sending network activity events to the network activity handler.  This PR also adds a RegisterActivity to `event.go` to track registration events. 

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

### Author checklist

- [] I have manually tested the change and/or added automation in the form of unit tests or integration tests

### Reviewer(s) checklist


pkg/gds/gds.go
```diff
@@ -112,6 +114,9 @@ func (s *GDS) Shutdown() (err error) {
 // Register generates a PKCS12 password, provided in the RPC response which can be
 // used to access the certificate private keys when they're emailed.
 func (s *GDS) Register(ctx context.Context, in *api.RegisterRequest) (out *api.RegisterReply, err error) {
+	// send register request activity to network activity handler
+	go activity.Register().Add()
```
pdeziel, line 118: It shouldn't be necessary to run this as a go routine anymore, what the `Add()` does is just writes the activity to a channel. That should now be a non-blocking thing now because I've modified the method to just drop the event if the channel buffer is full.

pkg/gds/gds.go
```diff
@@ -412,6 +417,13 @@ func (s *GDS) Lookup(ctx context.Context, in *api.LookupRequest) (out *api.Looku
 		return nil, status.Error(codes.NotFound, "no VASP record available")
 	}
 
+	// send lookup request activity to network activity handler
+	if vasp != nil {
+		go activity.Lookup().VASP(uuid.MustParse(vasp.Id)).Add()
```
pdeziel, line 422: I believe we should always have a non-nil vasp here, if it was nil then line 416 would have panicked. Also if you want to change the `VASP()` method to accept a string rather than a uuid that would be fine with me. That way we wouldn't have to worry about the panic from `MustParse` here, although it means we have to be more careful about making sure what we pass in is actually a VASP ID.

pkg/utils/activity/event.go
```diff
@@ -86,6 +87,8 @@ func (a Activity) String() string {
 		return "lookup"
 	case SearchActivity:
 		return "search"
+	case RegisterActivity:
+		return "register"
```
pdeziel, line 91: Thanks for adding this!

pdeziel (commented):
I had some small comments that are hopefully not too complicated to address, but otherwise looks good!

pkg/gds/gds.go
pdamodaran: got it - I have made the change

pdeziel (approved):
Thanks for making the changes!

PR #1118, opened 2023-08-25
Activity store and subscriber

pdeziel:
### Scope of changes

This adds models to the store for persisting activity data, which for right now is designed specifically for time series retrieval. e.g. to retrieve activity data for a specific day, the caller would have to retrieve the month's worth of data from the document in trtl, and then seek to the desired day. However, this reduces the number of calls that have to be made to trtl to retrieve the data for the activity plot.

This PR also adds a subscriber to the BFF for asynchronously updating the activity data in the trtl after receiving activity events from Ensign.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist

- [ ] Please review the activity model logic in particular


pkg/store/trtl/trtl.go
```diff
@@ -720,6 +720,111 @@ func (s *Store) DeleteAnnouncementMonth(ctx context.Context, date string) (err e
...
+
+	if err = proto.Unmarshal(reply.Value, m); err != nil {
+		return nil, err
+	}
+	return m, nil
+}
+
+// UpdateActivityMonth creates a new activity month record if it doesn't already
+// exist or replaces the existing record.
+func (s *Store) UpdateActivityMonth(ctx context.Context, m *bff.ActivityMonth) (err error) {
```
bbengfort, line 764: Ok - so just so we're on the same page; there is a race condition in Retrieve than Update patterns; e.g. if go routine A retrieves, then routine B retrieves, then B updates, then A updates; B's updates will be lost. This shouldn't be a problem given there is only a single go routine updating the store from the BFF (e.g. the activity subscriber); but I wanted to mention it.

pkg/store/mockdb/mockdb.go
```diff
@@ -46,6 +46,9 @@ type MockState struct {
 	RetrieveAnnouncementMonthInvoked bool
 	UpdateAnnouncementMonthInvoked   bool
 	DeleteAnnouncementMonthInvoked   bool
+	RetrieveActivityMonthInvoked     bool
+	UpdateActivityMonthInvoked       bool
+	DeleteActivityMonthInvoked       bool
```
bbengfort, line 51: Huh, I didn't realize that the mockdb kept track of all these bools - that must have been a long time ago. I much prefer our call counter pattern that we're using elsewhere!

pkg/store/leveldb/leveldb.go
```diff
@@ -609,6 +609,87 @@ func (s *Store) DeleteAnnouncementMonth(ctx context.Context, date string) (err e
 	return nil
 }
 
+//===========================================================================
+// ActivityStore Implementation
+//===========================================================================
+
+// RetrieveActivityMonth returns the activity month record for the given date
+// timestamp in the format YYYY-MM.
+func (s *Store) RetrieveActivityMonth(ctx context.Context, date string) (m *bff.ActivityMonth, err error) {
```
bbengfort, line 618: Thank you for implementing both the trtl and leveldb stores!

pkg/bff/models/v1/activity.go
bbengfort, line 1: Good helpers!

pkg/bff/activity.go
```diff
@@ -0,0 +1,161 @@
...
+
+// ActivitySubscriber is a struct with a go routine that subscribes to a network
+// activity topic in Ensign and applies asynchronous updates to trtl.
+type ActivitySubscriber struct {
+	client *ensign.Client
+	db     store.Store
+	emock  *mock.Ensign
+	topic  string
+	stop   chan struct{}
+}
```
bbengfort, line 28: It makes sense to put this in at the BFF top level because of the requirement to access the `store.Store`. I did just want to suggest another pattern for future reference. You could have the subscriber in the `activity` package, and it does all the work of connecting to ensign and subscribing to the topic since the configuration is already in that package -- it can also handle event parsing and sending acks/nacks back. The subscriber takes as input an interface:

```golang
type ActivitySubscriber interface {
     HandleActivity(*NetworkActivity) error 
}
```

`HandleActivity()` is kind of like a callback function (another pattern that could be implemented) - and it is called on each new activity received from the subscriber; if error then a nack is sent back otherwise an ack is sent back.

bbengfort (approved):
Looks good; I understand the motivation to collect activity days in a month object; it is very similar to the announcements collection and at most only 31 days of activity will be returned so it should be too high volume data, and certainly not unbounded.

pkg/utils/activity/publisher.go
```diff
@@ -93,6 +93,10 @@ func Publish() {
 				continue
 			}
 
+			if _, err = client.Subscribe(topic); err != nil {
+				log.Error().Err(err).Msg("could not subscribe to activity topic")
+			}
```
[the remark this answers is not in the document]
pdeziel: My bad, I was tracking down a panic in the Go SDK but it turned out to be a test issue, that's why there's some weird stuff and comments scattered through the code.

pkg/bff/activity.go
pdeziel: Yeah that might make more sense, I'll keep it in mind.

pkg/store/trtl/trtl.go
pdeziel: Definitely, thanks for pointing that out.

PR #1130, opened 2023-11-06
Update models with webhook info

pdeziel:
### Scope of changes

This adds certificate webhook info the VASP and certificate request models and updates the reissuance CLI command to provide an override option for webhook certificate delivery.

### Type of change

- [ ] bug fix
- [x] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

This involves updating the dependencies so we should make sure that nothing weird happened there.

### Author checklist

### Reviewer(s) checklist


cmd/reissuer/main.go
```diff
@@ -414,38 +434,60 @@ func reissueCerts(c *cli.Context) (err error) {
...
+		// Store the password with the webhook
+		req := &api.StorePasswordRequest{
+			ID:       certreq.Id,
+			Password: pkcs12password,
+		}
+		if err = client.StoreCertificatePassword(ctx, req); err != nil {
+			return cli.Exit(fmt.Errorf("could not store password with webhook: %s", err), 1)
+		}
+
+		fmt.Printf("successfully sent PKCS12 password to webhook %q\n", certreq.Webhook)
```
bbengfort, line 489: Would prefer the verb "send" or "deliver" rather than "store" -- I think this is what might have created the confusion with Google Secret manager above. 

```suggestion
		// Create a courier client to deliver the pkcs12 password to the TRISA member
		var client api.CourierClient
		if client, err = api.New(certreq.Webhook); err != nil {
			return cli.Exit(fmt.Errorf("could not create courier client: %s", err), 1)
		}

		// Deliver the pkcs12 password via the webhook
		req := &api.StorePasswordRequest{
			ID:       certreq.Id,
			Password: pkcs12password,
		}
		if err = client.StoreCertificatePassword(ctx, req); err != nil {
			return cli.Exit(fmt.Errorf("could not deliver pkcs12 password with webhook: %s", err), 1)
		}

		fmt.Printf("successfully sent PKCS12 password to webhook %q\n", certreq.Webhook)
```

proto/gds/models/v1/models.proto
```diff
@@ -122,6 +128,12 @@ message GDSExtraData {
 
     // Log which records emails sent to the TRISA admins regarding this VASP
     repeated EmailLogEntry email_log = 6;
+
+    // Optional webhook for certificate delivery
+    string certificate_webhook = 7;
+
+    // Don't deliver certificates by email
+    bool no_email_delivery = 8;
```
bbengfort, line 136: While it does make sense to put it on the extra data field; I would like to put it on the `VASP` record directly, since I think this is something we want to have the users be able to manage (e.g. they should be able to manage their delivery preferences). E.g. this is a TRISA-level field, not just something that is specific to the GDS.

pkg/models/v1/models.go
```diff
@@ -508,6 +508,11 @@ func NewCertificateRequest(vasp *pb.VASP) (certRequest *CertificateRequest, err
 		log.Debug().Str("vasp_id", vasp.Id).Str("certreq_id", certRequest.Id).Msg("location information not found or incomplete")
 	}
 
+	// Update certificate delivery method with the method on the VASP
+	if err = UpdateCertificateRequestDelivery(certRequest, vasp); err != nil {
+		return nil, err
+	}
```
bbengfort, line 514: I'm hoping that when the webhook/email delivery is on the VASP we can simply add the data after line 485 above without having to jump through the GDS extra hoops. See the comment in the proto file.

bbengfort (commented):
Thank you for putting this PR together @pdeziel -- it's a great start towards actual, production certificate delivery via webhook! 

It looked like there was a little confusion in the `reissuer` command, which hopefully will be easily resolved. 

As for the protocol buffers, rather than storing this information on extra, I'd like to store it on the VASP record. The webhook is going to be a client managed field, e.g. from the BFF/GDS UI and needs to be part of the TRISA API. Hopefully this will make things easier since you won't have to deal with the gds extra data being an Any, but it will require an update to the protocol buffers, generated code, and helper method/tests you wrote in this PR. 

Let me know if you have any questions!

proto/gds/models/v1/models.proto
pdeziel: Alright that makes a bit more sense to me, especially if we're committing to doing the migration to include the new fields.

cmd/reissuer/main.go
pdeziel: I think deliver makes more sense to me, thanks for the suggestion!

cmd/reissuer/main.go
```diff
@@ -404,6 +415,15 @@ func reissueCerts(c *cli.Context) (err error) {
...
 	}
 
+	// Override the certificate delivery webhook if specified
+	var webhook string
+	if webhook = c.String("webhook"); webhook != "" {
+		certreq.Webhook = webhook
+	}
+
+	// Override the email delivery preference
+	certreq.NoEmailDelivery = true
```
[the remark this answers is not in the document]
pdeziel: D'oh! Thanks for catching that, however I'm wondering if by default it should be `no-email=false` (always overriding the setting on the VASP), or by default it should use the current setting on the VASP? I haven't used the `reissuer` CLI before so do you have an opinion on that?

cmd/reissuer/main.go
```diff
@@ -404,6 +415,15 @@ func reissueCerts(c *cli.Context) (err error) {
 		certreq.DnsNames = append(certreq.DnsNames, dnsNames...)
 	}
 
+	// Override the certificate delivery webhook if specified
+	var webhook string
+	if webhook = c.String("webhook"); webhook != "" {
+		certreq.Webhook = webhook
+	}
```
bbengfort, line 422: Creating a variable in a single lexical block gives the compiler more opportunities for optimization. 

```suggestion
	if webhook := c.String("webhook"); webhook != "" {
		certreq.Webhook = webhook
	}
```

bbengfort (approved):
Thank you for making those changes!

cmd/reissuer/main.go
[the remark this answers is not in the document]
bbengfort: Hm, that's a good question. For the demo we'll want to set it manually, so I think the default should be `no-email = false` for now; perhaps in the future if we're getting annoyed with having to manually set things we can load it from the VASP first?

PR #1149, opened 2024-01-22
Add util to remove duplicate orgs from appdata

daniellemaxwell:
### Scope of changes

Adds `appdata:dedupeorgs` script to `bffutil` to aid in locating and remove duplicate organizations.

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/cmd/bffutil/main.go
+++ b/cmd/bffutil/main.go
@@ -240,6 +240,19 @@ func main() {
 				},
 			},
 		},
+		{
+			Name:   "appdata:dedupeorgs",
+			Usage:  "remove duplicate organizations from a user's app_metadata",
+			Action: dedupeAppdataOrgs,
+			Before: Before(loadConf, connectAuth0),
+			Flags: []cli.Flag{
+				&cli.BoolFlag{
+					Name:    "dry-run",
+					Aliases: []string{"d"},
+					Usage:   "display user duplicate orgs without removing them",
+				},
+			},
+		},
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/bffutil/main.go
```diff
@@ -1089,6 +1102,50 @@ func sortAppdataOrgs(c *cli.Context) (err error) {
...
+		}
+
+		// Get the user's current org list
+		orgs := appdata.GetOrganizations()
+		if len(orgs) == 0 {
+			continue
+		}
+
+		// Check for duplicate orgs and remove them.
+		seen := make(map[string]bool)
```
pdeziel, line 1126: One trick I usually use for things like this is to use an empty struct instead of the bool.

```
seen := make(map[string]struct{})
```

Then instead of `seen[org] = true`, you just do `seen[org] = struct{}{}`. It doesn't matter much since this is a script, but using the struct instead of the bool actually saves some memory and the existence checks will work the same way (`struct{}{}` doesn't take up any space but `bool` does).

pdeziel (approved):
Very useful! It's very clear what the command is doing and I like the inclusion of the dry-run argument.

cmd/bffutil/main.go
daniellemaxwell: That's great to know. Thanks @pdeziel!

PR #1162, opened 2024-02-28
Set review stepper state to complete after user submits registration

daniellemaxwell:
### Scope of changes

Sets the stepper status to complete after a user submits the registration form.

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

https://www.awesomescreenshot.com/video/25370460?key=2b8264fa9105338e9f47cf1ebeb01572

### Author checklist

### Reviewer(s) checklist

### DIFF
--- a/web/gds-user-ui/src/components/ReviewSubmit/index.tsx
+++ b/web/gds-user-ui/src/components/ReviewSubmit/index.tsx
@@ -18,12 +18,13 @@ import useCertificateStepper from 'hooks/useCertificateStepper';
 import { STEPPER_NETWORK } from 'utils/constants';
 
 import WarningBox from 'components/WarningBox';
-import { setHasReachSubmitStep } from 'application/store/stepper.slice';
+import { setHasReachSubmitStep, setStepStatus } from 'application/store/stepper.slice';
 import { useAppDispatch } from 'application/store';
 import { StepsIndexes } from 'constants/steps';
 import { useFetchCertificateStep } from 'hooks/useFetchCertificateStep';
 import { StepEnum } from 'types/enums';
 import useSubmissionStatus from 'modules/dashboard/registration/hooks/useSubmissionStatus';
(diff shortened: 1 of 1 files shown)


web/gds-user-ui/src/components/ReviewSubmit/index.tsx
```diff
@@ -71,6 +72,11 @@ const ReviewSubmit: React.FC<ReviewSubmitProps> = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [isSent]);
 
+  // Update the review step status to complete if the user submits a registration.
+  if (isTestNetSubmitted || isMainNetSubmitted) {
```
daniellemaxwell, line 76: The stepper label color doesn't immediately update after the registration is submitted. To fix this, I think `react-query` will have to be used which would require some refactoring that should be handled in a separate story.

web/gds-user-ui/src/components/ReviewSubmit/index.tsx
```diff
@@ -71,6 +72,11 @@ const ReviewSubmit: React.FC<ReviewSubmitProps> = ({
...
+    dispatch(setStepStatus({ step: 6, status: LSTATUS.COMPLETE }));
```
pdeziel, line 77: What is the significance of the 6? Is it the step number?

pdeziel (approved):
Makes sense to me, I just had one question about the `setStepStatus` method.

web/gds-user-ui/src/components/ReviewSubmit/index.tsx
daniellemaxwell: Yes, 6 is the review step number. There are constants for steps 1-5. I was unable to determine why 6 was excluded, since 6 is listed as a step number a few times in the code. However, I did notice some cases where the constants weren't used for other step numbers, so a story updating this should be added.

PR #1166, opened 2024-03-18
Send error details to Sentry

daniellemaxwell:
### Scope of changes

Fixes `Object captured as exception` Sentry error.

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

### Author checklist

### Reviewer(s) checklist


web/gds-user-ui/src/modules/auth/login/index.tsx
```diff
@@ -99,7 +99,7 @@ const StartPage: React.FC = () => {
       }
 
       // catch this error in sentry
-      Sentry.captureException(err);
+      Sentry.captureException(err?.original);
```
daniellemaxwell, line 102: After reading this comment, it seems that Sentry didn't think this was an error object: https://github.com/getsentry/sentry-react-native/issues/391#issuecomment-419579923

`err?.original` returns an object inside the error object with more details, so I believe this should resolve the issue in Sentry.

bbengfort: When it comes to the `?.` syntax, what will be passed into `captureException` if the `original` property is not on `err`? Probably we would just want to pass in `err` in that case, right?

I may just be unfamiliar with `?.` but how does this code differ?

```javascript
try {
    Sentry.captureException(err);
} catch {
     Sentry.captureException(err?.original);
}
```

or is it more like:

```javascript
if (err?.original) {
    Sentry.captureException(err?.original);
} else {
    Sentry.captureException(err);
}
```

Or is it that we can always guarantee that there is an `err.original` on the error object?

bbengfort (approved):
Thanks for making this fix!

web/gds-user-ui/src/modules/auth/login/index.tsx
daniellemaxwell: `err` returns an object with the error, error code, error description, and another object (`original`) that contains the error and error description.
Screenshot 2024-03-19 at 2 50 25 PM

`?.` will check if `err` is undefined before attempting to read what's inside the original object. It was added to prevent an `undefined` error from being returned.

If original isn't included, then `err` wouldn't be passed into the exception. I didn't include an else because we currently pass `err` into Sentry.captureException and get an error message. I thought possibly passing in this object directly could be helpful. The error didn't trigger in development mode for me to confirm if this is truly the solution.

From what I can see, `err.original` will be included in the `err` object. However, this is coming from `Auth0` so I can't quite guarantee this won't ever change.

bbengfort: Ok, I see now that this is already in a try/catch block for Auth0 specific handling. Thank you for clarifying this!

PR #1176, opened 2024-11-19
Fix Ensign Config Preventing GDS from Running

bbengfort:
### Scope of changes

The confire library automatically calls a configurations `Validate()` method if it implements it in order to process a valid configuration. We did not want the Ensign config to be validated unless the Activity config (which it was a member of) was enabled. Similarly the `Network` field also had a `Validate` method which is why those two errors were coming up. This PR fixes those problems. 

### Type of change

- [ ] bug fix
- [ ] new feature
- [ ] documentation
- [ ] other (describe)

### Acceptance criteria

This PR will be merged without review.

### Author checklist

### DIFF
--- a/pkg/utils/activity/config.go
+++ b/pkg/utils/activity/config.go
@@ -1,30 +1,34 @@
 package activity
 
 import (
+	"errors"
 	"time"
 
 	"github.com/trisacrypto/directory/pkg/utils/ensign"
 )
 
 type Config struct {
-	Enabled           bool          `split_words:"true" default:"false"`
-	Topic             string        `split_words:"true"`
-	Network           Network       `split_words:"true"`
+	Enabled           bool          `default:"false"`
+	Topic             string        `required:"false"`
+	Network           Network       `required:"false"`
 	AggregationWindow time.Duration `split_words:"true" default:"5m"`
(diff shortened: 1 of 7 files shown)


pkg/utils/activity/config.go
```diff
@@ -1,30 +1,34 @@
...
 	"github.com/trisacrypto/directory/pkg/utils/ensign"
 )
 
 type Config struct {
-	Enabled           bool          `split_words:"true" default:"false"`
-	Topic             string        `split_words:"true"`
-	Network           Network       `split_words:"true"`
+	Enabled           bool          `default:"false"`
+	Topic             string        `required:"false"`
+	Network           Network       `required:"false" validate:"ignore"`
```
bbengfort, line 13: This fixes the unknown network error

pkg/utils/activity/config.go
```diff
@@ -1,30 +1,34 @@
...
 	AggregationWindow time.Duration `split_words:"true" default:"5m"`
-	Testing           bool          `split_words:"true" default:"false"`
-	Ensign            ensign.Config
+	Testing           bool          `default:"false" `
+	Ensign            ensign.Config `validate:"ignore"`
```
bbengfort, line 16: This fixes the client id missing error

